PR #1, opened 2026-04-15
chore: v1.1 polish — skills, panel fixes, README SEO

Astro-Han:
## Summary

- **Rewrite 3 built-in skills** with CSO-optimized structure (description uses "Use when..." triggers, unified sections, ~30% fewer words)
- **Fix right panel auto-open** — switch data source from message summary (turnDiffs) to Snapshot (sync.data.session_diff), the authoritative source
- **Fix file-tree button highlight** — button now stays active when side panel is open on any tab (not just Files), since the standalone diff icon was removed earlier
- **Rewrite dual-language READMEs** — concrete use cases, badges, "Who This Is For" section, fix license label from MIT to Apache-2.0

## Test plan

- [ ] Open a session, trigger file changes — right panel should auto-expand
- [ ] With panel open on Changes tab, verify file-tree button in header is highlighted
- [ ] Click file-tree button to toggle panel open/closed on both tabs
- [ ] Verify skill cards still work (document-processing, data-analysis, writing-assistant)
- [ ] Check README renders correctly on GitHub

### DIFF
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -1093,12 +1093,15 @@ export default function Page() {
   createEffect(() => {
     if (!params.id) return
 
+    // Use Snapshot diffs (SSE-pushed, authoritative) with turnDiffs as fallback
+    // for reopened sessions where session_diff hasn't been fetched yet.
+    const source = diffs().length > 0 ? diffs() : turnDiffs()
     const next = nextFilesPanelAutoOpen(
       {
         seenAdded: view().sidePanel.filesAutoOpenSeen(),
         dismissed: view().sidePanel.filesAutoOpenDismissed(),
       },
-      diffs(),
+      source,
     )
 
     if (next.open) {
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -1098,7 +1098,7 @@ export default function Page() {
         seenAdded: view().sidePanel.filesAutoOpenSeen(),
         dismissed: view().sidePanel.filesAutoOpenDismissed(),
       },
-      turnDiffs(),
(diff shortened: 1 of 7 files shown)


packages/app/src/pages/session.tsx
```diff
@@ -1098,7 +1098,7 @@ export default function Page() {
         seenAdded: view().sidePanel.filesAutoOpenSeen(),
         dismissed: view().sidePanel.filesAutoOpenDismissed(),
       },
-      turnDiffs(),
+      diffs(),
```
Astro-Han: [P1] Files auto-open now depends on data that is not loaded while the panel is closed

Replacing `turnDiffs()` with `diffs()` breaks the auto-open path for reopened sessions. `diffs()` comes from `sync.data.session_diff`, but that store is only fetched later in this file when `wantsReview()` is true. With the side panel still closed, that fetch never runs, so this effect sees an empty diff list and will not auto-open the Files tab even when the last turn created files.

Astro-Han (commented):
Posting the requested review finding.

packages/app/src/pages/session.tsx
Astro-Han: Verified — `sync.session.diff(id)` at line 1405 is gated by `wantsReview()`, which is false when the panel is closed. So reopened sessions would never fetch `session_diff`, and `diffs()` stays empty.

Active sessions are fine because the `session.diff` SSE event pushes directly to the store (event-reducer.ts:166), bypassing the `wantsReview` guard.

Fixed in 412a8ec: use Snapshot when available (SSE-pushed, authoritative for active sessions), fall back to `turnDiffs` (message summary) for reopened sessions where Snapshot hasn't been fetched yet.

PR #3, opened 2026-04-16
fix(config): wait for background installs on dispose

Astro-Han:
## Summary
- wait for background config dependency installs during config instance teardown
- cover the Windows `config-install:win32` lock leak with a regression test that models disposal plus a follow-up install
- keep the fix scoped to config lifecycle cleanup, without broad refactors

## Root Cause
On Windows, config dependency installs share the global flock key `config-install:win32`. We started those installs in the background during config loading, but instance disposal did not wait for them to settle. A previous instance could therefore keep holding the lock after teardown, and later tests would time out waiting on the leaked lock.

## Testing
- bun test ./test/config/config.test.ts
- bun test ./test/tool/registry.test.ts
- bun test ./test/snapshot/snapshot.test.ts -t "snapshot state isolation between projects"
- bun test ./test/project/worktree.test.ts

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,10 @@ jobs:
         with:
           bun-version: "1.3.11"
 
+      # Restoring Bun's download cache on Windows was empirically slower than a cold install
+      # and consumed enough of the 30-minute job budget to starve opencode:test:ci.
       - uses: actions/cache@v4
+        if: runner.os != 'Windows'
         with:
           path: |
             ~/.bun/install/cache
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -27,7 +27,6 @@ import { Bus } from "@/bus"
 import { GlobalBus } from "@/bus/global"
 import { Event } from "../server/event"
 import { Glob } from "../util/glob"
-import { iife } from "@/util/iife"
 import { Account } from "@/account"
 import { isRecord } from "@/util/record"
 import { ConfigPaths } from "./paths"
@@ -1417,6 +1416,9 @@ export namespace Config {
           }
 
           const deps: Promise<void>[] = []
+          yield* Effect.addFinalizer(() =>
+            Effect.promise(() => Promise.allSettled(deps).then(() => undefined)),
+          )
 
           for (const dir of unique(directories)) {
             if (dir.endsWith(".opencode") || dir === Flag.OPENCODE_CONFIG_DIR) {
@@ -1430,9 +1432,7 @@ export namespace Config {
               }
             }
 
-            const dep = iife(async () => {
-              await installDependencies(dir)
-            })
+            const dep = installDependencies(dir)
             void dep.catch((err) => {
               log.warn("background dependency install failed", { dir, error: err })
             })
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1073,7 +1073,10 @@ export namespace Config {
   type State = {
     config: Info
     directories: string[]
-    deps: Fiber.Fiber<Exit.Exit<void, unknown>, never>[]
+    deps: {
+      dir: string
+      fiber: Fiber.Fiber<Exit.Exit<void, unknown>, never>
+    }[]
     consoleState: ConsoleState
   }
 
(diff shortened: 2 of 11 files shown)


Astro-Han:
Updated this PR with two follow-up commits:

- `ff8ec43be` `fix(config): wait for background installs on dispose`
- `a201d0a91` `fix(config): harden scoped dependency installs`

What changed in this update:

- Reworked config-scoped dependency installs so waiting on the Windows global config lock is interruptible, while the actual install phase remains protected once the lock is acquired.
- Fixed the `disposeAll()` path so background config installs waiting on the lock do not hang instance teardown.
- Restored real failure propagation from `Config.waitForDependencies()` instead of reducing install failures to warning logs only.
- Stopped swallowing lock release errors.
- Aligned the `tool.registry` caller with the new error semantics.
- Tightened config tests by isolating `OPENCODE_CONFIG_DIR` cases from shared global config state.
- Added and strengthened Windows regressions for:
  - disposing while a background config install is waiting on the win32 lock
  - aborting while waiting on the win32 lock

Re-verified locally after the latest changes:

- `bun run typecheck`
- `bun test ./test/config/config.test.ts`
- `bun test ./test/tool/registry.test.ts`
- `bun test ./test/provider/provider.test.ts -t "plugin config providers persist after instance dispose"`

Also ran a fresh-eyes review loop on the final diff. The last review came back with no P0/P1 findings.

Astro-Han:
Quick status update on this PR.

The scope has narrowed quite a bit. The recent commits have all been in the same area: config-scoped dependency installs, tool discovery timing, and Windows lock cleanup. The latest commit adds a retry in `Flock.release()` for transient lock-dir removal errors on Windows, plus a regression test for that path. Before pushing, I re-ran `bun turbo typecheck` and `bun turbo test:ci` locally on macOS, and both passed.

The current blocker in the latest CI run is on Ubuntu. It fails in `packages/opencode/test/tool/registry.test.ts`, in the case `does not wait for unrelated global config installs before importing local tools with bare imports`. That test is specifically checking that local tool discovery does not get blocked by an unrelated global install. In CI it timed out waiting for the local tool id `late`, so that expectation is still not being met there.

Windows is still running as I post this. `bun install --frozen-lockfile` and `typecheck` have completed, and the `test` step is in progress. Once that job finishes, I will know whether we are looking at the same wait-path on both platforms or one Ubuntu-specific issue plus something separate on Windows.

Assuming Windows does not reveal a different failure first, the next fix should stay narrow: adjust the registry/config wait boundary so unrelated global installs do not stall local tool loading.

Astro-Han:
Update on the last four commits:

- `0f7c21c` only changes CI and the registry timing expectation. I disabled the Windows Bun cache in `.github/workflows/ci.yml` and tightened the related registry test setup after the runner kept behaving differently from Ubuntu.
- `4426727` fixes a Windows lock handoff race in config-scoped dependency installs. This is the commit that hardens the `flock` path and adds focused regression coverage for that handoff.
- `1c800b7` moves one layer up and stabilizes Windows cleanup around config, worktree, and local tool loading. That commit adds regression tests for transient cleanup failures and for config-scoped local tool installs.
- `eafddd5` closes the remaining `tui.test.ts` flake. The test tmpdir now waits for background `TuiConfig` installs before cleanup, and the one test in that file that also calls `Config.get()` now waits for `Config.waitForDependencies()` as well.

So the recent work is not four unrelated patches. It narrows down the Windows failures to one class of problems: background installs, lock handoff, and temp-dir cleanup happening in the wrong order on the Windows runner.

Latest local verification on this branch is green:

- `bun turbo typecheck`
- `bun turbo test:ci`

Most recent local `test:ci` result: `1932 pass / 11 skip / 1 todo / 0 fail`.

Astro-Han:
关闭，改走拆分路径：

原始意图（Instance dispose 等待 background install）是有实质的，但 12 个 commit 滚雪球叠了三类无关改动：

1. flock/worktree EBUSY 重试 — 全是 synthetic mock 测试，缺真实 CI 日志佐证
2. flock 'handoff microtask race' — 理论构造，测试靠模块重写注入钩子
3. Windows 跳 bun cache — 无对照数据

CI 连红 15+ 小时，最近的 revert→restore 类型错误说明已经在改不理解的代码。

后续拆：
- fix(config): interrupt background installs on dispose  ← 保留 fiber 化
- fix(worktree): retry Windows transient rm failures  ← 先附真实 EBUSY CI log 再提
- 其余不再单提

PR #6 已经证明同一 scope 可以做到 200 行聚焦修复，按那个粒度推进。

PR #5, opened 2026-04-17
fix(config): harden scoped dependency loading

Astro-Han:
## Summary
- harden config-scoped dependency installs and waiting behavior
- follow transitive config dependencies across local tool and plugin helpers
- ignore type-only dependency signals while preserving runtime aliases
- include the supporting Windows cleanup and lock reliability fixes already in this branch stack

## Verification
- bun test --timeout 30000 test/config/config.test.ts test/tool/registry.test.ts test/plugin/loader-shared.test.ts
- git diff --check

## Notes
- This PR is the clean replacement for PR #3 so review can happen against a single branch-level diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,10 @@ jobs:
         with:
           bun-version: "1.3.11"
 
+      # Restoring Bun's download cache on Windows was empirically slower than a cold install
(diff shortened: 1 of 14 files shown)


Astro-Han:
Superseded by #6, which uses the correct head branch: fix/config-scoped-deps-review.

PR #7, opened 2026-04-17
fix(config): wait for background installs on dispose

Astro-Han:
## Summary
Cherry-pick of commit `ff8ec43b` from the now-closed PR #3, isolated to the one change that actually fixes Windows CI red.

Root cause: On Windows, background config dependency installs share the global flock key `config-install:win32`. The install fibers outlive their originating `Instance` context, keep holding the lock after dispose, and the next test times out waiting for the leaked lock.

Fix: add an `Effect.addFinalizer` so `Instance.disposeAll()` waits (via `Promise.allSettled`) for background installs before releasing scope.

## Why only this commit
PR #3 bundled 12 commits. Analysis showed only this one had a clear hypothesis → fix relationship. The rest were either:
- Speculative (flock handoff microtask race — fabricated via module-rewrite test hooks)
- Defensive without evidence (rm EBUSY retries with synthetic mocks, no real CI log showing EBUSY)
- Test-infra chasing test-infra (9 commits patching flake introduced by earlier test additions)
- Revert/restore thrash on a typecheck error the author was ping-ponging on

PR #3 CI was red for 15+ hours across 12 commits. This single commit on `dev` should reproduce the original hypothesis cleanly.

## Product change
`packages/opencode/src/config/config.ts`: +4/-4. Removed `iife` wrapper, added finalizer that awaits `deps: Promise<void>[]`.

## Test change
`packages/opencode/test/config/config.test.ts`: +81. Windows-scenario test `Instance.disposeAll waits for background config dependency installs to finish` — asserts the second install starts only after the first completes, not before dispose returns.

## Verification
- `bun test --timeout 30000 test/config/config.test.ts` — 77/77 pass locally (macOS)
- `bun run typecheck` — clean
- Windows CI is the real validation; this PR's goal is to turn that green

## Target failing tests on dev (Windows)
- installs dependencies in writable OPENCODE_CONFIG_DIR — timeout 30s
- dedupes concurrent config dependency installs for the same dir — timeout 30s
- tool.registry > waits for config-scoped dependencies before importing local tools with bare imports — timeout 30s

All three share the `config-install:win32` flock-leak root cause this PR addresses. `tracks deleted files correctly` in snapshot.test.ts is a separate failure not addressed here.

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -27,7 +27,6 @@ import { Bus } from "@/bus"
 import { GlobalBus } from "@/bus/global"
 import { Event } from "../server/event"
 import { Glob } from "../util/glob"
-import { iife } from "@/util/iife"
 import { Account } from "@/account"
 import { isRecord } from "@/util/record"
 import { ConfigPaths } from "./paths"
@@ -1420,6 +1419,9 @@ export namespace Config {
           }
 
           const deps: Promise<void>[] = []
+          yield* Effect.addFinalizer(() =>
+            Effect.promise(() => Promise.allSettled(deps).then(() => undefined)),
+          )
 
           for (const dir of unique(directories)) {
             if (dir.endsWith(".opencode") || dir === Flag.OPENCODE_CONFIG_DIR) {
@@ -1433,9 +1435,7 @@ export namespace Config {
               }
             }
 
-            const dep = iife(async () => {
-              await installDependencies(dir)
-            })
+            const dep = installDependencies(dir)
             void dep.catch((err) => {
               log.warn("background dependency install failed", { dir, error: err })
             })
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1419,8 +1419,13 @@ export namespace Config {
           }
 
           const deps: Promise<void>[] = []
+          const abortInstalls = new AbortController()
           yield* Effect.addFinalizer(() =>
-            Effect.promise(() => Promise.allSettled(deps).then(() => undefined)),
+            Effect.sync(() => abortInstalls.abort()).pipe(
+              Effect.flatMap(() =>
+                Effect.promise(() => Promise.allSettled(deps).then(() => undefined)),
+              ),
+            ),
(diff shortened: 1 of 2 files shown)


Astro-Han:
Closing. Multi-reviewer crosscheck confirmed the fix is net-negative.

## Why close (not iterate)

**The motivating bug is not a current-dev blocker.** PR #6's early-exit in `installDependencies` (`if (hasDep && ignore && installed.every(Boolean)) return`) means most test scenarios no longer trigger real `Npm.install`. Dev CI run #38 green corroborates. Run #39 on same test tree failed 3 unrelated tests — runner-level flake, not lock leak.

**The implementation breaks teardown on Windows.** The added `Effect.addFinalizer` in `loadInstanceState` awaits `Promise.allSettled(deps)`. The `AbortController` added in the second commit only propagates to `Flock.acquire` — it does NOT reach `Npm.install(dir)` at config.ts:196, which calls `Arborist.reify()` (`@npmcli/arborist@9.4.0`) with no `AbortSignal` plumbing (verified by reading arborist source). Any install already past the outer flock blocks the finalizer for the full reify duration — ~30s on Windows CI — causing the `beforeEach/afterEach hook timed out` cascade we observed in runs #36 (16 fails) and #40 (18 fails).

**The design conflicts with an existing opt-in pattern.** `Config.waitForDependencies()` already exists (config.ts:1583) and callers that need it already use it (tool/registry.ts:184, plugin/index.ts:172). Making dispose auto-wait forces every test/path to pay this cost.

## Tests are not reliable evidence here

Local tests (86/86 pass) mock `Npm.install` with an instant-returning promise, so they never exercise the real `arb.reify()` path that's the root of the Windows hang. Any future fix in this area needs either Windows-environment validation or a mock that simulates realistic install duration.

## Residual flake

Dev run #38/#39 delta (same tree, one docs deletion) shows Windows runner-level flake affecting tests like `loop waits while shell runs` at 3s timeout. The right fix is a separate CI-level retry mechanism (`bun test --rerun-each N` inside a new PR), not any code change here.

## Branch

`fix-dispose-background-installs` deleted on close.

PR #18, opened 2026-04-17
chore(branding): strip OpenCode residue for Phase 1

Astro-Han:
## Summary

Sweeps user-visible OpenCode brand residue discovered during the v0.2.3 smoke test (rolled back on 2026-04-17). The published v0.2.3 Release was deleted because the dmg shipped with Apple's design-template placeholder icons and various OpenCode strings in dialogs, settings, update toasts, etc. This PR covers the explicit-brand surfaces and narrows the Phase-1 product scope (1 theme, 2-3 locales) so a future re-release lands clean.

## Scope and rationale

| Commit | What | Why |
|---|---|---|
| [`622a648`](../commit/622a648c7) | HTML title, PWA manifest, root + desktop-electron `package.json` author/homepage/description/repo URL | Direct `OpenCode` strings the user would see in tab title / PWA shortcut / About-panel metadata |
| [`ff02a27`](../commit/ff02a27db) | Prune theme system from 38 themes to 1 (pawwork only) | 38 themes = 38-file sync burden per CSS-token refactor; target user (non-programmer) won't open a theme picker; future-proofs frontend restructure |
| [`086deb4`](../commit/086deb444) | i18n locale set from 17 → en/zh/zht | Inherited locales had no active maintainer. Phase 1 audience = zh (纯刻 + 中文市场) + en (global OSS). Non-kept locales would silently display stale OpenCode text once we drifted |
| [`b5696b8`](../commit/b5696b8be) | Sed sweep of `OpenCode` → `PawWork` in `en/zh/zht` values | Used sentinel two-pass to preserve `OpenCode Zen` / `OpenCode Go` service names and `opencode.ai/zen` URL — Zen is a real third-party model-routing service PawWork uses as backend, renaming would misattribute upstream |
| [`5c346546`](../commit/5c346546f) | Config loader accepts `pawwork.json` / `pawwork.jsonc` alongside legacy `opencode.json` / `opencode.jsonc` | Additive change: 5 loader touchpoints + all user-visible help strings updated. Existing configs keep working, new installs see `pawwork.json` as the canonical filename |

## Deliberately out of scope

- `.opencode/` directory-name detection (additive `.pawwork/` = larger refactor; legacy still works)
- `@opencode-ai/*` import path renames (~574 refs, blocked on Zen proxy header dependency verification per `project_rename_blockers` memory)
- `opencode-theme-id` / `opencode.global.dat` localStorage keys (DevTools-only visibility, renaming breaks existing user data)
- JSON schema `$id` URL in `desktop-theme.schema.json`
- CLI TUI theme `packages/opencode/src/cli/cmd/tui/context/theme/opencode.json` (separate subsystem)
- Brew/scoop manifest URLs in `installation/index.ts` (upstream OpenCode distribution channels)
- Proper `pawwork-dark` palette — `pawwork.json`'s dark section is currently a copy of light; `context.tsx` forces light via hardcoded check; designing a real dark variant is follow-up

## Verification

- `bun run typecheck` green across all 7 packages after each commit
- Parity test (`packages/app/src/i18n/parity.test.ts`) now asserts zh + zht translate target keys (down from 16)

## Test plan

- [ ] After merge: trigger `build.yml` with `phase=full, channel=prod, arch=arm64`
- [ ] Verify draft Release `v0.2.X` produced with correct paw-print icon visible on mounted dmg
- [ ] Smoke test app launches on clean macOS (Gatekeeper silent, menu bar shows PawWork, About dialog shows PawWork, browser tab — if web dev — shows PawWork)
- [ ] Confirm existing `opencode.json` configs still load without warnings
- [ ] Confirm app detects `pawwork.json` when present

## Not included in this PR

- README update (was reverted during v0.2.3 rollback; will redo on next real release)
- `pawwork-dark` palette design
- Public Release republish (do after merge + smoke test pass)


Astro-Han:
I don't think this is ready to merge.

This change affects the upgrade path for existing installs, not just the settings surface. Startup still trusts `opencode-theme-id` from localStorage, but this PR removes every bundled theme except `pawwork`. If an existing user previously selected `dracula`, `nightowl`, or still has the old `oc-1` / `oc-2` path in storage, the app can start with a theme id that no longer exists. In that state, we do not fall back to `pawwork`, and preload can still apply stale cached CSS.

I'm treating this as P1 and merge-blocking. Before merge, we should migrate unknown stored theme ids to `pawwork` during startup and add a regression test that covers persisted legacy themes. Otherwise we fix the fresh install default and break the upgrade path for existing users.

Astro-Han:
Addressed the review plus two additional crosscheck rounds. Three follow-up commits pushed:

**`715612e41` fix(theme): migrate legacy theme ids to pawwork on startup**

Fixes the original concern. `oc-theme-preload.js` and `context.tsx` now treat any stored theme id other than `pawwork` as legacy and migrate it to `pawwork`, clearing the cached CSS keys (`opencode-theme-css-light/dark`). No more stale dracula/nightowl/oc-1/oc-2 CSS flashing on upgrade. Regression tests in `packages/app/src/theme-preload.test.ts` cover `oc-1`, `oc-2`, `dracula`, `nightowl`, `amoled`.

**`691404eba` fix(review): address crosscheck findings**

Second crosscheck round (Codex + Claude) surfaced:

- P1 `zh-TW` / `zh-HK` / `zh-MO` were resolving to Simplified Chinese because the detectors only matched the literal `hant` substring. Extracted `isTraditionalChinese()` helpers in `packages/app/src/context/language.tsx` and `packages/desktop-electron/src/renderer/i18n/index.ts`, unit tests added.
- P2 Removed the dead `oc-2` branches and the now-unused `normalize()` helper from `packages/ui/src/theme/context.tsx`; trimmed the theme label map to `pawwork` only.
- P2 Rebrand residue: `apple-mobile-web-app-title` in `favicon.tsx`, schema title/description in `desktop-theme.schema.json`, `server-errors.ts` fallback, and the `status-popover-body.tsx` plugin-empty split token (was splitting the translation by `opencode.json` after the translation was updated to `pawwork.json`).

**`6aa109dcd` fix(tests): address fresh crosscheck findings**

Third round caught stale e2e tests:

- Rewrote `legacy oc-1 theme migrates to oc-2` → `legacy theme ids migrate to pawwork and clear cached CSS`, seeding `dracula` to match the Phase-1 behavior.
- Skipped two e2es that depended on multi-theme and dark-mode switching (the color-scheme toggle and theme-dropdown `count > 1` assertion). Phase-1 bundles a single light-locked pawwork theme, so these can't meaningfully run until a dark palette or a second theme lands; skip comments flag that.
- Dropped the Dark-click + dark-mode assertion from the font-rehydrate e2e and renamed it accordingly — font persistence still exercised.
- `changing language` e2e switched from Deutsch (removed locale) to 简体中文 / 通用.
- Renamed the tautological preload test ("does not touch localStorage when the stored theme is already PawWork" was wrong because preload still writes `opencode-color-scheme=light`) and added the color-scheme assertion.

Pre-existing on `dev`, not this PR's regression, left for a future PR: zh/zht parity gap for `command.project.next/previous` and `session.review.noBranch/NoUncommittedChanges`; `Config.update()` hardcoded write to `config.json`; `.opencode/` dotfolder rename; JSON schema `$id` URL; `@opencode-ai/*` package rename.

All tests green (`bun test` 330/330 app, 11/11 ui theme/preload, 3/3 electron i18n); `bun run typecheck` clean across all 7 workspaces.

Astro-Han:
Confirmed — `config.ts` went through `ConfigPaths.projectFiles("opencode", ...)` only, so a project-root `pawwork.json` was invisible on startup.

Fixed in `ff0d43105`: walk the directory tree twice (opencode.json(c) first, pawwork.json(c) second) and feed both sequences into the existing merge loop. At the same directory level, pawwork wins over opencode, matching the precedence that `globalConfigFile` and the managed-dir loops already use. `.opencode/` dotfolder configs are unchanged — that rename is still deferred.

`bun run typecheck` clean; `bun test` in packages/opencode: 1926 pass / 0 fail.

Astro-Han:
Retargeted two remaining OpenCode URLs in `c20614c20`:

- **Sidebar help button** (bottom-left ⓘ) pointed at `opencode.ai/desktop-feedback`. Switched to `github.com/Astro-Han/pawwork/issues`, the actual feedback surface.
- **"What's new" release-notes fetcher** pointed at `opencode.ai/changelog.json`, which would have surfaced OpenCode changelog entries on the next version bump. Switched to the GitHub Releases API for this repo; `parseRelease` now falls back to synthesizing a single PawWork highlight from `body` when the structured `highlights` array is absent (markdown headings skipped, bullet markers stripped, truncated to 200 chars). Structured-schema path is preserved for any future switch to a hosted `changelog.json`. Tests cover both code paths.

Not touched in this PR (docs links to upstream OpenCode docs that are low-frequency and still useful for advanced users; queued as a separate docs-link-cleanup):
- `dialog-custom-provider.tsx` → `opencode.ai/docs/providers/#custom-provider`
- `settings-general.tsx` → `opencode.ai/docs/themes/`
- `opencode.ai/zen` links are intentionally preserved (Zen is a real third-party provider).

Astro-Han:
Follow-up `fb3d02a36`: the dev-only `DebugBar` in the bottom-right corner was overlapping the "next step" button on Show Case cards during local review. Added a persistent collapse toggle — click × to shrink to a small "debug" pill in the same corner, click the pill to re-expand. State persists via localStorage (`pawwork-debug-bar-collapsed`). Dev-only (`import.meta.env.DEV`), no effect in packaged builds.

Separate product bug discovered while reviewing but **out of scope** for this PR, filed as [#21](https://github.com/Astro-Han/pawwork/issues/21): document-processing and data-analysis Show Case skills skip `AskUserQuestion`. Root cause is prompt-level (all three SKILL.md files only say "ask the user" in prose; need to name the tool), not code. Writing skill is unaffected.

Astro-Han:
Round 6 crosscheck (both reviewers complete, no truncation). Two commits pushed:

**`34366ca9e` fix: address fresh crosscheck findings**
- **P1 project config precedence** — `ConfigPaths.projectFiles` now accepts an array of names; `loadAll` passes `["opencode", "pawwork"]` so a single `findUp` pass interleaves them per directory. Innermost directory wins overall (via `rootFirst`); within the same directory pawwork wins over opencode. Fixes the case where a parent `pawwork.json` could override a nearer child `opencode.json`.
- **P2 brand residue** in OAuth callback pages (`codex.ts`, `oauth-callback.ts` HTML_SUCCESS/ERROR titles + body text) and `uninstall.ts` CLI intro/outro/describe.
- **P2 tautological test** — dropped `getElementById("oc-theme-preload")` null-checks that were always true (preload.js never creates that element).

**`27a29f91c` fix(brand): MCP OAuth client + TUI terminal title**
- **P2 `mcp/oauth-provider.ts`** — `client_name` / `client_uri` sent during OAuth registration were still `"OpenCode"` / `https://opencode.ai`, visible on MCP provider consent pages. Rebranded.
- **P2 `cli/cmd/tui/app.tsx`** — terminal window title + post-update dialog text + session-title prefix `OC | ...` rebranded to PawWork / `PW | ...`.
- **P3 theme-preload test** — renamed `keeps PawWork light-only even when the system prefers dark` to describe what's actually verified (stored dark scheme → forced to light). Dropped the dead `matchMedia` mock; the simplified preload script never reads `matchMedia`.

**Pre-existing upstream bug filed separately** as [#22](https://github.com/Astro-Han/pawwork/issues/22): `.opencode/` directory cascade merges inner→outer, so outer configs override inner. `Filesystem.up` yields innermost-first and merge is last-wins, reversing the intended "innermost wins" precedence. PR #18 did not touch this iteration; fix needs its own review + regression tests.

Declined (intentional / out of scope):
- `STORAGE_KEYS` / language cookie legacy `opencode-*` prefixes — preserved for read-compat with existing installs. Sweeping these would require a migration shim.
- GitHub Releases API rate-limiting (403 silent fall-through in `highlights.tsx`) — Phase 1 acceptable.
- Design proposal to centralize brand constants — valid but too invasive for this PR.

Astro-Han:
Round 7 (final) crosscheck — both reviewers complete, no truncation. One commit `753357ad3` pushed:

**TUI brand sweep (user-visible strings the earlier sweep missed):**
- `routes/session/permission.tsx` × 3 — "until OpenCode is restarted" × 2 and "Tell OpenCode what to do differently"
- `feature-plugins/sidebar/footer.tsx` — "OpenCode includes free models" on the Getting Started footer
- `feature-plugins/home/tips-view.tsx` × 5 — prose tips describing OAuth auto-handle, auto-format, LSP, plugin example, and `opencode serve`

CLI command strings (`opencode run`, `opencode serve`, `opencode upgrade`, etc.) kept as-is because the binary name is still `opencode`; renaming that lives with the `@opencode-ai/*` package rename PR.

**Comments (no behavior change):**
- `oc-theme-preload.js` top-of-file note: the unconditional `color-scheme=light` write assumes pawwork is the only bundled light-locked theme; future dark-capable theme must add a guard. Tracked in [#23](https://github.com/Astro-Han/pawwork/issues/23).
- `config.ts` project-root cascade comment expanded to cross-reference `Filesystem.findUp({ rootFirst: true })` semantics so a future refactor can't silently flip innermost-wins.
- `config.ts` managed-dir cascade now carries the same "pawwork wins within directory" explanation that the project-root site had; the two sites were identical in logic but one was undocumented.

**Deferred to [#23](https://github.com/Astro-Han/pawwork/issues/23):** the color-scheme settings dropdown shows system/light/dark even though pawwork is light-locked. Pawwork-dark is next up; the dropdown becomes functional when that lands. Not patched in this PR to avoid throwaway interim UI.

Declined as out of scope / intentional:
- `summarizeBody` dropping `#`-only release bodies — test-confirmed intended behavior; real release bodies typically have content bullets under headings.
- `oc_locale` cookie name on the write path — kept aligned with the read-compat policy (changing would orphan existing users' language preference).

14 commits total on this branch, 7 crosscheck rounds passed.

Astro-Han:
Round 8 crosscheck results addressed in `b009a46e4` (11 files, +28 -144, net reduction):

**P1 defensive / diagnostic:**
- `oc-theme-preload.js`: wrap localStorage calls in try/catch so private-mode / blocked-storage browsers don't throw before the dataset attributes are set.
- `highlights.tsx`: log HTTP status on non-ok GitHub API response. Previously silent — 403 rate-limit, 304, 404, all dropped without a trace. Now visible in devtools.

**P2 brand routing:**
- `error-component.tsx` (TUI crash reporter): retargeted `github.com/anomalyco/opencode/issues/new` → `github.com/Astro-Han/pawwork/issues/new` so PawWork crash reports land in this repo, plus renamed the `opentui: fatal:` title prefix and `opencode-version` query param.
- `tui/app.tsx` Open docs, `dialog-custom-provider` Learn more, `settings-general` theme Learn more: all three `opencode.ai/docs/...` links retargeted to `github.com/Astro-Han/pawwork#readme`. README is the current docs source of truth (no standalone docs site yet).

**P2 remove OpenCode Go upsell (net −95 lines):**
- Deleted `dialog-go-upsell.tsx` entirely.
- Removed the `session.status` event handler in `routes/session/index.tsx` that showed the upsell + its kv dedup keys (`go_upsell_last_seen_at`, `go_upsell_dont_show`, `go_upsell_window`).
- Removed `SessionRetry.GO_UPSELL_MESSAGE` constant + the FreeUsageLimitError branch in `retry.ts`; the error now just returns a plain `"Free usage exceeded"`.
- PawWork doesn't integrate with or resell OpenCode Go; the upsell pointing to `opencode.ai/go` was misleading.

**P2 dev-only a11y (debug-bar):** pill + close button bumped from `opacity-40` → `opacity-60` plus `focus-visible:ring-2 focus-visible:ring-interactive-base`. Keyboard users can now see the pill and focus ring.

**P3 stale tips:** removed two `tips-view.tsx` entries referencing unshipped services — `/share` (publishes to opencode.ai) and `ghcr.io/anomalyco/opencode` docker image.

**Declined / intentional (in order of the review table):**
- Theme `context.tsx` `init` vs `onMount` duplication (P2): non-regression, single-threaded JS between those steps gives no observable flicker. Upstream pattern.
- `ids()` memoization (P2): single-theme today, O(1); deferred until a second theme ships.
- `summarizeBody` emoji surrogate boundary split (P2): real edge case but impact is cosmetic (half-emoji at position 200), release body first lines rarely contain emoji.
- debug-bar Show/fallback indent quirk (P3): pure style nit.

15 commits, 8 crosscheck rounds passed.

PR #98, opened 2026-04-21
fix: improve desktop menu updates and feedback

### LINKED ISSUE #84: [Bug] Desktop menu actions and labels need cleanup
### What happened?

The macOS desktop menu has several user-visible problems reported during internal testing. PawWork > Check for Updates does not surface visible feedback on click; the updater code path exists (`packages/desktop-electron/src/main/index.ts:353` → `autoUpdater.checkForUpdates()`), so the root cause is likely missing UI feedback, release metadata, or an updater gate condition — not a missing wire. PawWork > Reload Webview uses stale wording now that the shell is Electron-based. The desktop menu is still mostly English even when the app is used in Chinese. Help > Share Feedback and Help > Report a Bug currently open the same generic GitHub issue creation page, so the two entries do not lead to distinct workflows.

### Steps to reproduce

1. Open PawWork on macOS.
2. Open the PawWork application menu and click Check for Updates, then inspect Reload Webview.
3. Inspect the rest of the desktop menu while using the Chinese product flow.
4. Open Help and click Share Feedback, then Report a Bug.

Check for Updates should show visible feedback whether or not an update is available (dialog or status in the menu item). Reload Webview should use current Electron-era wording. The macOS menu should be localized for Chinese users where PawWork controls the labels. Share Feedback and Report a Bug should lead to distinct, appropriate destinations.

### PawWork version

v0.2.4

### OS version

macOS, exact version not captured in the internal report.

Yes, every time

Implementation pointer from local triage: `packages/desktop-electron/src/main/menu.ts` currently hardcodes English menu labels, while `packages/desktop-electron/src/renderer/i18n/{zh,en}.ts` already has some desktop menu translation keys. The updater wire itself is connected (`main/index.ts:353`), so first diagnose which of `menu-click handler → feedback surfacing → updater gate → release metadata` is actually swallowing the interaction before assuming a functional break.

Astro-Han:
## Summary

- Localize PawWork-controlled desktop menu labels and keep Electron-era wording.
- Add a Report a Problem flow that copies a diagnostic markdown report with active window context, recent logs, and session messages, while keeping Open GitHub Issue available.
- Enable the non-GitHub feedback path with a Feishu form URL injected through `PAWWORK_FEEDBACK_FORM_URL`.
- Make desktop update checks report visible results from menu, settings, and crash page paths.
- Preserve updater metadata across release finalization so macOS architectures do not clobber each other.

## Why

Fixes desktop menu cleanup and update feedback gaps from #84. Also gives users without GitHub accounts a lower-friction Feishu form based bug report path while preserving the GitHub issue entry for developer workflows.

This also addresses the real v0.2.5 release failure mode: the release published successfully, but `latest-mac.yml` ended up containing only `pawwork-mac-x64.zip`/`.dmg` and dropped the arm64 entries. The old workflow packaged each macOS architecture independently and did not run a final metadata merge/upload step. This PR adds that workflow step and hardens `finalize-latest-yml.ts` so existing, live, and current-run updater metadata are merged before `latest*.yml` is re-uploaded.

## Related Issue

Fixes #84

## How To Verify

```bash
cd packages/desktop-electron && bun test src/main/updater.test.ts src/main/menu-labels.test.ts src/main/menu.test.ts src/main/feedback.test.ts src/main/desktop-context-store.test.ts src/main/problem-report.test.ts src/main/updater-dialog-labels.test.ts src/renderer/i18n/index.test.ts scripts/release-metadata-contract.test.ts
cd packages/app && bun test src/utils/desktop-context.test.ts src/context/language.test.ts src/pages/error-update.test.ts
cd packages/opencode && bun test test/github/build-workflow.test.ts
cd packages/desktop-electron && bun run typecheck
cd packages/app && bun run typecheck
git diff --check origin/dev..HEAD
```

Manual/config checks:

- Confirmed v0.2.5 release assets exist, but current `latest-mac.yml` only includes x64 metadata, validating the release metadata clobbering path this PR fixes.
- Created and cleaned up the Feishu Base form `PawWork Problem Report` in the `CENO&纯刻用户服务` project knowledge space.
- Enabled form sharing and verified `shared=true`, `shared_limit=anyone_editable`.
- Set and read back GitHub repo variable `PAWWORK_FEEDBACK_FORM_URL`.

## Screenshots or Recordings

Not included. Changes are native menu/dialog flows plus release metadata behavior. The external Feishu form was created and configured separately, and its URL is supplied by repository variable at build time.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Desktop context sync with native windows; persisted menu locale and localized menus (EN/ZH).
  * “Report a Problem” flow with diagnostics export, clipboard copy, and configurable external feedback form.
  * New updater controller with explicit statuses, install/dismiss actions, and localized updater dialogs.
* **Bug Fixes**
  * Update-check now surfaces busy/disabled/failed states with targeted messages/toasts.
* **Tests**
  * Extensive tests for updater, menus, problem reports, feedback, desktop context, and metadata finalizer.
* **Chores**
  * CI/workflow: package version exposure, feedback URL injection, updated run grouping, and updater-metadata consolidation.


Astro-Han:
@coderabbitai review

.github/workflows/build.yml
```diff
@@ -297,19 +297,44 @@ jobs:
...
+
+          existing_dir="$RUNNER_TEMP/existing-latest-yml"
+          mkdir -p "$existing_dir"
+          tag="v${{ steps.package_version.outputs.version }}"
+
+          gh release download "$tag" --pattern latest-mac.yml --dir "$existing_dir" --repo "$GITHUB_REPOSITORY" || echo "No existing latest-mac.yml found; expected for first release."
+          gh release download "$tag" --pattern latest.yml --dir "$existing_dir" --repo "$GITHUB_REPOSITORY" || echo "No existing latest.yml found; expected for first release."
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 0a96c002b. The workflow now downloads only latest-mac.yml on macOS and only latest.yml on Windows, preserving the existing missing-metadata log messages.

packages/app/src/app.tsx
```diff
@@ -77,6 +79,12 @@ declare global {
     }
     api?: {
       setTitlebar?: (theme: { mode: "light" | "dark" }) => Promise<void>
+      setDesktopContext?: (context: {
+        directory: string | null
+        sessionID: string | null
+        route: string
+        locale: "en" | "zh"
+      }) => Promise<void>
```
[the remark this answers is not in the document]
Astro-Han: Fixed in aa00154d1. The app-side window.api declaration now imports and reuses the shared DesktopContext type from utils/desktop-context.

packages/app/src/pages/error-update.ts
```diff
@@ -0,0 +1,43 @@
...
+type Translator = (key: string, vars?: Record<string, string | number | boolean>) => string
+
+export function updateErrorPageState(result: UpdateInfo, t: Translator) {
+  if (result.updateAvailable && result.version) {
+    return {
+      version: result.version,
+      actionError: undefined,
+      actionMessage: undefined,
+    }
+  }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 13db177b8. The error update helper now checks status: "ready" explicitly, matching the UpdateInfo discriminated union.

packages/app/src/pages/session.tsx
```diff
@@ -332,9 +333,22 @@ export default function Page() {
...
+    if (!window.api?.setDesktopContext) return
+    void window.api.setDesktopContext(
+      buildDesktopContext({
+        directory: sdk.directory,
+        sessionID: params.id ?? null,
+        route: `${location.pathname}${location.search}${location.hash}`,
+        locale: language.locale(),
+      }),
+    )
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in aa00154d1. The session page setDesktopContext IPC call now attaches .catch(() => undefined) to avoid unhandled renderer rejections during navigation or teardown.

packages/desktop-electron/scripts/release-metadata-contract.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    const proc = Bun.spawn({
+      cmd: ["bun", "./scripts/finalize-latest-yml.ts"],
+      cwd: import.meta.dir.replace(/\/scripts$/, ""),
+      env: {
+        ...process.env,
+        GH_REPO: "Astro-Han/pawwork",
+        LATEST_YML_DIR: latestDir,
+        OPENCODE_VERSION: "0.2.4",
+        PATH: `${binDir}:${process.env.PATH ?? ""}`,
+        RUNNER_TEMP: runnerTemp,
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 0a96c002b. The release metadata contract test now derives package paths from import.meta.url, uses path.delimiter for PATH, and creates a platform-aware fake gh wrapper for Windows and POSIX.

packages/desktop-electron/src/main/feedback.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      ...overrides,
+    }),
+  }
+}
+
+describe("feedback handler", () => {
+  test("has localized confirmation labels for Simplified Chinese", () => {
+    expect(feedbackDialogLabels("zh").title).toBe("复制问题报告？")
+    expect(feedbackDialogLabels("zh").confirm).toBe("复制报告并打开表单")
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in aa00154d1. Added English feedback dialog label coverage alongside the Simplified Chinese assertion.

packages/desktop-electron/src/main/index.ts
```diff
@@ -338,81 +452,59 @@ function setupAutoUpdater() {
...
-    return { updateAvailable: true, version }
-  } catch (error) {
-    logger.error("update check failed", error)
-    return { updateAvailable: false, failed: true }
+  const result = await updater.check()
+  if (result.status === "ready") return { updateAvailable: true as const, status: result.status, version: result.version }
+  if (result.status === "failed") {
+    return { updateAvailable: false as const, status: result.status, reason: result.reason, message: result.message }
   }
+  return { updateAvailable: false as const, status: result.status }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 13db177b8. checkForUpdates no longer returns before checkUpdate when the updater is disabled, and the disabled status now flows to callers with localized user feedback.

packages/desktop-electron/src/main/updater.ts
```diff
@@ -0,0 +1,86 @@
...
+  return error instanceof Error ? error.message : String(error)
+}
+
+export function createUpdaterController(deps: Deps) {
+  let inflight: Promise<UpdateResult> | undefined
+  let updateReady = false
+
+  const run = async (): Promise<UpdateResult> => {
+    if (!deps.enabled) return { status: "disabled" }
+    updateReady = false
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 13db177b8. updateReady now remains latched across later checks and is cleared only when install() starts. Added a regression test for the downloaded-update case.

packages/desktop-electron/src/renderer/index.tsx
```diff
@@ -31,7 +31,20 @@ if (import.meta.env.DEV && !(root instanceof HTMLElement)) {
...
+let currentLocale: Locale = "en"
+void initI18n().then((locale) => {
+  currentLocale = locale
+  void window.api.setDesktopContext({
+    directory: null,
+    sessionID: null,
+    route: "/",
+    locale,
+  })
+})
```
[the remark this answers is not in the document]
Astro-Han: Fixed in aa00154d1. The desktop renderer initI18n bootstrap chain now returns setDesktopContext and has a final .catch(() => undefined).

packages/app/src/app.tsx
```diff
@@ -122,13 +125,33 @@ function RouterRoot(props: ParentProps<{ appChildren?: JSX.Element }>) {
...
+  createEffect(() => {
+    if (!window.api?.setDesktopContext) return
+    void window.api
+      .setDesktopContext(
+        buildDesktopContext({
+          route: `${location.pathname}${location.search}${location.hash}`,
+          locale: language.locale(),
+        }),
+      )
+      .catch(() => undefined)
```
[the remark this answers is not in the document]
Astro-Han: Fixed in edc4e7dfc. The root DesktopContextRouteBridge now skips session routes so the session page remains the single writer for directory/sessionID while it is active.

packages/desktop-electron/src/main/feedback.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const report = buildProblemReport({
+      diagnostics: deps.diagnostics(context),
+      logTail: deps.logTail(),
+      sessionExport,
+    })
+
+    await deps.copy(report.markdown)
+    await deps.openExternal(deps.feedbackUrl)
+  }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in edc4e7dfc. The feedback handler now wraps the full flow, reports non-session failures through onError, and has coverage for a copy failure that no longer rejects or opens the form.

packages/desktop-electron/src/main/index.ts
```diff
@@ -259,6 +357,22 @@ registerIpcHandlers({
...
+          desktopContexts.delete(win.id)
+          contextWindowCleanup.delete(win.id)
+        })
+      }
+    }
+    if (context.locale !== menuLocale) {
+      menuLocale = context.locale
+      wireMenu()
+    }
+  },
```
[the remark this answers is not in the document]
Astro-Han: Fixed in edc4e7dfc. setDesktopContext now normalizes renderer-provided IPC payloads before storing them or updating menuLocale, with unsupported locale values falling back to en.

packages/desktop-electron/src/main/problem-report.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  while (bytes(output) > maxBytes && logTail.length > 0) {
+    const remove = Math.max(1, Math.ceil(logTail.length / 2))
+    omittedLogBytes += bytes(logTail.slice(0, remove))
+    logTail = logTail.slice(remove)
+    output = markdown(makePayload())
+  }
+
+  return { markdown: output }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in edc4e7dfc. Problem report truncation now also omits oversized session info and compresses diagnostics before enforcing maxBytes, with a final hard failure if the contract still cannot be met.

packages/desktop-electron/src/main/updater.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        return { isUpdateAvailable: false, updateInfo: { version: "0.2.4", files: [] } }
+      },
+    })
+
+    await expect(setup.updater.check()).resolves.toEqual({ status: "ready", version: "0.2.5" })
+    await expect(setup.updater.check()).resolves.toEqual({ status: "none" })
+    expect(setup.updater.install()).toBe(true)
+    expect(setup.calls.install).toBe(1)
+    expect(setup.updater.install()).toBe(false)
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in a9a4e9487. The updater controller now keeps readyVersion latched until install() starts, and the regression test expects subsequent checks to keep returning the ready state.

.github/workflows/build.yml
```diff
@@ -297,19 +297,48 @@ jobs:
...
+
+          existing_dir="$RUNNER_TEMP/existing-latest-yml"
+          mkdir -p "$existing_dir"
+          tag="v${{ steps.package_version.outputs.version }}"
+
+          if [[ "$RUNNER_OS" == "macOS" ]]; then
+            gh release download "$tag" --pattern latest-mac.yml --dir "$existing_dir" --repo "$GITHUB_REPOSITORY" || echo "No existing latest-mac.yml found; expected for first release."
+          fi
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            gh release download "$tag" --pattern latest.yml --dir "$existing_dir" --repo "$GITHUB_REPOSITORY" || echo "No existing latest.yml found; expected for first release."
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 359f10d0d. The workflow now uses download_or_warn, only continuing for explicit missing-asset messages and failing the job for auth, API, or transient download errors.

packages/desktop-electron/src/main/index.ts
```diff
@@ -259,6 +368,23 @@ registerIpcHandlers({
...
+        contextWindowCleanup.add(win.id)
+        win.once("closed", () => {
+          desktopContexts.delete(win.id)
+          contextWindowCleanup.delete(win.id)
+        })
+      }
+    }
+    if (next.locale !== menuLocale) {
+      menuLocale = next.locale
+      wireMenu()
```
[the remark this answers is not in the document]
Astro-Han: Fixed in dc6c50e4d. setDesktopContext now persists locale changes through writeStoredMenuLocale, using the same opencode.global.dat language store read at startup.

packages/desktop-electron/src/main/updater-dialog-labels.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    expect(labels.ready.message("0.2.5")).toBe("更新 0.2.5 已下载。现在重启？")
+    expect(labels.ready.buttons).toEqual(["重启", "稍后"])
+  })
+
+  test("falls back to English labels", () => {
+    const labels = updaterDialogLabels("en")
+
+    expect(labels.busy.title).toBe("Update Check in Progress")
+    expect(labels.ready.message(undefined)).toBe("Update downloaded. Restart now?")
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in cd6973871. Renamed the test to "returns English labels" so the test name matches the explicit en locale path it exercises.

packages/desktop-electron/src/main/updater.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        throw new Error("download failed")
+      },
+    })
+    await expect(setup.updater.check()).resolves.toEqual({
+      status: "failed",
+      reason: "download",
+      message: "download failed",
+    })
+  })
+})
```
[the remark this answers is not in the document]
Astro-Han: Fixed in cd6973871. Added a disabled-controller regression test that expects status: "disabled" and verifies no update check, download, or install side effects occur.

packages/desktop-electron/src/main/index.ts
```diff
@@ -59,11 +65,111 @@ let initStep: InitStep = { phase: "server_waiting" }
...
 const loadingComplete = defer<void>()
+let menuLocale: MenuLocale = readStoredMenuLocale(app.getLocale())
+const defaultDesktopContext: DesktopContext = {
+  directory: null,
+  sessionID: null,
+  route: "/",
+  locale: menuLocale,
+}
+const desktopContexts = createDesktopContextStore(defaultDesktopContext)
+const contextWindowCleanup = new Set<number>()
```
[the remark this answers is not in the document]
Astro-Han: Fixed in f74deaef1. The desktop context store now takes a dynamic fallback, so when no window context exists it reads the current menuLocale instead of a frozen startup locale.

packages/desktop-electron/src/main/ipc.ts
```diff
@@ -27,10 +34,21 @@ type Deps = {
...
+    | { updateAvailable: false; status: "disabled" | "none" | "busy"; version?: undefined }
+    | { updateAvailable: true; status: "ready"; version: string }
+    | {
+        updateAvailable: false
+        status: "failed"
+        reason: "check" | "download" | "metadata" | "unsupported"
+        message: string
+        version?: undefined
+      }
+  >
```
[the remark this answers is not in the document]
Astro-Han: Addressed in 91fac2fc9 without importing from the app package. I extracted UpdateInfo into the desktop preload types and reused it in ipc.ts, which avoids a desktop main to app package dependency while removing the inline IPC union duplication.

packages/desktop-electron/src/main/menu.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      deps,
+      locale: "en",
+      updaterEnabled: true,
+      feedbackEnabled: true,
+    })
+    const appMenu = template[0]?.submenu ?? []
+
+    expect(appMenu.some((item) => item.label === "Reload Window")).toBe(true)
+    expect(appMenu.some((item) => item.label === "Reload Webview")).toBe(false)
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 01a44278c. The menu tests now locate the PawWork and Help menus by label instead of relying on template[0] and template.at(-1).

packages/desktop-electron/src/main/problem-report.ts
```diff
@@ -0,0 +1,171 @@
...
+  if (bytes(output) > maxBytes) {
+    throw new Error("Problem report exceeds maxBytes after truncation")
+  }
+
+  return { markdown: output }
+}
+
+export function parseProblemReportPayload(input: string): Payload {
+  const match = input.match(/```json\n([\s\S]*?)\n```/)
+  if (!match) throw new Error("Problem report JSON block not found")
```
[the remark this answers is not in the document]
Astro-Han: Fixed in b911ca7fb. parseProblemReportPayload now accepts LF and CRLF fenced JSON, with a regression test for CRLF reports.

Astro-Han:
@coderabbitai review

.github/workflows/build.yml
```diff
@@ -67,8 +67,8 @@ on:
         type: string
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.phase || 'submit' }}-${{ inputs.channel || 'dev' }}-${{ inputs.target || 'macos' }}-${{ inputs.arch || 'arm64' }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ inputs.source_ref || github.ref_name }}-${{ inputs.phase || 'submit' }}-${{ inputs.channel || 'dev' }}-${{ inputs.target || 'macos' }}
+  cancel-in-progress: false
```
Astro-Han, line 73: **P1 — confirm intent of the concurrency change.**

Two changes compound here: `arch` was dropped from the group key (line 70), and `cancel-in-progress` flipped from `true` to `false`. The combined semantics: for the same `ref + phase + target`, macOS `arm64` and `x64` now share a single concurrency group AND queue instead of cancelling each other.

Side effect: when `phase=full` dispatches both arches, they will run serially rather than in parallel — full-build wall time doubles.

If the goal is to prevent the `Finalize updater metadata` read-modify-write race, this is an intentional belt-and-suspenders on top of the download-merge logic. If you expected the download-merge to be sufficient on its own, then the serialization here is redundant and `cancel-in-progress: true` with `arch` back in the key would restore parallelism. Please confirm which one you meant.

packages/desktop-electron/src/main/menu-i18n.ts
```diff
@@ -0,0 +1,14 @@
+import { getStore } from "./store"
+import { detectSystemMenuLocale, parseMenuLocale, type MenuLocale } from "./menu-labels"
+
+export function readStoredMenuLocale(systemLocale: string | null | undefined): MenuLocale {
+  const raw = getStore("opencode.global.dat").get("language")
+  const stored = parseMenuLocale(raw)
+  // Keep an explicit stored locale, including English; otherwise detect from the OS locale.
+  if (stored !== "en" || raw) return stored
```
Astro-Han: **P2 — simplify.** `if (stored !== "en" || raw) return stored` collapses to a cleaner form:

- `raw` truthy → use `stored`
- `raw` falsy + `stored === "zh"` → impossible (when `raw` is falsy, `parseMenuLocale` always returns `"en"`)
- `raw` falsy + `stored === "en"` → fall through to `detectSystemMenuLocale`

So the body is equivalent to:

```ts
if (raw) return stored
return detectSystemMenuLocale(systemLocale)
```

That matches the comment above the line verbatim and removes the subtle `!==` reasoning.

packages/desktop-electron/src/preload/types.ts
```diff
@@ -15,6 +15,19 @@ export type TitlebarTheme = {
...
+  sessionID: string | null
+  route: string
+  locale: "en" | "zh"
+}
+
+type UpdateFailureReason = "check" | "download" | "metadata" | "unsupported"
+export type UpdateInfo =
+  | { updateAvailable: false; status: "disabled" | "none" | "busy"; version?: undefined }
+  | { updateAvailable: true; status: "ready"; version: string }
+  | { updateAvailable: false; status: "failed"; reason: UpdateFailureReason; message: string; version?: undefined }
```
Astro-Han: **P2 — `UpdateInfo` is declared twice with the same shape.** The identical union lives in `packages/app/src/context/platform.tsx:10-13`. Since preload is a product boundary, duplicating here is defensible, but the literal strings (`status`, `reason`) are exactly the kind of thing that drifts silently.

Suggest one of:
- Re-export this type from a shared package and import it on both sides.
- Keep two declarations but add a `// keep in sync with packages/app/src/context/platform.tsx` comment on both.

packages/desktop-electron/src/main/updater.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function createUpdaterController(deps: Deps) {
+  let inflight: Promise<UpdateResult> | undefined
+  let updateReady = false
+  let readyVersion: string | undefined
+
+  const run = async (): Promise<UpdateResult> => {
+    if (!deps.enabled) return { status: "disabled" }
+    if (updateReady && readyVersion) return { status: "ready", version: readyVersion }
```
Astro-Han: **P2 — `updateReady` never invalidates until `install()` runs.** Once `run()` sets `updateReady = true`, every subsequent `check()` short-circuits here and returns the cached `{ status: "ready", version: readyVersion }` without re-fetching.

User-visible effect: after `check` once surfaces a ready update and the user presses **Later** in the dialog, clicking **Check for Updates** again is a silent no-op (no network, no toast) until the app is restarted. From the menu it looks like the click did nothing.

Options:
- Clear `updateReady`/`readyVersion` when the user picks **Later** (caller-side).
- Add a TTL so the cache expires after N minutes and the next `check()` re-fetches.
- At minimum, log the short-circuit so the silence is debuggable.

packages/desktop-electron/src/main/index.ts
```diff
@@ -59,11 +65,111 @@ let initStep: InitStep = { phase: "server_waiting" }
...
+      defaultId: 0,
+      cancelId: 1,
+    })
+    return response.response === 0
+  },
+  copy: (value) => clipboard.writeText(value),
+  openExternal: (url) => shell.openExternal(url),
+  diagnostics: (context) => diagnostics((context as DesktopContext | undefined) ?? currentDesktopContext()),
+  logTail: tail,
+  sessionExport: (context) => sessionExport((context as DesktopContext | undefined) ?? currentDesktopContext()),
```
Astro-Han: **P3 — dead fallback on lines 168 and 170.** Both `diagnostics: (context) => diagnostics((context as DesktopContext | undefined) ?? currentDesktopContext())` and the same pattern for `sessionExport` can never hit the `?? currentDesktopContext()` branch.

`feedback.ts::createFeedbackHandler` resolves `context = deps.context?.()` before `confirm` and always passes that value down to `diagnostics`/`sessionExport`. Since `deps.context = currentDesktopContext` is set here, `context` is never undefined at call time.

The fallback reads as defensive but it is dead, and it misleads readers into thinking context could be missing. Drop the coalesce: `diagnostics: (context) => diagnostics(context as DesktopContext)`.

packages/desktop-electron/src/renderer/index.tsx
```diff
@@ -31,7 +31,22 @@ if (import.meta.env.DEV && !(root instanceof HTMLElement)) {
...
+    return window.api.setDesktopContext({
+      directory: null,
+      sessionID: null,
+      route: "/",
+      locale,
+    })
+  })
+  .catch(() => undefined)
+
+export function desktopLocale() {
```
Astro-Han: **P3 — `desktopLocale()` export has no visible consumer in this diff.** If nothing imports it, demote to module-local (`function desktopLocale()` without `export`) or delete. If it is intended for a future call site, a one-line comment pointing at who will use it would keep the export honest.

packages/app/src/pages/session.tsx
```diff
@@ -332,9 +333,24 @@ export default function Page() {
   const prompt = usePrompt()
   const comments = useComments()
   const terminal = useTerminal()
+  const location = useLocation()
   const [searchParams, setSearchParams] = useSearchParams<{ prompt?: string }>()
   const { params, sessionKey, tabs, view } = useSessionLayout()
 
+  createEffect(() => {
```
Astro-Han, line 372: **P3 — this `createEffect` fires on every route change inside the session page.** It tracks `sdk.directory`, `params.id`, `location.pathname + search + hash`, and `language.locale()`. Fast tab switching or search-param updates will trigger one IPC per change.

`setDesktopContext` is fire-and-forget so this does not block the UI, but on the main side it runs the full handler (store update + menu rebuild when locale changes). If coalescing isn't already done in main, consider either:
- Last-value dedupe inside the main-process handler, or
- A small debounce here, or
- Track a derived string once and only invoke on change.

Low priority — flagging because route churn is common in session view.

Astro-Han (commented):
Overall quality is high and most Codex/CodeRabbit feedback has already been absorbed. No P0 findings. Posting 7 inline comments: 1 P1 to confirm (concurrency trade-off), plus P2/P3 readability and dead-code items.

Praise worth calling out:
- `problem-report.ts` has clean staged truncation (messages → logTail → sessionInfo → diagnostics) against a byte budget, with unit tests covering each truncation branch.
- `release-metadata-contract.test.ts` uses a fake-gh shim to black-box verify the finalize script across platforms — good template.
- Already-absorbed review items: `constants.ts` trim, `desktop-context-store.ts` Map re-insertion order, `parseProblemReportPayload` non-greedy + CRLF, `finalize-latest-yml.ts` missing-asset vs real-error split, `updater.ts` Deps.checkForUpdates allowing null.

Only P1 needs confirmation before merge.

Astro-Han:
@coderabbitai review

.github/workflows/build.yml
Astro-Han: Confirmed intentional. I kept same target and phase runs queued across arch because latest*.yml release metadata is still a read modify write upload to the same release asset. Added a workflow comment so this tradeoff is explicit.

packages/desktop-electron/src/main/menu-i18n.ts
Astro-Han: Fixed. Simplified this to `if (raw) return stored`, which matches the comment and removes the subtle `stored !== "en"` reasoning.

packages/desktop-electron/src/preload/types.ts
Astro-Han: Addressed with sync comments on both declarations. I kept the duplicate type declarations because preload is a product boundary, but made the drift risk explicit.

packages/desktop-electron/src/main/updater.ts
Astro-Han: Fixed with the caller side option. The updater controller now exposes `dismissReady()`, and `checkForUpdates()` calls it when the user chooses Later. Added a regression test that confirms a dismissed ready update can be checked again.

packages/desktop-electron/src/main/index.ts
Astro-Han: Fixed. Removed the dead `?? currentDesktopContext()` fallback and now pass the captured `DesktopContext` directly into diagnostics and session export.

packages/desktop-electron/src/renderer/index.tsx
Astro-Han: Fixed. Removed the unused `desktopLocale()` export and the now unnecessary `currentLocale` state.

packages/app/src/pages/session.tsx
Astro-Han: Fixed. Added last value dedupe in the session page before calling `setDesktopContext`, so route churn only sends IPC when the effective context changes.

Astro-Han:
@coderabbitai review

Astro-Han:
@coderabbitai review

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -0,0 +1,111 @@
...
+    return JSON.parse(value) as unknown
+  } catch {
+    return value
+  }
+}
+
+export function parseMenuLocale(value: unknown): MenuLocale {
+  const parsed = parseStored(value)
+  if (typeof parsed === "string") {
+    const normalized = parsed.toLowerCase().replace("_", "-")
```
Astro-Han: `replace("_", "-")` only replaces the first underscore. For locale strings like `zh_Hant_TW`, this produces `zh-Hant_TW` instead of `zh-Hant-TW`. Consider using `replaceAll("_", "-")`.

packages/desktop-electron/src/main/feedback.ts
```diff
@@ -0,0 +1,68 @@
...
+    zh: {
+      title: "复制问题报告？",
+      message:
+        "PawWork 会复制一份问题报告到剪贴板，并打开反馈表单。报告可能包含应用诊断信息、最近应用日志、当前会话消息、工具输出、文件名、路径和文件片段。提交前请先检查。",
+      confirm: "复制报告并打开表单",
+      cancel: "取消",
+    },
+  } satisfies Record<MenuLocale, { title: string; message: string; confirm: string; cancel: string }>
+
+  return labels[locale]
```
Astro-Han: `labels[locale]` returns `undefined` at runtime if `locale` is not exactly `"en"` or `"zh"`. Consider adding a fallback like `labels[locale] ?? labels.en` for safety, similar to `updaterDialogLabels`.

Astro-Han:
@coderabbitai review

packages/desktop-electron/src/main/updater.ts
```diff
@@ -0,0 +1,97 @@
...
```
Astro-Han: `updateReady && readyVersion` uses truthiness, so an empty-string version (`""`) would be treated as not ready even though `updateReady` is `true`. If `readyVersion` should never be empty, consider asserting that invariant or using `readyVersion !== undefined` instead.

packages/desktop-electron/src/main/index.ts
```diff
@@ -68,11 +74,111 @@ let mainWindow: BrowserWindow | null = null
...
+      title: labels.title,
+      message: labels.message,
+      buttons: [labels.confirm, labels.cancel],
+      defaultId: 0,
+      cancelId: 1,
+    })
+    return response.response === 0
+  },
+  copy: (value) => clipboard.writeText(value),
+  openExternal: (url) => shell.openExternal(url),
```
Astro-Han: `shell.openExternal(url)` returns `Promise<string>`, but `FeedbackDeps.openExternal` expects `Promise<void> | void`. This type mismatch may cause issues under strict TypeScript checks. Consider wrapping it with `void` like in `menu.ts`: `openExternal: (url) => { void shell.openExternal(url) }`.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. Switched to `replaceAll("_", "-")` and added `zh_Hant_TW` coverage.

packages/desktop-electron/src/main/feedback.ts
Astro-Han: Fixed. Added `labels[locale] ?? labels.en` and a runtime fallback test.

packages/desktop-electron/src/main/updater.ts
Astro-Han: Fixed. The ready latch now checks `readyVersion !== undefined` instead of truthiness.

packages/desktop-electron/src/main/index.ts
Astro-Han: Fixed. Wrapped `shell.openExternal(url)` in a void block to match the feedback dependency contract.

packages/app/src/app.tsx
```diff
@@ -122,13 +125,38 @@ function RouterRoot(props: ParentProps<{ appChildren?: JSX.Element }>) {
...
+  })
+
+  return null
+}
+
+function isSessionRoute(pathname: string) {
+  return /(^|\/)session(\/|$)/.test(pathname)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Narrowed the guard to the nested `/:dir/session/:id?` shape by checking path segments instead of any `session` segment.

packages/app/src/pages/error-update.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  test("shows failed feedback", () => {
+    expect(
+      updateErrorPageState({ updateAvailable: false, status: "failed", reason: "check", message: "network down" }, t),
+    ).toEqual({
+      version: undefined,
+      actionError: "network down",
+      actionMessage: undefined,
+    })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Added the empty message failed state test to pin the translated fallback.

packages/desktop-electron/scripts/release-metadata-contract.test.ts
```diff
@@ -0,0 +1,207 @@
...
+        ...process.env,
+        GH_REPO: "Astro-Han/pawwork",
+        LATEST_YML_DIR: latestDir,
+        OPENCODE_VERSION: "0.2.4",
+        PATH: `${binDir}${delimiter}${process.env.PATH ?? ""}`,
+        RUNNER_TEMP: runnerTemp,
+      },
+      stderr: "pipe",
+      stdout: "pipe",
+    })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Extracted the repeated finalizer process wiring into `spawnFinalizer()`.

packages/desktop-electron/src/main/index.ts
```diff
@@ -68,11 +74,111 @@ let mainWindow: BrowserWindow | null = null
...
+    directory: context.directory,
+    sessionID: context.sessionID,
+    logPath: filePath(),
+  }
+}
+
+async function sessionExport(context = currentDesktopContext()) {
+  if (!context.sessionID) return { status: "none" as const }
+  const ready = await serverReady.promise
+  const url = new URL(`/session/${context.sessionID}/message`, ready.url)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `sessionExport()` now uses `encodeURIComponent(context.sessionID)` before building the session messages URL.

packages/desktop-electron/src/main/ipc.ts
```diff
@@ -64,6 +73,9 @@ export function registerIpcHandlers(deps: Deps) {
   ipcMain.handle("report-deep-link-ready", (event: IpcMainInvokeEvent) =>
     deps.reportDeepLinkReady(BrowserWindow.fromWebContents(event.sender)),
   )
+  ipcMain.handle("set-desktop-context", (event: IpcMainInvokeEvent, context: DesktopContext) =>
+    deps.setDesktopContext(context, BrowserWindow.fromWebContents(event.sender)),
+  )
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `set-desktop-context` now returns early when the sender no longer maps to a `BrowserWindow`, and the dependency type now requires a real window.

packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -0,0 +1,131 @@
...
+  if (feedbackEnabled) {
+    helpSubmenu.push({ label: t("reportProblem"), click: () => deps.reportProblem() })
+  }
+
+  helpSubmenu.push({ label: t("openGithubIssue"), click: () => deps.openExternal(githubIssueUrl) })
+
+  return [
+    {
+      label: "PawWork",
+      submenu: [
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `buildMenuTemplate()` now accepts `appName`, and `createMenu()` passes `app.getName()` so Dev and Beta menu labels match runtime app name.

packages/desktop-electron/src/main/menu.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    const template = buildMenuTemplate({
+      deps,
+      locale: "en",
+      updaterEnabled: true,
+      feedbackEnabled: false,
+    })
+    const help = submenu(template, "Help")
+
+    expect(help.some((item) => item.label === "Report a Problem")).toBe(false)
+    expect(help.some((item) => item.label === "Open GitHub Issue")).toBe(true)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Added a positive `feedbackEnabled: true` menu test that asserts `Report a Problem` is present.

packages/desktop-electron/src/main/updater.ts
```diff
@@ -0,0 +1,97 @@
...
+        inflight = undefined
+      })
+      return inflight
+    },
+    install() {
+      if (!updateReady) return false
+      updateReady = false
+      readyVersion = undefined
+      deps.quitAndInstall()
+      return true
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `install()` now clears the ready latch only after `quitAndInstall()` returns, with a regression test for synchronous throw preserving ready state.

packages/desktop-electron/src/preload/types.ts
```diff
@@ -15,6 +15,20 @@ export type TitlebarTheme = {
...
 
+export type DesktopContext = {
+  directory: string | null
+  sessionID: string | null
+  route: string
+  locale: "en" | "zh"
+}
+
+type UpdateFailureReason = "check" | "download" | "metadata" | "unsupported"
+// Keep in sync with packages/app/src/context/platform.tsx.
```
Astro-Han: **Type sync risk**: `UpdateInfo` is defined in two places (`packages/app/src/context/platform.tsx` and here). The comment says to keep them in sync, but there's no automated verification.

**Suggestion**: Either:
1. Export `UpdateInfo` from a shared location
2. Add a test that asserts type compatibility

packages/desktop-electron/src/main/feedback.ts
```diff
@@ -0,0 +1,68 @@
...
+
+function errorMessage(error: unknown) {
+  return error instanceof Error ? error.message : String(error)
+}
+
+export function feedbackDialogLabels(locale: MenuLocale) {
+  const labels = {
+    en: {
+      title: "Copy problem report?",
+      message:
```
Astro-Han, line 21: **Privacy consideration**: The message warns users about session messages and file names, but paths like `/Users/name/private-project` contain usernames that users may overlook.

**Suggestion**: Add explicit mention of username exposure in the warning, or truncate usernames in the diagnostics function.

Astro-Han (commented):
**Review: Approve** with minor suggestions.

PR architecture is solid with clear separation of concerns (menu-labels.ts + menu-template.ts + menu-i18n.ts). Tests cover key behaviors. CI all passing.

Two suggestions below for type sync and user privacy.

packages/desktop-electron/src/preload/types.ts
Astro-Han: Fixed with option 1. `UpdateInfo` now has a single source in `packages/app/src/context/platform.tsx`, and preload imports/re-exports that type, so desktop typecheck catches drift.

packages/desktop-electron/src/main/feedback.ts
Astro-Han: Fixed by making the confirmation warning explicit that paths may include the user system username. I kept diagnostics unredacted because full paths are useful for debugging and the user gets a clipboard review step before submission.

packages/desktop-electron/scripts/finalize-latest-yml.ts
```diff
@@ -76,18 +77,78 @@ async function read(subdir: string, filename: string): Promise<LatestYml | undef
...
+  const existingDir = path.join(tmp, "existing-latest-yml")
+  await mkdir(existingDir, { recursive: true })
+  try {
+    await $`gh release download ${tag} --pattern ${filename} --dir ${existingDir} --repo ${repo}`.quiet()
+  } catch (error) {
+    const message = shellErrorText(error)
+    if (isMissingAssetError(message)) return undefined
+    throw new Error(`Failed to download existing ${filename}: ${message}`)
+  }
+  return readFile(path.join(existingDir, filename))
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `downloadExisting()` now reads the workflow snapshot only as a fallback, first attempts to download the live release asset, and falls back to the cached snapshot only for missing-asset errors. Added tests for preferring live metadata and for cached fallback when live metadata is missing.

packages/desktop-electron/src/preload/types.ts
```diff
@@ -15,6 +19,13 @@ export type TitlebarTheme = {
   mode: "light" | "dark"
 }
 
+export type DesktopContext = {
+  directory: string | null
+  sessionID: string | null
+  route: string
+  locale: "en" | "zh"
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Preload now imports and re-exports the app-side `DesktopContext` type from `packages/app/src/utils/desktop-context.ts`, matching the earlier `UpdateInfo` single-source pattern.

Astro-Han:
@coderabbitai review

packages/app/src/pages/session.tsx
```diff
@@ -333,8 +334,43 @@ export default function Page() {
...
+          pendingDesktopContext = ""
+        }
+      })
+      .catch(() => {
+        setTimeout(() => {
+          if (pendingDesktopContext !== serialized || lastDesktopContext === serialized) return
+          syncDesktopContext(context, serialized)
+        }, 250)
+      })
+  }
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Desktop-context retry timers are now tracked and cleared on cleanup; retries also check a disposed flag before resending context, so a stale session page cannot overwrite a newer route/window context after unmount.

Astro-Han:
@coderabbitai review

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -0,0 +1,111 @@
...
+    toggleTerminal: "切换终端",
+    toggleFileTree: "切换文件树",
+    back: "后退",
+    forward: "前进",
+    previousSession: "上一个会话",
+    nextSession: "下一个会话",
+    previousProject: "上一个项目",
+    nextProject: "下一个项目",
+    pawworkOnGithub: "PawWork GitHub",
+    reportProblem: "报告问题",
```
Astro-Han, line 76: **Missing "on" in Chinese translation**

English: `pawworkOnGithub: "PawWork on GitHub"` (line 51)
Chinese: `pawworkOnGithub: "PawWork GitHub"` (this line)

The Chinese translation drops the "on" preposition. Consider using `"PawWork 在 GitHub"` for consistency.

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+}
+
+export function parseMenuLocale(value: unknown): MenuLocale {
+  const parsed = parseStored(value)
+  if (typeof parsed === "string") {
+    const normalized = parsed.toLowerCase().replaceAll("_", "-")
+    if (normalized.startsWith("zh")) return "zh"
+    if (normalized === "en") return "en"
+  }
```
Astro-Han, line 96: **Locale detection inconsistency for English variants**

Line 95 uses `startsWith("zh")` for Chinese (matches `"zh-CN"`, `"zh-TW"`).
Line 96 uses `=== "en"` for English (does NOT match `"en-US"`, `"en-GB"`).

Both fall through correctly, but logic is asymmetric. Consider `startsWith("en")` for symmetry.

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -0,0 +1,111 @@
...
+  return "en"
+}
+
+export function detectSystemMenuLocale(locale: string | null | undefined): MenuLocale {
+  if (locale?.toLowerCase().startsWith("zh")) return "zh"
+  return "en"
+}
+
+export function menuLabel(locale: MenuLocale, key: MenuLabelKey) {
+  return labels[locale][key]
```
Astro-Han: **Runtime safety for menuLabel function**

Direct access `labels[locale][key]` without fallback. If invalid locale passed at runtime (type assertion edge case), this throws.

Consider defensive fallback:
```typescript
return labels[locale]?.[key] ?? labels.en[key] ?? key
```

packages/desktop-electron/scripts/finalize-latest-yml.ts
```diff
@@ -76,18 +77,76 @@ async function read(subdir: string, filename: string): Promise<LatestYml | undef
...
 }
 
+async function readFile(filepath: string): Promise<LatestYml | undefined> {
+  const file = Bun.file(filepath)
+  if (!(await file.exists())) return undefined
+  return parse(await file.text())
+}
+
+function mergeLatest(...items: Array<LatestYml | undefined>): LatestYml | undefined {
+  const present = items.filter((item): item is LatestYml => Boolean(item))
```
Astro-Han, line 87: **mergeLatest base selection naming/clarification**

`const base = present.at(-1)!` means "last wins" for metadata. The function name `mergeLatest` doesn't clearly communicate this.

Consider adding comment: `// Use last item metadata; live data wins over cached snapshots`

packages/desktop-electron/src/main/desktop-context-store.ts
```diff
@@ -0,0 +1,23 @@
...
+    set(windowID: number, context: DesktopContext) {
+      contexts.delete(windowID)
+      contexts.set(windowID, context)
+      latest = context
+    },
+    delete(windowID: number) {
+      const removed = contexts.get(windowID)
+      contexts.delete(windowID)
+      if (removed === latest) latest = [...contexts.values()].at(-1)
+    },
```
Astro-Han, line 17: **latest naming clarity**

Variable `latest` tracks chronological *set* order, not semantic "latest/best". When removed===latest, fallback finds next most-recently-set context.

Consider renaming to `lastSet` or `mostRecent` to clarify this distinction.

packages/desktop-electron/src/main/feedback.ts
```diff
@@ -0,0 +1,68 @@
...
+      title: "复制问题报告？",
+      message:
+        "PawWork 会复制一份问题报告到剪贴板，并打开反馈表单。报告可能包含应用诊断信息、最近应用日志、当前会话消息、工具输出、文件名、包含系统用户名的路径和文件片段。提交前请先检查。",
+      confirm: "复制报告并打开表单",
+      cancel: "取消",
+    },
+  } satisfies Record<MenuLocale, { title: string; message: string; confirm: string; cancel: string }>
+
+  return labels[locale] ?? labels.en
+}
```
Astro-Han, line 44: **Redundant-but-intentional fallback**

`?? labels.en` is technically redundant given `MenuLocale` type (only "en" or "zh"), but intentional for runtime safety. Test at feedback.test.ts:1290 explicitly tests `"fr" as never`.

Consider documenting with comment: `// Runtime fallback for unexpected locale values`

packages/desktop-electron/src/main/updater-dialog-labels.ts
```diff
@@ -0,0 +1,62 @@
...
+      title: "更新已准备好",
+      message: (version) => (version ? `更新 ${version} 已下载。现在重启？` : "更新已下载。现在重启？"),
+      buttons: ["重启", "稍后"],
+    },
+  },
+}
+
+export function updaterDialogLabels(locale: MenuLocale) {
+  return labels[locale] ?? labels.en
+}
```
Astro-Han, line 63: **Same redundant fallback pattern**

Same observation as feedback.ts:39 - `?? labels.en` is redundant given `MenuLocale` type but intentional for runtime safety. Consider adding similar comment documenting the intentional redundancy.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. Updated the Chinese label to PawWork 在 GitHub so it matches the English intent.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. parseMenuLocale now accepts English locale variants with startsWith("en"), and the behavior is covered with an en-US test case.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. menuLabel now falls back through labels.en and then the key for unexpected runtime values, with a regression assertion for an invalid locale.

packages/desktop-electron/scripts/finalize-latest-yml.ts
Astro-Han: Fixed. Added a short comment at the base selection explaining that the last item is the metadata base and that live data wins over cached snapshots.

packages/desktop-electron/src/main/desktop-context-store.ts
Astro-Han: Fixed. Renamed the internal tracker from latest to mostRecent and updated the fallback test name to match the dynamic fallback behavior.

packages/desktop-electron/src/main/feedback.ts
Astro-Han: Fixed. Added a comment documenting that the English fallback is intentional runtime protection for unexpected locale values crossing process boundaries.

packages/desktop-electron/src/main/updater-dialog-labels.ts
Astro-Han: Fixed. Added the same runtime-fallback comment to updaterDialogLabels for consistency.

Astro-Han:
@coderabbitai review

Astro-Han (commented):
## Review

A few nits and questions:

### 1. `updater.ts` — `install()` returns `boolean` but callers ignore it

`install()` returns `true`/`false` to indicate whether it actually triggered the install, but `index.ts:installUpdate()` just calls `updater.install()` and discards the result. Consider either using the return value or changing the signature to `void`.

### 2. `menu-template.ts` — `checkForUpdates` menu item lost `enabled: UPDATER_ENABLED`

The old menu had `enabled: UPDATER_ENABLED` on the \"Check for Updates...\" item. The new template always enables it. If the updater is disabled, clicking it will silently do nothing (the handler in `index.ts` still guards with `UPDATER_ENABLED`, but the menu item is no longer greyed out). Was this intentional?

### 3. `problem-report.ts` — `truncateString` is a no-op for the first 1024 chars

`truncateString(value, 1024)` returns `value` unchanged when `value.length <= 1024`. The first iteration of the diagnostics-truncation loop therefore does nothing except increment `omittedDiagnosticsBytes` to 0. You could start `diagnosticStringLimit` at 512 to skip the wasted iteration.

### 4. `index.ts` — `sessionExport` mixes `fetch` timeout with `AbortController`

The 10-second timeout uses `AbortController`, but `fetch` in Electron’s main process may not always respect `AbortSignal` depending on the Node version. Consider adding a `Promise.race` with a manual timeout rejection for safety.

### 5. `menu-i18n.ts` — `writeStoredMenuLocale` stores a JSON string, but `readStoredMenuLocale` parses it back

`writeStoredMenuLocale` does `JSON.stringify({ locale })`, then `readStoredMenuLocale` calls `parseMenuLocale` which does `JSON.parse`. This round-trip is harmless but slightly odd — the store already serializes values. Storing the raw string `\"zh\"` or `\"en\"` would be simpler and avoid the extra parse.

### 6. `build.yml` — `downloadExistingMetadataStep` condition is complex and duplicated

The `if` condition appears verbatim in three places (download step, collect step, finalize step). A workflow-level `env` or a reusable expression would reduce duplication and the risk of the conditions drifting out of sync.

### 7. `feedback.ts` — `errorMessage` is duplicated from `updater.ts`

Both files define an identical `errorMessage` helper. Consider extracting it to a shared utility to keep error formatting consistent.

### 8. `index.ts` — `syncMenuLocaleForWindow` calls `wireMenu()` on every focus change

If a user rapidly switches between two windows with different locales, `wireMenu()` rebuilds the entire application menu each time. In practice this is probably fine, but it’s worth noting that macOS menu rebuilding is not free and can cause subtle flicker or accessibility tree churn.

packages/desktop-electron/src/main/updater.ts
```diff
@@ -0,0 +1,97 @@
...
+      })
+      return inflight
+    },
+    install() {
+      if (!updateReady || readyVersion === undefined) return false
+      deps.quitAndInstall()
+      updateReady = false
+      readyVersion = undefined
+      return true
+    },
```
Astro-Han, line 83: `install()` returns `true`/`false` to indicate whether it actually triggered the install, but `index.ts:installUpdate()` just calls `updater.install()` and discards the result. Consider either using the return value or changing the signature to `void`.

packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  helpSubmenu.push({ label: t("openGithubIssue"), click: () => deps.openExternal(githubIssueUrl) })
+
+  return [
+    {
+      label: appName,
+      submenu: [
+        { role: "about" },
+        {
+          label: t("checkForUpdates"),
```
Astro-Han, line 53: The old menu had `enabled: UPDATER_ENABLED` on the "Check for Updates..." item. The new template always enables it. If the updater is disabled, clicking it will silently do nothing (the handler in `index.ts` still guards with `UPDATER_ENABLED`, but the menu item is no longer greyed out). Was this intentional?

packages/desktop-electron/src/main/problem-report.ts
```diff
@@ -0,0 +1,172 @@
...
+    output = markdown(makePayload())
+  }
+
+  if (bytes(output) > maxBytes && input.sessionExport.status === "ok" && sessionInfo !== null) {
+    omittedSessionInfoBytes += jsonBytes(sessionInfo)
+    sessionInfo = null
+    output = markdown(makePayload())
+  }
+
+  let diagnosticStringLimit = 1024
```
Astro-Han: `truncateString(value, 1024)` returns `value` unchanged when `value.length <= 1024`. The first iteration of the diagnostics-truncation loop therefore does nothing except increment `omittedDiagnosticsBytes` to 0. You could start `diagnosticStringLimit` at 512 to skip the wasted iteration.

packages/desktop-electron/src/main/index.ts
```diff
@@ -68,11 +74,114 @@ let mainWindow: BrowserWindow | null = null
...
+  downloadUpdate: () => autoUpdater.downloadUpdate(),
+  quitAndInstall: () => {
+    killSidecar()
+    autoUpdater.quitAndInstall()
+  },
+  log: (message, data) => logger.log(message, data),
+  error: (message, error) => logger.error(message, error),
+})
+
+function diagnostics(context = currentDesktopContext()) {
```
Astro-Han, line 105: The 10-second timeout uses `AbortController`, but `fetch` in Electron main process may not always respect `AbortSignal` depending on the Node version. Consider adding a `Promise.race` with a manual timeout rejection for safety.

packages/desktop-electron/src/main/menu-i18n.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function readStoredMenuLocale(systemLocale: string | null | undefined): MenuLocale {
+  const raw = getStore("opencode.global.dat").get("language")
+  const stored = parseMenuLocale(raw)
+  // Keep an explicit stored locale, including English; otherwise detect from the OS locale.
+  if (raw) return stored
+  return detectSystemMenuLocale(systemLocale)
+}
+
+export function writeStoredMenuLocale(locale: MenuLocale) {
```
Astro-Han, line 12: `writeStoredMenuLocale` does `JSON.stringify({ locale })`, then `readStoredMenuLocale` calls `parseMenuLocale` which does `JSON.parse`. This round-trip is harmless but slightly odd — the store already serializes values. Storing the raw string `"zh"` or `"en"` would be simpler and avoid the extra parse.

.github/workflows/build.yml
```diff
@@ -297,19 +299,62 @@ jobs:
...
           OPENCODE_CHANNEL: ${{ inputs.channel || 'dev' }}
+          PAWWORK_FEEDBACK_FORM_URL: ${{ vars.PAWWORK_FEEDBACK_FORM_URL || '' }}
 
       - name: Setup Apple API Key
         if: runner.os == 'macOS'
         run: printenv APPLE_API_KEY_CONTENT > $RUNNER_TEMP/apple-api-key.p8
         env:
           APPLE_API_KEY_CONTENT: ${{ secrets.APPLE_API_KEY }}
 
+      - name: Download existing updater metadata
```
Astro-Han, line 324: The `if` condition appears verbatim in three places (download step, collect step, finalize step). A workflow-level `env` or a reusable expression would reduce duplication and the risk of the conditions drifting out of sync.

packages/desktop-electron/src/main/feedback.ts
```diff
@@ -0,0 +1,69 @@
...
+  confirm: (context?: unknown) => Promise<boolean>
+  copy: (value: string) => Promise<void> | void
+  openExternal: (url: string) => Promise<void> | void
+  diagnostics: (context?: unknown) => ProblemReportDiagnostics
+  logTail: () => string
+  sessionExport: (context?: unknown) => Promise<SessionExport>
+  onError?: (error: unknown) => Promise<void> | void
+}
+
+function errorMessage(error: unknown) {
```
Astro-Han: Both files define an identical `errorMessage` helper. Consider extracting it to a shared utility to keep error formatting consistent.

packages/desktop-electron/src/main/index.ts
```diff
@@ -183,10 +292,12 @@ function mainWindowGlobals() {
 function openMainWindow() {
   const win = createMainWindow(mainWindowGlobals())
   mainWindow = win
+  win.on("focus", () => syncMenuLocaleForWindow(win))
```
Astro-Han, line 311: If a user rapidly switches between two windows with different locales, `wireMenu()` rebuilds the entire application menu each time. In practice this is probably fine, but its worth noting that macOS menu rebuilding is not free and can cause subtle flicker or accessibility tree churn.

Astro-Han:
@coderabbitai review

packages/desktop-electron/src/main/updater.ts
Astro-Han: Fixed. installUpdate() now uses the boolean return value from updater.install(), logs a skipped install when no ready update is latched, and shows the user the no-update dialog if the ready state disappeared before install started. The IPC type now returns boolean as well.

packages/desktop-electron/src/main/menu-template.ts
Astro-Han: Intentional. The handler no longer silently no-ops when updates are disabled: updater.check() returns status disabled and checkForUpdates(true) shows the localized disabled dialog. Keeping the item clickable lets users get visible feedback instead of a permanently greyed-out menu item with no explanation.

packages/desktop-electron/src/main/problem-report.ts
Astro-Han: Fixed. The diagnostics truncation loop now starts at 512 so it avoids the no-op 1024-byte pass.

packages/desktop-electron/src/main/index.ts
Astro-Han: Fixed. sessionExport() now races fetch() against an explicit timeout rejection while still aborting the request, so timeout behavior does not depend only on AbortSignal support.

packages/desktop-electron/src/main/menu-i18n.ts
Astro-Han: Leaving this as-is. The renderer bootstrap currently reads the same key as a serialized payload with a locale field, and earlier review caught that raw string storage could break restoration. parseMenuLocale supports both forms, but writing the JSON payload keeps the stored shape consistent with the renderer path.

.github/workflows/build.yml
Astro-Han: Not changing this in this PR. GitHub Actions does not give us a clean reusable expression for step if conditions, and hiding this behind env indirection would make the runner/phase/arch gate harder to audit. The duplicated condition is deliberate here because the three steps must stay visibly aligned.

packages/desktop-electron/src/main/feedback.ts
Astro-Han: Fixed. Extracted the shared errorMessage helper into src/main/error.ts and imported it from feedback and updater paths.

packages/desktop-electron/src/main/index.ts
Astro-Han: Acknowledged, no code change. wireMenu() is already gated by locale changes for the focused window, and when two focused windows genuinely have different locales the macOS application menu should follow focus. Debouncing would add lag and could leave the focused-window menu in the wrong language.

packages/app/src/pages/session.tsx
```diff
@@ -333,8 +334,54 @@ export default function Page() {
...
+        }
+      })
+      .catch(() => {
+        if (disposed) return
+        if (desktopContextRetryTimer !== undefined) window.clearTimeout(desktopContextRetryTimer)
+        desktopContextRetryTimer = window.setTimeout(() => {
+          desktopContextRetryTimer = undefined
+          if (disposed || pendingDesktopContext !== serialized || lastDesktopContext === serialized) return
+          syncDesktopContext(context, serialized)
+        }, 250)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The success and failure handlers now return immediately unless pendingDesktopContext still matches the serialized payload, so stale completions cannot clear or replace the retry timer for a newer payload.

packages/desktop-electron/src/main/index.ts
```diff
@@ -68,11 +74,114 @@ let mainWindow: BrowserWindow | null = null
...
+    return response.response === 0
+  },
+  copy: (value) => clipboard.writeText(value),
+  openExternal: (url) => {
+    return shell.openExternal(url).then(() => undefined)
+  },
+  diagnostics: (context) => diagnostics(context as DesktopContext),
+  logTail: tail,
+  sessionExport: (context) => sessionExport(context as DesktopContext),
+  onError: (error) => logger.error("problem report failed", error),
```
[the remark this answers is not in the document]
Astro-Han: Fixed. createFeedbackHandler onError now logs and shows a localized error dialog with the underlying error detail, so clipboard or external-link failures are visible to the user.

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -0,0 +1,111 @@
+export type MenuLocale = "en" | "zh"
+
```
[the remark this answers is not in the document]
Astro-Han: Not changing this. The current app locale set in this branch is only en and zh: packages/app/src/i18n contains en.ts and zh.ts, and normalizeLocale only returns en or zh. The referenced packages/app/src/i18n/en.ts is the English translation table, not an advertised list of additional supported locales.

Astro-Han:
Closing in favor of a fresh replacement PR from the same branch after resolving the accumulated review threads.

Issue #92, opened 2026-04-21
[Feature] Sync OpenCode upstream v1.14.20 intake


Astro-Han:
## What task are we trying to do?
Sync PawWork's vendored OpenCode foundation from the completed upstream `v1.4.11` cycle to the current upstream release tag `v1.14.20`. `v1.14.20` is now the release anchor for this intake cycle. Keep post-`v1.14.20` fixes as follow-up candidates unless a specific post-tag fix is proven to block macOS or Windows release work.

## What changed since this issue was opened?
Updated 2026-04-22: upstream published `v1.14.20` at 2026-04-21 19:07 UTC, tag commit `3175a3c618` (`release: v1.14.20`). The `v1.14.19..v1.14.20` delta contains 41 commits. Several commits previously listed as post-`v1.14.19` candidates are now release-tagged scope: `92c005866` fixes local dynamic import on Windows and Node by using `file://` URLs; `3406f1874` adds the plugin SDK `env` type; `38e2f4cdd` adds desktop webRequest CORS handling; `91468fe45` and `febadc558` are UI primitive fixes. The delta also includes upstream app UI, TUI, console, web, docs, Nix, debug-server, and experimental HTTP API work that should not automatically enter PawWork's core sync scope.

## What do we do today?
PawWork completed the previous upstream sync cycle at OpenCode `v1.4.11` through scoped squash PRs against `dev`. Current verification on 2026-04-22 shows upstream now publishes `v1.14.20`. PawWork product packages such as `packages/opencode`, `packages/app`, and `packages/desktop-electron` are at `0.2.5`, while upstream-tracked package surfaces such as `packages/plugin`, `packages/function`, `packages/shared`, `packages/ui`, and `packages/sdk/js` are now at `1.14.19` after PR #102. The full upstream tag range `v1.4.11..v1.14.20` reports 149 commits in GitHub compare, but prior PawWork sync PRs have already landed part of that range. Decide remaining work by comparing current code state and behavior, not by upstream commit hash alone.

## What would a good result look like?
Create a new upstream sync cycle that keeps PawWork's existing policy: use graft plus scoped squash merges, do not import upstream contributor history, and do not directly sync `packages/app/src/{components,pages,styles,theme,i18n}/**` because the PawWork app UI rewrite owns that layer. Treat the list below as the candidate intake scope, not literal `git cherry-pick` commands.

## Candidate intake commits
[...]

PR #102, opened 2026-04-21
refactor: sync opencode config schemas

Astro-Han:
## Summary

- Sync the OpenCode v1.14.19 config and schema base into PawWork while keeping `packages/opencode` on product version `0.2.4`.
- Regenerate OpenAPI and SDK surfaces, and bump upstream-tracked package versions from `1.4.11` to `1.14.19`.
- Preserve PawWork compatibility for `pawwork.json` aliases, config dependency bootstrap, console org state, config error serialization, and local tool dependency loading.

## Why

This is the Base PR for the OpenCode sync cycle. It brings over the schema and generated surfaces that later slices can build on without mixing provider, TUI, desktop, or runtime feature decisions into this PR.

## Compatibility Notes

- `@/config` now exposes config modules through namespace barrels such as `Config`, `ConfigAgent`, `ConfigPermission`, and `ConfigError`. Direct imports from the underlying files remain the safest internal path when a caller needs a specific schema or error class.
- `parseManagedPlist()` now returns sanitized JSON text for the caller to parse through the normal config parser. Malformed `plutil` JSON is propagated instead of becoming an empty config.
- Permission action schemas now live under `ConfigPermission.Action`; config errors now live under `ConfigError.JsonError` and `ConfigError.InvalidError`.
- MCP timeout is intentionally still a positive integer. The generated OpenAPI and SDK surfaces now reflect `integer`, `exclusiveMinimum: 0`.

## Related Issue

Part of #92.

## How To Verify

```bash
bun turbo typecheck
bun --cwd packages/opencode test test/config/config.test.ts test/tool/registry.test.ts test/util/effect-zod.test.ts test/server/middleware.test.ts
bun turbo test:ci
```

Latest local results for the review-fix batch:

- `bun turbo typecheck`: 8 successful tasks
- `bun --cwd packages/opencode test test/config/config.test.ts test/server/middleware.test.ts test/util/effect-zod.test.ts`: 170 pass, 0 fail
- `bun packages/sdk/js/script/build.ts`: completed and regenerated SDK v2 types
- `git diff --check`: clean

Earlier full local result for this PR:

- `bun turbo test:ci`: 2098 pass, 11 skip, 1 todo, 0 fail

## Screenshots or Recordings

No visible UI changes.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Expanded configuration: agents, commands, plugins, providers, permissions, MCP, keybinds, layout, server, skills, variables, managed preferences, and model-id validation.
* **Bug Fixes**
  * Improved dependency detection and install gating with skip-on-failure and clearer configuration error responses; safer event payload handling.
* **Tests**
  * Large test additions for config loading, parsing, schema interop, dependency installs, plugin/tool resolution, and managed prefs.
* **Chores**
  * Version bumped to 1.14.19 across packages.


Astro-Han:
@coderabbitai review

Astro-Han:
@coderabbitai review

Astro-Han (commented):
Superseded by English inline review comments. The previous non-inline review body was cleared because review findings should be left inline on the affected diff lines.

Astro-Han:
@coderabbitai review

packages/opencode/src/config/mcp.ts
```diff
@@ -0,0 +1,62 @@
...
+  command: Schema.mutable(Schema.Array(Schema.String)).annotate({
+    description: "Command and arguments to run the MCP server",
+  }),
+  environment: Schema.optional(Schema.Record(Schema.String, Schema.String)).annotate({
+    description: "Environment variables to set when running the MCP server",
+  }),
+  enabled: Schema.optional(Schema.Boolean).annotate({
+    description: "Enable or disable the MCP server on startup",
+  }),
+  timeout: Schema.optional(Schema.Number).annotate({
```
Astro-Han: [P2] MCP timeout accepts invalid values

This regresses the previous `number.int().positive()` validation to a bare `Schema.Number`, so local MCP timeout now accepts `0`, negative values, and fractional values. The remote timeout field has the same issue at line 50. These parsed values flow into the MCP connection timeout path, so invalid config values are no longer rejected during config loading. Please restore the positive integer constraint for both timeout fields and add an invalid timeout config test.

packages/opencode/src/config/server.ts
```diff
@@ -0,0 +1,20 @@
...
+  hostname: Schema.optional(Schema.String).annotate({ description: "Hostname to listen on" }),
+  mdns: Schema.optional(Schema.Boolean).annotate({ description: "Enable mDNS service discovery" }),
+  mdnsDomain: Schema.optional(Schema.String).annotate({
+    description: "Custom domain name for mDNS service (default: opencode.local)",
+  }),
+  cors: Schema.optional(Schema.mutable(Schema.Array(Schema.String))).annotate({
+    description: "Additional domains to allow for CORS",
+  }),
+}) {
+  static readonly zod = zod(this)
```
Astro-Han: [P2] Nested server config no longer rejects typos

`ServerConfig` now uses `zod(this)` directly, and the current `effect-zod` bridge emits a non-strict `z.object` for a plain struct. The top-level `Info` strictness only rejects unknown top-level keys; it does not reject unknown keys inside `server`, so a typo such as `hostnmae` is silently stripped. The previous schema on `dev` was `.strict()`. Please keep this derived schema strict, or make `ServerConfig` explicitly strict, and add a nested unknown-key test.

Astro-Han (commented):
Found two config validation regressions that should be fixed before merge.

Astro-Han:
@coderabbitai review

packages/opencode/src/tool/registry.ts
```diff
@@ -177,9 +177,15 @@ export namespace ToolRegistry {
...
             const spec = process.platform === "win32" ? match : pathToFileURL(match).href
-            if (!depsReady && (yield* Effect.promise(() => needsConfigDependencies(match, path.dirname(path.dirname(match)))))) {
+            const configDir = path.dirname(path.dirname(match))
+            const usesDeps = yield* Effect.promise(() => usesConfigDependencies(match))
+            if (!depsReady && usesDeps) {
               depsReady = true
+              const needsDeps = yield* Effect.promise(() => needsConfigDependencies(match, configDir))
               yield* config.waitForDependencies()
+              if (needsDeps) {
+                yield* Effect.tryPromise(() => Config.installDependencies(configDir)).pipe(Effect.orDie)
```
Astro-Han: [P2] Avoid turning dependency install failures into defects

This retry path now runs `Config.installDependencies(configDir)` inside `Effect.orDie`, so an install failure such as network loss, permission errors, or an invalid package file becomes a defect and can terminate tool registry initialization. The earlier flow only waited for the config dependency bootstrap and did not add this new fatal install path here. Please either handle this as a recoverable tool loading error, or document why dependency install failure should deliberately stop the whole registry.

packages/opencode/src/config/parse.ts
```diff
@@ -1,46 +1,44 @@
...
-        .join("\n")
-
-      throw new ConfigPaths.JsonError({
-        path: filepath,
-        message: `\n--- JSONC Input ---\n${text}\n--- Errors ---\n${issues}\n--- End ---`,
+import { type ParseError as JsoncParseError, parse as parseJsoncImpl, printParseErrorCode } from "jsonc-parser"
+import z from "zod"
+import { InvalidError, JsonError } from "./error"
+
+type Schema<T> = z.ZodType<T>
```
Astro-Han: [P3] Keep the parser schema interface duck-typed unless Zod-only is intentional

`ConfigParse.schema` used to accept any object with a compatible `safeParse` result shape, but this narrows the contract to `z.ZodType<T>`. All current callers pass Zod schemas, so this is not a runtime regression, but it makes the parser harder to reuse with an Effect Schema adapter or another compatible validator. If Zod-only is the intended boundary, please leave a short note here so future config work does not assume the old generic contract still exists.

packages/opencode/test/server/middleware.test.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, test } from "bun:test"
+import { Hono } from "hono"
+import { JsonError } from "../../src/config/error"
+import { ErrorMiddleware } from "../../src/server/middleware"
+
+describe("server error middleware", () => {
+  test("serializes config named errors instead of wrapping them as unknown errors", async () => {
```
Astro-Han, line 7: [P3] Cover the InvalidError serialization path too

This verifies that `JsonError` keeps its config error shape through `ErrorMiddleware`, but the common schema-validation path is `InvalidError`, which carries `issues` rather than `message`. Please add the symmetric case for `InvalidError` so a future change cannot accidentally drop `data.issues` while this test still passes.

Astro-Han (commented):
Migrated the remaining actionable top-level review notes into inline comments.

packages/opencode/src/tool/registry.ts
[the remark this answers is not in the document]
Astro-Han: Fixed in `f8cae08e9`: dependency install failures now return `false`, the registry logs the failure and skips dependency-backed local tools for that config dir, while still loading other tools. Added registry coverage for the failed-install path.

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,186 @@
...
+  "tools",
+])
+
+// Post-parse normalisation:
+//  - Promote any unknown-but-present keys into `options` so they survive the
+//    round-trip in a well-known field.
+//  - Translate the deprecated `tools: { name: boolean }` map into the new
+//    `permission` shape (write-adjacent tools collapse into `permission.edit`).
+//  - Coalesce `steps ?? maxSteps` so downstream can ignore the deprecated alias.
+const normalize = (agent: z.infer<typeof Info>) => {
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `e145f214c`: `normalize` now takes the pre-normalized `AgentInput` type instead of referencing `Info` before it is defined.

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,186 @@
...
+  const permission: ConfigPermission.Info = {}
+  for (const [tool, enabled] of Object.entries(agent.tools ?? {})) {
+    const action = enabled ? "allow" : "deny"
+    if (tool === "write" || tool === "edit" || tool === "patch" || tool === "multiedit") {
+      permission.edit = action
+      continue
+    }
+    permission[tool] = action
+  }
+  globalThis.Object.assign(permission, agent.permission)
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `e145f214c`: replaced `globalThis.Object.assign` with the standard `Object.assign`.

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,186 @@
...
+  return { ...agent, options, permission, steps: agent.steps ?? agent.maxSteps }
+}
+
+export const Info = zod(AgentSchema).transform(normalize).meta({ ref: "AgentConfig" }) as unknown as z.ZodType<
+  Omit<z.infer<ReturnType<typeof zod<typeof AgentSchema>>>, "options" | "permission" | "steps"> & {
+    options?: Record<string, unknown>
+    permission?: ConfigPermission.Info
+    steps?: number
+  }
+>
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `e145f214c`: normalized agent config now returns `AgentInfo`, omits deprecated `tools` and `maxSteps`, and the config test asserts those deprecated fields are not present after normalization.

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,186 @@
...
+      name,
+      ...md.data,
+      prompt: md.content.trim(),
+    }
+    const parsed = Info.safeParse(config)
+    if (parsed.success) {
+      result[config.name] = parsed.data
+      continue
+    }
+    throw new InvalidError({ path: item, issues: parsed.error.issues }, { cause: parsed.error })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `e145f214c`: invalid agent and mode schemas now publish a config load error and continue loading the remaining valid agents. Added coverage for one invalid agent alongside one valid agent.

packages/opencode/src/config/mcp.ts
```diff
@@ -0,0 +1,62 @@
...
+    description: "Timeout in ms for MCP server requests. Defaults to 5000 (5 seconds) if not specified.",
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `e145f214c`: local and remote MCP timeout fields now share a positive-integer schema. Added invalid timeout config coverage for zero, negative, and fractional values.

packages/opencode/src/config/server.ts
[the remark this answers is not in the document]
Astro-Han: Fixed in `e145f214c`: nested server config now uses the strict `Server.zod` through the schema override path, so typo keys such as `hostnmae` are rejected. Added the nested unknown-key regression test.

packages/opencode/src/config/mcp.ts
Astro-Han: Fixed in `e145f214c`: both local and remote MCP timeout fields now reject zero, negative, and fractional values. The new config test covers those invalid cases.

packages/opencode/src/config/server.ts
Astro-Han: Fixed in `e145f214c`: `server` now routes through the strict server zod schema, preserving nested typo rejection. Added a `hostnmae` regression test.

packages/opencode/src/tool/registry.ts
Astro-Han: Fixed in `f8cae08e9`: dependency install failures are recoverable. The registry records the failed config dir, logs the failure, skips tools that need those dependencies, and keeps loading tools that do not.

packages/opencode/src/config/parse.ts
Astro-Han: Fixed in `e145f214c`: `ConfigParse.schema` is duck-typed again around the compatible `safeParse` result shape instead of requiring `z.ZodType`.

packages/opencode/test/server/middleware.test.ts
Astro-Han: Fixed in `330c91dc7`: added the symmetric `InvalidError` middleware test and verified `data.issues` survives serialization.

Astro-Han:
@coderabbitai review

Astro-Han (commented):
Moved to inline review comment: https://github.com/Astro-Han/pawwork/pull/102#discussion_r3118626901

packages/opencode/src/config/config.ts
```diff
@@ -1,1699 +1,935 @@
...
-            //     so pawwork.{json,jsonc} are emitted after opencode.{json,jsonc}
-            //     and overwrite them in the merge.
-            for (const file of yield* Effect.promise(() =>
-              ConfigPaths.projectFiles(PROJECT_CONFIG_NAMES, ctx.directory, ctx.worktree),
-            )) {
-              yield* merge(file, yield* loadFile(file), "local")
-            }
+        const managedDir = ConfigManaged.managedConfigDir()
+        if (existsSync(managedDir)) {
+          for (const file of ["opencode.json", "opencode.jsonc"]) {
```
Astro-Han: [P2] Managed config skips PawWork aliases

PR #102 says it preserves PawWork config aliases, but this managed config path now only loads `opencode.json` and `opencode.jsonc`. On `dev`, the managed directory used the shared filename list, so `pawwork.json` and `pawwork.jsonc` were also honored. Add those filenames back here and cover it with a managed-settings alias regression test.

Astro-Han:
@coderabbitai review

packages/opencode/src/config/config.ts
Astro-Han: Fixed in `529cda450`: managed config now iterates the shared project config filenames, so `pawwork.json` and `pawwork.jsonc` are loaded again. Added a managed-settings alias regression test.

packages/opencode/src/config/command.ts
```diff
@@ -0,0 +1,62 @@
...
+      name,
+      ...md.data,
+      template: md.content.trim(),
+    }
+    const parsed = Info.zod.safeParse(config)
+    if (parsed.success) {
+      result[config.name] = parsed.data
+      continue
+    }
+    throw new InvalidError({ path: item, issues: parsed.error.issues }, { cause: parsed.error })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: command schema validation now reports the `InvalidError`, logs it, and continues loading the remaining command files. Added a regression test with one valid and one invalid command file.

packages/opencode/src/config/entry-name.ts
```diff
@@ -0,0 +1,16 @@
+import path from "path"
+
+function sliceAfterMatch(filePath: string, searchRoots: string[]) {
+  const normalizedPath = filePath.replaceAll("\\", "/")
+  for (const searchRoot of searchRoots) {
+    const index = normalizedPath.indexOf(searchRoot)
+    if (index === -1) continue
+    return normalizedPath.slice(index + searchRoot.length)
+  }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: config entry roots are normalized, trailing slashes are optional, empty roots are ignored, and the longest matching root wins. Added Windows-style and untrailed-root coverage.

packages/opencode/src/config/keybinds.ts
```diff
@@ -0,0 +1,127 @@
...
+import { Effect, Schema } from "effect"
+import type z from "zod"
+import { zod } from "@/util/effect-zod"
+
+// Every keybind field has the same shape: an optional string with a default
+// binding and a human description.  `keybind()` keeps the declaration list
+// below dense and readable.
+const keybind = (value: string, description: string) =>
+  Schema.String.pipe(Schema.optional, Schema.withDecodingDefault(Effect.succeed(value))).annotate({ description })
+
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `d22862025`: extracted a branded `KeybindString` scalar and reused it in the keybind helper while preserving the existing defaults and descriptions.

packages/opencode/src/config/managed.ts
```diff
@@ -0,0 +1,70 @@
+export * as ConfigManaged from "./managed"
+
+import { existsSync } from "fs"
+import os from "os"
+import path from "path"
+import { Log, Process } from "../util"
+import { warn } from "console"
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: removed the unused `warn` import from managed config.

packages/opencode/src/config/managed.ts
```diff
@@ -0,0 +1,70 @@
...
+  return process.env.OPENCODE_TEST_MANAGED_CONFIG_DIR || systemManagedConfigDir()
+}
+
+export function parseManagedPlist(json: string): string {
+  const raw = JSON.parse(json)
+  for (const key of Object.keys(raw)) {
+    if (PLIST_META.has(key)) delete raw[key]
+  }
+  return JSON.stringify(raw)
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: `parseManagedPlist` now catches malformed JSON, logs a warning, and returns an empty config object. Added malformed JSON coverage.

packages/opencode/src/config/markdown.ts
```diff
@@ -1,99 +1,99 @@
-import { NamedError } from "@opencode-ai/util/error"
+import { NamedError } from "@opencode-ai/shared/util/error"
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: config markdown, agent, and command now use the same `@opencode-ai/util/error` `NamedError` boundary as the rest of opencode, so `instanceof NamedError` checks stay coherent.

packages/opencode/src/config/mcp.ts
```diff
@@ -0,0 +1,64 @@
...
+}) {
+  static readonly zod = zod(this)
+}
+
+export const Info = Schema.Union([Local, Remote])
+  .annotate({ discriminator: "type" })
+  .pipe(withStatics((s) => ({ zod: zod(s) })))
+export type Info = Schema.Schema.Type<typeof Info>
+
+export * as ConfigMCP from "./mcp"
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `d22862025`: moved the `ConfigMCP` namespace re-export to the top of the module to match the surrounding config modules.

packages/opencode/src/config/model-id.ts
```diff
@@ -0,0 +1,13 @@
...
+import z from "zod"
+import { zod, ZodOverride } from "@/util/effect-zod"
+import { withStatics } from "@/util/schema"
+
+// Keep config model IDs as provider/model strings in generated SDK types.
+// External model schema references make the OpenAPI client generator resolve
+// these fields to the full Model object, which is not the config file shape.
+export const ConfigModelID = Schema.String.annotate({
+  [ZodOverride]: z.string(),
+}).pipe(withStatics((s) => ({ zod: zod(s) })))
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: `ConfigModelID` now rejects empty strings and values without non-empty provider and model parts. The regex intentionally allows additional slashes inside the model id because `Provider.parseModel` supports ids like `synthetic/hf:zai-org/GLM-4.7`. Added schema coverage for valid and invalid cases.

packages/opencode/src/config/plugin.ts
```diff
@@ -1,13 +1,88 @@
...
+  const file = (() => {
+    if (spec.startsWith("file://")) return spec
+    if (path.isAbsolute(spec) || /^[A-Za-z]:[\\/]/.test(spec)) return pathToFileURL(spec).href
+    return pathToFileURL(path.resolve(base, spec)).href
+  })()
+
+  const resolved = await resolvePathPluginTarget(file).catch(() => file)
+
+  if (Array.isArray(plugin)) return [resolved, plugin[1]]
+  return resolved
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: removed the blanket catch around `resolvePathPluginTarget`, so invalid plugin directories surface the specific validation error. Added a bad-directory regression test.

packages/opencode/src/config/provider.ts
```diff
@@ -0,0 +1,114 @@
...
+        }),
+        timeout: Schema.optional(
+          Schema.Union([PositiveInt, Schema.Literal(false)]).annotate({
+            description:
+              "Timeout in milliseconds for requests to this provider. Default is 300000 (5 minutes). Set to false to disable timeout.",
+          }),
+        ).annotate({
+          description:
+            "Timeout in milliseconds for requests to this provider. Default is 300000 (5 minutes). Set to false to disable timeout.",
+        }),
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `d22862025`: removed the redundant inner timeout annotation and kept the field description on the outer optional schema.

packages/opencode/src/config/skills.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const Info = Schema.Struct({
+  paths: Schema.optional(Schema.Array(Schema.String)).annotate({
+    description: "Additional paths to skill folders",
+  }),
+  urls: Schema.optional(Schema.Array(Schema.String)).annotate({
+    description: "URLs to fetch skills from (e.g., https://example.com/.well-known/skills/)",
+  }),
+}).pipe(withStatics((s) => ({ zod: zod(s) })))
+
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `d22862025`: converted the skills config schema to `Schema.Class` with a static `zod` property while preserving the existing field annotations.

packages/opencode/src/util/effect-zod.ts
```diff
@@ -1,20 +1,227 @@
...
+// Effectful / middleware-based transforms will surface as Effect defects.
+function decode(transformation: SchemaAST.Link["transformation"], value: unknown): unknown {
+  const exit = Effect.runSyncExit(
+    (transformation.decode as any).run(Option.some(value), EMPTY_PARSE_OPTIONS) as Effect.Effect<
+      Option.Option<unknown>
+    >,
+  )
+  if (exit._tag === "Failure") throw new Error(`effect-zod: transform failed: ${String(exit.cause)}`)
+  return Option.getOrElse(exit.value, () => value)
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `d22862025`: the transform decode error now explicitly calls out unsupported async or service-dependent transforms while preserving the raw Effect cause for diagnosis.

packages/opencode/test/config/config.test.ts
```diff
@@ -123,6 +126,27 @@ test("loads config with defaults when no files exist", async () => {
...
+      switchableOrgCount: 1.5,
+    }).success,
+  ).toBe(false)
+  expect(
+    ConsoleState.zod.safeParse({
+      consoleManagedProviders: [],
+      switchableOrgCount: -1,
+    }).success,
+  ).toBe(false)
+})
```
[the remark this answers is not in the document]
Astro-Han: Fixed in `529cda450`: added the zero boundary assertion for `switchableOrgCount`.

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,205 @@
...
+    if (!md) continue
+
+    const patterns = ["/.opencode/agent/", "/.opencode/agents/", "/agent/", "/agents/"]
+    const name = configEntryNameFromPath(item, patterns)
+
+    const config = {
+      name,
+      ...md.data,
+      prompt: md.content.trim(),
+    }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 561d9afd6 by spreading frontmatter first and writing the derived name last in both agent and mode loaders. Added regression coverage for agent and mode frontmatter name overrides.

packages/opencode/src/config/command.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!md) continue
+
+    const patterns = ["/.opencode/command/", "/.opencode/commands/", "/command/", "/commands/"]
+    const name = configEntryNameFromPath(item, patterns)
+
+    const config = {
+      name,
+      ...md.data,
+      template: md.content.trim(),
+    }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 561d9afd6 by spreading frontmatter first and writing the derived command name last. Added a regression test that keeps the filename-derived command key.

packages/opencode/src/config/keybinds.ts
```diff
@@ -0,0 +1,129 @@
...
+  history_previous: keybind("up", "Previous history item"),
+  history_next: keybind("down", "Next history item"),
+  session_child_first: keybind("<leader>down", "Go to first child session"),
+  session_child_cycle: keybind("right", "Go to next child session"),
+  session_child_cycle_reverse: keybind("left", "Go to previous child session"),
+  session_parent: keybind("up", "Go to parent session"),
+  // `terminal_suspend` was formerly `.default("ctrl+z").transform((v) => win32 ? "none" : v)`,
+  // but `tui.ts` already forces the binding to "none" on win32 before calling
+  // `Keybinds.parse(...)`, so the schema-level transform was redundant.
+  terminal_suspend: keybind("ctrl+z", "Suspend terminal"),
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 561d9afd6 by making the `terminal_suspend` schema default platform-aware, so direct `Keybinds.shape.terminal_suspend.parse(undefined)` also returns `none` on Windows.

packages/opencode/src/config/mcp.ts
```diff
@@ -0,0 +1,64 @@
...
+import { zod } from "@/util/effect-zod"
+import { withStatics } from "@/util/schema"
+
+const PositiveInt = Schema.Number.check(Schema.isInt()).check(Schema.isGreaterThan(0))
+
+export class Local extends Schema.Class<Local>("McpLocalConfig")({
+  type: Schema.Literal("local").annotate({ description: "Type of MCP server connection" }),
+  command: Schema.mutable(Schema.Array(Schema.String)).annotate({
+    description: "Command and arguments to run the MCP server",
+  }),
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 561d9afd6. The local MCP command schema now applies `Schema.isMinLength(1)` after `Schema.mutable(...)`, which the generated zod schema enforces. Added an invalid empty-command regression test.

packages/opencode/src/config/plugin.ts
```diff
@@ -1,13 +1,88 @@
...
+
+  for (const plugin of plugins.toReversed()) {
+    const spec = pluginSpecifier(plugin.spec)
+    const name = spec.startsWith("file://") ? spec : parsePluginSpecifier(spec).pkg
+    if (seen.has(name)) continue
+    seen.add(name)
+    list.push(plugin)
+  }
+
+  return list.toReversed()
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 561d9afd6. Local virtual config sources now resolve plugin specs before merge and dedupe, with OPENCODE_CONFIG_CONTENT regression coverage. The dedupe key also treats any remaining path-like spec as a path identity instead of parsing it as an npm package.

packages/opencode/test/config/config.test.ts
```diff
@@ -897,6 +1172,59 @@ it.live(
...
+
+      expect(pkg.dependencies?.["@opencode-ai/plugin"]).toBe(target)
+      expect(install).toHaveBeenCalledWith(tmp.extra)
+    } finally {
+      if (prev === undefined) delete process.env.OPENCODE_CONFIG_DIR
+      else process.env.OPENCODE_CONFIG_DIR = prev
+      install.mockRestore()
+    }
+  })
+})
```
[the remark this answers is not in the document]
Astro-Han: Not changing this one in this PR. I verified these tests follow the surrounding config dependency test style in this file: plain async tests with `tmpdir`, `withConfigDepsLock`, `Instance.provide`, and explicit env/mock cleanup. Migrating only these two cases to `it.live(...)` would be a harness refactor without changing the behavior under review, so I am keeping the local style here.

Astro-Han:
@coderabbitai review

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,205 @@
...
+    const action = enabled ? "allow" : "deny"
+    if (tool === "write" || tool === "edit" || tool === "patch" || tool === "multiedit") {
+      permission.edit = action
+      continue
+    }
+    permission[tool] = action
+  }
+  Object.assign(permission, configuredPermission)
+
+  return { ...rest, options, permission, steps: steps ?? maxSteps }
```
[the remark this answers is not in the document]
Astro-Han: Not changing this one. I checked the current schema and `permission: ".safe"` is rejected before `normalize()` runs: `ConfigPermission.Info` accepts permission objects or action strings (`ask`, `allow`, `deny`), and action strings are transformed to `{ "*": action }`. `ConfigAgent.Info.safeParse({ permission: ".safe" })` returns false, so this code path cannot corrupt a permission reference in the current contract.

packages/opencode/src/config/command.ts
```diff
@@ -0,0 +1,66 @@
...
+  log.error("failed to load command", { command: item, err: cause })
+}
+
+export const Info = Schema.Struct({
+  template: Schema.String,
+  description: Schema.optional(Schema.String),
+  agent: Schema.optional(Schema.String),
+  model: Schema.optional(ConfigModelID),
+  subtask: Schema.optional(Schema.Boolean),
+}).pipe(withStatics((s) => ({ zod: zod(s) })))
```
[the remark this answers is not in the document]
Astro-Han: Not changing this one in this PR. This is a style-only refactor, and `Schema.Struct` is still used in nearby config modules, including `formatter.ts`, `provider.ts`, `permission.ts`, `lsp.ts`, `config.ts`, and `keybinds.ts`. The current command schema also matches upstream `upstream/dev`, so converting just this schema to `Schema.Class` would add churn without fixing a behavior issue.

packages/opencode/src/config/command.ts
```diff
@@ -0,0 +1,66 @@
...
+    const name = configEntryNameFromPath(item, patterns)
+
+    const config = {
+      ...md.data,
+      name,
+      template: md.content.trim(),
+    }
+    const parsed = Info.zod.safeParse(config)
+    if (parsed.success) {
+      result[config.name] = parsed.data
```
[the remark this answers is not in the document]
Astro-Han: Fixed in ec9905dfc. The command loader now tracks the source path for each derived command name, reports duplicate names instead of overwriting, and keeps the first loaded command. Added regression coverage for `.opencode/command/dup.md` versus `.opencode/commands/dup.md`.

Astro-Han (commented):
## Review: nitpicks and edge cases

### `config/managed.ts:37-48`

`parseManagedPlist` 在 JSON 解析失败时静默返回 `"{}"`。调用方 `readManagedPreferences` 拿到 `{ text: "{}" }` 后会继续走正常的 config parse 流程，用户永远不会知道他们的 managed plist 其实坏了。建议让错误向上传播，或者至少让调用方有机会处理。

### `config/agent.ts:108-117`

`normalize` 中 `tools` 到 `permission` 的迁移有优先级问题：

```typescript
const permission: ConfigPermission.Info = {}
for (const [tool, enabled] of Object.entries(tools ?? {})) {
  permission[tool] = action  // 写入 "write", "edit" 等 key
}
Object.assign(permission, configuredPermission)  // configuredPermission 后写入，会覆盖
```

如果用户同时有 `tools: { edit: false }` 和 `permission: { edit: "allow" }`，`permission` 会赢。文档没说清楚哪个优先。建议明确文档化或调整顺序。

### `config/variable.ts:35-36`

`{env:VAR}` 替换在变量不存在时静默返回空字符串：

```typescript
return process.env[varName] || ""
```

环境变量不存在时变成空字符串，可能导致 config 值意外为空（如 API key 变成 `""`）。建议至少 warn 一下，或者支持 `{env:VAR?}` 可选语法。

### `config/entry-name.ts:3-16`

`sliceAfterMatch` 在 Windows 路径上的行为：如果路径中有大小写不一致（Windows 不区分大小写），`indexOf` 会失败。建议加 `toLowerCase()` 比较或注释说明。

### `config/command.ts:59` vs `config/agent.ts:158`

不一致的 parse API：
- `command.ts`: `Info.zod.safeParse(config)`
- `agent.ts`: `Info.safeParse(config)`

`Info` 在两个文件中类型不同（一个纯 Effect Schema，一个被 `zod(...).transform(...)` 包装），这种不一致会让读者困惑。建议统一风格。

### `util/effect-zod.ts:310`

`union` 函数在只有一个 string literal 成员时不会走 `z.enum`（要求 `>= 2`）。单个 literal 会输出 `{ const: "foo" }` 而非 `{ enum: ["foo"] }`，某些 OpenAPI 生成器对 `const` 支持不好。建议 `>= 1`。

### `config/keybinds.ts:125-127`

`as unknown as` 类型断言过于宽泛：

```typescript
export const Keybinds = zod(KeybindsSchema) as unknown as z.ZodObject<...>
```

如果 `KeybindsSchema` 未来加了非 string 字段，这个断言就撒谎了。建议用更窄的断言，或者不加。

### `config/tui.ts:122-124`

Windows 平台 `input_undo` 的 `??=` 逻辑：

```typescript
keybinds.input_undo ??= unique(["ctrl+z", ...Config.Keybinds.shape.input_undo.parse(undefined).split(",")]).join(",")
```

`Config.Keybinds.shape.input_undo.parse(undefined)` 在 Windows 上已经是 `"ctrl+z,ctrl+-,super+z"`，所以这里只是去重，没有实际新增。逻辑正确但冗余，建议注释说明为什么需要这一行。

### `config/provider.ts:5`, `config/mcp.ts:7`, `config/agent.ts:18`

`PositiveInt` 定义重复了 3 次：

```typescript
const PositiveInt = Schema.Number.check(Schema.isInt()).check(Schema.isGreaterThan(0))
```

建议抽到共享位置，比如 `util/schema.ts` 或 `config/shared.ts`。

### `test/server/middleware.test.ts`

没有测试 `InvalidError` 的 `issues: undefined` 情况（`issues` 是 optional 的）。建议补一个 case。

### `config/permission.ts:26-31`

`ZodPreprocess` 的 `__originalKeys` hack 注释说明了这是临时方案，但没有 issue 链接。建议创建一个跟踪 issue 来移除这个 hack。

### `config/lsp.ts:28-43`

`requiresExtensionsForCustomServers` 的 filter 返回 `string | undefined`：

```typescript
return ok ? undefined : "For custom LSP servers, 'extensions' array is required."
```

Effect Schema 的 `makeFilter` 期望 `boolean | undefined`，这里返回 `string` 作为错误消息。虽然能工作（truthy 就失败），但类型不匹配。建议确认 `makeFilter` 的契约。

### `config/mcp.ts:21-22`

`timeout` 从 `integer` 变成了 `number`（OpenAPI diff 可见）。这是有意放宽吗？如果是，应该在 PR 描述里说明。

### `config/server.ts:18`

`.strict()` 的 cast：

```typescript
static readonly zod = (zod(this) as unknown as z.ZodObject<any>).strict()
```

又一个 `as unknown as`。如果 `zod()` 函数能保留更多类型信息就好了。

packages/opencode/src/config/managed.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function managedConfigDir() {
+  return process.env.OPENCODE_TEST_MANAGED_CONFIG_DIR || systemManagedConfigDir()
+}
+
+export function parseManagedPlist(json: string): string {
+  let raw: Record<string, unknown>
+  try {
+    raw = JSON.parse(json)
+  } catch (error) {
```
Astro-Han, line 41: `parseManagedPlist` silently returns `"{}"` when JSON parsing fails. The caller `readManagedPreferences` receives `{ text: "{}" }` and continues with normal config parse flow — users will never know their managed plist is broken. Suggest propagating the error upward or giving the caller a chance to handle it.

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,205 @@
...
+  const permission: ConfigPermission.Info = {}
+  for (const [tool, enabled] of Object.entries(tools ?? {})) {
+    const action = enabled ? "allow" : "deny"
+    if (tool === "write" || tool === "edit" || tool === "patch" || tool === "multiedit") {
+      permission.edit = action
+      continue
+    }
+    permission[tool] = action
+  }
+  Object.assign(permission, configuredPermission)
```
Astro-Han, line 118: The `tools` to `permission` migration in `normalize` has a priority issue:

```typescript
const permission: ConfigPermission.Info = {}
for (const [tool, enabled] of Object.entries(tools ?? {})) {
  permission[tool] = action  // writes "write", "edit", etc.
}
Object.assign(permission, configuredPermission)  // configuredPermission overwrites
```

If a user has both `tools: { edit: false }` and `permission: { edit: "allow" }`, `permission` wins. The docs do not clarify which takes precedence. Suggest documenting this explicitly or reversing the order.

packages/opencode/src/config/variable.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function dir(input: ParseSource) {
+  return input.type === "path" ? path.dirname(input.path) : input.dir
+}
+
+/** Apply {env:VAR} and {file:path} substitutions to config text. */
+export async function substitute(input: SubstituteInput) {
+  const missing = input.missing ?? "error"
+  let text = input.text.replace(/\{env:([^}]+)\}/g, (_, varName) => {
+    return process.env[varName] || ""
```
Astro-Han: The `{env:VAR}` substitution silently returns empty string when the variable does not exist:

```typescript
return process.env[varName] || ""
```

Missing env vars become empty strings, which could cause config values to be unexpectedly empty (e.g., API key becomes `""`). Suggest at least logging a warning, or supporting an optional syntax like `{env:VAR?}`.

packages/opencode/src/config/entry-name.ts
```diff
@@ -0,0 +1,22 @@
...
+function sliceAfterMatch(filePath: string, searchRoots: string[]) {
+  const normalizedPath = filePath.replaceAll("\\", "/")
+  const normalizedRoots = searchRoots
+    .map((root) => root.replaceAll("\\", "/").replace(/\/+$/, ""))
+    .filter(Boolean)
+    .sort((a, b) => b.length - a.length)
+
+  for (const searchRoot of normalizedRoots) {
+    const needle = `${searchRoot}/`
+    const index = normalizedPath.indexOf(needle)
```
Astro-Han: `sliceAfterMatch` behavior on Windows paths: if there is case inconsistency (Windows is case-insensitive), `indexOf` will fail. Suggest using `toLowerCase()` for comparison or adding a comment explaining this.

packages/opencode/src/config/command.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const patterns = ["/.opencode/command/", "/.opencode/commands/", "/command/", "/commands/"]
+    const name = configEntryNameFromPath(item, patterns)
+
+    const config = {
+      ...md.data,
+      name,
+      template: md.content.trim(),
+    }
+    const parsed = Info.zod.safeParse(config)
```
Astro-Han, line 67: Inconsistent parse API:
- `command.ts`: `Info.zod.safeParse(config)`
- `agent.ts`: `Info.safeParse(config)`

`Info` has different types in these files (one is pure Effect Schema, one is wrapped by `zod(...).transform(...)`). This inconsistency confuses readers. Suggest unifying the style.

packages/opencode/src/util/effect-zod.ts
```diff
@@ -54,41 +265,99 @@ function body(ast: SchemaAST.AST): z.ZodTypeAny {
...
+    if (probe._tag !== "Success") continue
+    if (Option.isSome(probe.value)) return { value: probe.value.value }
+  }
+  return undefined
 }
 
 function union(ast: SchemaAST.Union): z.ZodTypeAny {
+  // When every member is a string literal, emit z.enum() so that
+  // JSON Schema produces { "enum": [...] } instead of { "anyOf": [{ "const": ... }] }.
+  if (ast.types.length >= 2 && ast.types.every((t) => t._tag === "Literal" && typeof t.literal === "string")) {
```
Astro-Han: The `union` function will not use `z.enum` for a single string literal member (requires `>= 2`). A single literal outputs `{ const: "foo" }` instead of `{ enum: ["foo"] }`. Some OpenAPI generators do not handle `const` well. Suggest changing to `>= 1`.

packages/opencode/src/config/keybinds.ts
```diff
@@ -0,0 +1,127 @@
...
+  plugin_manager: keybind("none", "Open plugin manager dialog"),
+  display_thinking: keybind("none", "Toggle thinking blocks visibility"),
+}).annotate({ identifier: "KeybindsConfig" })
+
+export type Keybinds = Schema.Schema.Type<typeof KeybindsSchema>
+
+// Consumers access `Keybinds.shape` and `Keybinds.shape.X.parse(undefined)`,
+// which requires the runtime type to be a ZodObject, not just ZodType.  Every
+// field is `string().optional().default(...)` at runtime, so widen to that.
+export const Keybinds = zod(KeybindsSchema) as unknown as z.ZodObject<
```
Astro-Han: The `as unknown as` type assertion is too broad:

```typescript
export const Keybinds = zod(KeybindsSchema) as unknown as z.ZodObject<...>
```

If `KeybindsSchema` adds a non-string field in the future, this assertion lies. Suggest using a narrower assertion or removing it.

packages/opencode/src/config/provider.ts
```diff
@@ -0,0 +1,109 @@
+import { Schema } from "effect"
+import { zod } from "@/util/effect-zod"
+import { withStatics } from "@/util/schema"
+
+const PositiveInt = Schema.Number.check(Schema.isInt()).check(Schema.isGreaterThan(0))
```
Astro-Han, line 5: `PositiveInt` is defined 3 times in different files:

```typescript
const PositiveInt = Schema.Number.check(Schema.isInt()).check(Schema.isGreaterThan(0))
```

Suggest extracting to a shared location like `util/schema.ts` or `config/shared.ts`.

packages/opencode/src/config/permission.ts
```diff
@@ -0,0 +1,76 @@
...
+  .annotate({ identifier: "PermissionRuleConfig" })
+  .pipe(withStatics((s) => ({ zod: zod(s) })))
+export type Rule = Schema.Schema.Type<typeof Rule>
+
+// Captures the user's original property insertion order before Schema.Struct
+// canonicalises the object.  See the `ZodPreprocess` comment in
+// `util/effect-zod.ts` for the full rationale — in short: rule precedence is
+// encoded in JSON key order (`evaluate.ts` uses `findLast`, so later keys win)
+// and `Schema.StructWithRest` would otherwise drop that order.
+const permissionPreprocess = (val: unknown) => {
```
Astro-Han, line 26: The `ZodPreprocess` `__originalKeys` hack comments explain it is a temporary workaround, but there is no issue link. Suggest creating a tracking issue to remove this hack.

packages/opencode/src/config/lsp.ts
```diff
@@ -1,64 +1,49 @@
...
+    Object.values(LSPServer)
+      .filter((server): server is LSPServer.Info => typeof server === "object" && server !== null && "id" in server)
+      .map((server) => server.id),
+  )
+  const ok = Object.entries(data).every(([id, config]) => {
+    if ("disabled" in config && config.disabled) return true
+    if (serverIds.has(id)) return true
+    return "extensions" in config && Boolean(config.extensions)
   })
+  return ok ? undefined : "For custom LSP servers, 'extensions' array is required."
```
Astro-Han, line 42: The `requiresExtensionsForCustomServers` filter returns `string | undefined`:

```typescript
return ok ? undefined : "For custom LSP servers, 'extensions' array is required."
```

Effect Schema's `makeFilter` expects `boolean | undefined`. Returning a `string` as error message works (truthy fails), but the types do not match. Suggest confirming `makeFilter`'s contract.

packages/opencode/src/config/mcp.ts
```diff
@@ -0,0 +1,64 @@
...
+  environment: Schema.optional(Schema.Record(Schema.String, Schema.String)).annotate({
+    description: "Environment variables to set when running the MCP server",
+  }),
+  enabled: Schema.optional(Schema.Boolean).annotate({
+    description: "Enable or disable the MCP server on startup",
+  }),
+  timeout: Schema.optional(PositiveInt).annotate({
+    description: "Timeout in ms for MCP server requests. Defaults to 5000 (5 seconds) if not specified.",
```
Astro-Han, line 21: `timeout` changed from `integer` to `number` (visible in OpenAPI diff). Is this intentional relaxation? If so, should be noted in PR description.

packages/opencode/src/config/server.ts
```diff
@@ -0,0 +1,21 @@
...
+  hostname: Schema.optional(Schema.String).annotate({ description: "Hostname to listen on" }),
+  mdns: Schema.optional(Schema.Boolean).annotate({ description: "Enable mDNS service discovery" }),
+  mdnsDomain: Schema.optional(Schema.String).annotate({
+    description: "Custom domain name for mDNS service (default: opencode.local)",
+  }),
+  cors: Schema.optional(Schema.mutable(Schema.Array(Schema.String))).annotate({
+    description: "Additional domains to allow for CORS",
+  }),
+}) {
+  static readonly zod = (zod(this) as unknown as z.ZodObject<any>).strict()
```
Astro-Han: The `.strict()` cast:

```typescript
static readonly zod = (zod(this) as unknown as z.ZodObject<any>).strict()
```

Another `as unknown as`. If `zod()` could preserve more type information, this would not be needed.

packages/opencode/test/server/middleware.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    expect(response.status).toBe(500)
+    expect(body.name).toBe("ConfigJsonError")
+    expect(body.data.path).toBe("opencode.json")
+    expect(body.data.message).toBe("bad json")
+  })
+
+  test("serializes config invalid errors with issues", async () => {
+    const app = new Hono().get("/boom", () => {
+      throw new InvalidError({
+        path: "opencode.json",
```
Astro-Han, line 25: Missing test for `InvalidError` with `issues: undefined` (the `issues` field is optional). Suggest adding a case to cover this.

Astro-Han (commented):
鸡蛋里挑骨头版 review。所有具体的观察已转为 inline 评论，请查看各文件上的具体建议。

**总结**：P1 建议修复的 10 条（错误处理、优先级、API 不一致等），P2 记录即可的 4 条（临时方案、类型契约、OpenAPI 变化等）。都是细节问题，不阻塞合并。

Astro-Han:
@coderabbitai review

Astro-Han (commented):
Nitpick review. All specific observations are inline comments on the affected files.

**Summary**: 10 P1 suggestions (error handling, priority issues, API inconsistencies, etc.), 4 P2 notes (temporary hacks, type contracts, OpenAPI changes, etc.). All are minor details that do not block merge.

Astro-Han (commented):
## Summary

This PR brings in the OpenCode v1.14.19 config schema sync while preserving PawWork compatibility. Overall well-structured for a base sync PR. A few observations:

### 1. Config namespace export style change

The index.ts export style changed from:
```typescript
// Before (dev)
import { Config } from "./config"
export { Config }
```

To:
```typescript
// After (PR)
export * as Config from "./config"
```

This changes `Config` from a direct value/namespace import to a barrel re-export. While TypeScript consumers using `import { Config } from "@/config/config"` directly should remain unaffected, callers using `import { Config } from "@/config"` now receive a namespace wrapper rather than the original Config namespace.

### 2. Self-referencing barrel exports pattern

Each config module file now starts with:
```typescript
export * as ConfigAgent from "./agent"
// ... rest of module implementation
```

This pattern allows internal code to reference `ConfigAgent.Info` before the actual export is defined. While TypeScript handles this correctly, the ordering is conceptually unusual — consider documenting why this is intentional if it's a deliberate pattern.

### 3. PawWork compatibility ✓

- `PROJECT_CONFIG_NAMES = ["opencode", "pawwork"]` 
- `GLOBAL_CONFIG_LOAD_FILES` and `GLOBAL_CONFIG_UPDATE_FILES` include `pawwork.json`/`pawwork.jsonc` 
- `globalConfigFile()` defaults to `pawwork.json` 
- `needsConfigDependencies` and `usesConfigDependencies` retained 
- `ConsoleState` Effect Schema migration with `.zod` compatibility 

### 4. Test coverage

New tests for pawwork.json config file loading and updates cover the PawWork-specific paths.

---

CI passes, PawWork compatibility preserved. Ready to merge.

packages/opencode/src/config/index.ts
```diff
@@ -1,11 +1,16 @@
...
-export { Config }
-
-export const ConfigManaged = {
-  parseManagedPlist(json: string) {
-    return JSON.stringify(Config.parseManagedPlist(json, "test:mobileconfig"))
-  },
-}
-
-export { ConfigMarkdown } from "./markdown"
+export * as Config from "./config"
```
Astro-Han, line 1: The export style changed from `export { Config }` (value import) to `export * as Config` (namespace barrel). While TypeScript consumers importing directly from `./config/config` remain unaffected, callers using `import { Config } from '@/config'` now receive a namespace wrapper. This is a structural shift worth documenting for downstream consumers.

packages/opencode/src/config/agent.ts
```diff
@@ -0,0 +1,205 @@
+export * as ConfigAgent from "./agent"
```
Astro-Han, line 1: Self-referencing barrel export: the file exports `ConfigAgent` namespace at the top before defining the actual `Info` schema at the bottom. This pattern works in TypeScript due to hoisting but is conceptually unusual — consider documenting the ordering rationale if intentional.

packages/opencode/src/config/managed.ts
```diff
@@ -0,0 +1,75 @@
...
+// Keys injected by macOS/MDM into the managed plist that are not OpenCode config
+const PLIST_META = new Set([
+  "PayloadDisplayName",
+  "PayloadIdentifier",
+  "PayloadType",
+  "PayloadUUID",
+  "PayloadVersion",
+  "_manualProfile",
+])
+
```
Astro-Han, line 21: The `parseManagedPlist` signature simplified: removed the `source` parameter and now returns `string` instead of `Info`. Callers must parse the returned JSON themselves. This is a behavior change from the previous `parseConfig(JSON.stringify(raw), source)` flow.

packages/opencode/src/config/config.ts
```diff
@@ -1,1699 +1,943 @@
-import { Log } from "../util/log"
+import { Log } from "../util"
 import path from "path"
 import { pathToFileURL } from "url"
 import os from "os"
```
Astro-Han, line 4: PawWork compatibility: `GLOBAL_CONFIG_UPDATE_FILES` puts `pawwork.jsonc` and `pawwork.json` first, ensuring PawWork config files take precedence when writing updates. Correct ordering for PawWork-first behavior.

packages/opencode/src/config/config.ts
```diff
@@ -1,1699 +1,943 @@
...
-import { Process } from "../util/process"
 import z from "zod"
-import { mergeDeep, unique } from "remeda"
+import { mergeDeep, pipe } from "remeda"
```
Astro-Han, line 6: Default to `pawwork.json` when no global config file exists — PawWork-first fallback. Good.

packages/opencode/src/config/console-state.ts
```diff
@@ -1,15 +1,16 @@
...
 
-export const ConsoleState = z.object({
-  consoleManagedProviders: z.array(z.string()),
-  activeOrgName: z.string().optional(),
-  switchableOrgCount: z.number().int().nonnegative(),
-})
-
-export type ConsoleState = z.infer<typeof ConsoleState>
+export class ConsoleState extends Schema.Class<ConsoleState>("ConsoleState")({
+  consoleManagedProviders: Schema.mutable(Schema.Array(Schema.String)),
```
Astro-Han, line 5: Effect Schema migration with `.zod` static for Zod compatibility. This pattern preserves OpenAPI $ref generation while enabling Effect Schema benefits. Good approach for gradual migration.

packages/opencode/src/config/dependency.ts
```diff
@@ -58,3 +64,33 @@ export async function needsConfigDependencies(file: string, configDir: string, v
 
   return false
```
Astro-Han, line 65: New function `usesConfigDependencies` complements `needsConfigDependencies`. The distinction: `needs` checks if dependencies are missing, `uses` checks if any package imports exist at all. Useful for waiting on in-progress installs. Good addition.

packages/opencode/src/config/permission.ts
```diff
@@ -0,0 +1,76 @@
+export * as ConfigPermission from "./permission"
+import { Schema } from "effect"
+import { zod, ZodPreprocess } from "@/util/effect-zod"
+import { withStatics } from "@/util/schema"
+
+export const Action = Schema.Literals(["ask", "allow", "deny"])
```
Astro-Han, line 6: Permission action schema moved to Effect Schema with `Action` type. Note: `Config.PermissionAction` was previously exported from `config.ts` but now lives here as `ConfigPermission.Action`. Internal callers using `Config.PermissionAction` will need to reference `ConfigPermission.Action` or use the `Config.Permission` namespace alias.

packages/opencode/src/config/error.ts
```diff
@@ -0,0 +1,21 @@
+export * as ConfigError from "./error"
+
+import z from "zod"
+import { NamedError } from "@opencode-ai/util/error"
+
+export const JsonError = NamedError.create(
+  "ConfigJsonError",
```
Astro-Han, line 7: Error classes moved from `ConfigPaths.JsonError`/`ConfigPaths.InvalidError` to standalone `JsonError`/`InvalidError` exports in `error.ts`. Callers using `ConfigPaths.JsonError` should import from `./error` or use `ConfigError.JsonError` namespace.

packages/opencode/src/config/parse.ts
```diff
@@ -1,46 +1,48 @@
...
+        return `${error}\n   Line ${line}: ${problemLine}\n${"".padStart(column + 9)}^`
       })
-    }
-
-    return data
+      .join("\n")
+    throw new JsonError({
+      path: filepath,
+      message: `\n--- JSONC Input ---\n${text}\n--- Errors ---\n${issues}\n--- End ---`,
+    })
```
Astro-Han, line 34: Throws `JsonError` instead of `ConfigPaths.JsonError`. The error class now lives in `./error.ts` and is imported directly here. Same behavior, cleaner module separation.

packages/opencode/src/config/managed.ts
Astro-Han: Fixed in c4cc852cf. `parseManagedPlist()` now logs and throws instead of returning `{}` on malformed JSON, so callers cannot silently continue with an empty managed config. Added a regression test for the malformed JSON path.

packages/opencode/src/config/agent.ts
Astro-Han: Kept the current precedence, but made it explicit in c4cc852cf. `permission` wins over deprecated `tools` on conflicts because `permission` is the replacement surface. Added a code comment and a regression test for `tools.edit = false` plus `permission.edit = allow`.

packages/opencode/src/config/variable.ts
Astro-Han: Fixed in c4cc852cf. `{env:VAR}` now raises `ConfigInvalidError` when the env var is missing. `{env:VAR?}` is supported for the intentional optional-empty case. Added tests for both paths.

packages/opencode/src/config/entry-name.ts
Astro-Han: Fixed in c4cc852cf. `sliceAfterMatch()` now performs case-insensitive matching for normalized roots while preserving the original returned path casing. Added a Windows-style mixed-case regression test.

packages/opencode/src/config/command.ts
Astro-Han: Leaving this as-is for this Base PR. The difference is intentional: command `Info` is still an Effect Schema with an attached `.zod`, while agent `Info` is the Zod transform because normalization changes the output shape. Unifying that style would be a broader schema API cleanup, not a behavior fix for this PR.

packages/opencode/src/util/effect-zod.ts
Astro-Han: Fixed in c4cc852cf. Single string-literal unions now use `z.enum()` as well, and `test/util/effect-zod.test.ts` covers the single-literal JSON Schema output.

packages/opencode/src/config/keybinds.ts
Astro-Han: Fixed in c4cc852cf. Replaced the `as unknown as` with a runtime `ZodObject` guard plus a narrower typed cast, keeping the `.shape` API available to existing callers.

packages/opencode/src/config/provider.ts
Astro-Han: Not extracting this in this PR. The three local `PositiveInt` definitions are small and keep each config module self-contained. A shared schema module would be reasonable cleanup later, but it is extra churn for this Base PR.

packages/opencode/src/config/permission.ts
Astro-Han: Fixed in c4cc852cf. Created tracking issue #113 and added the reference beside the `__originalKeys` workaround comment.

packages/opencode/src/config/lsp.ts
Astro-Han: Leaving this as-is. The bridge already supports string-returning filters as custom failure messages, and `test/util/effect-zod.test.ts` covers that contract. Returning only boolean would drop the useful validation message here.

packages/opencode/src/config/mcp.ts
Astro-Han: Fixed in c4cc852cf. This was not an intentional relaxation. Regenerated OpenAPI and SDK surfaces now show MCP timeout as `integer` with `exclusiveMinimum: 0`; the PR body also notes this explicitly.

packages/opencode/src/config/server.ts
Astro-Han: Fixed in c4cc852cf. Replaced the `as unknown as` path with a runtime `ZodObject` guard and preserved schema metadata so OpenAPI still emits the `ServerConfig` ref.

packages/opencode/test/server/middleware.test.ts
Astro-Han: Fixed in c4cc852cf. Added the symmetric middleware test for `InvalidError` without `issues`, covering the optional-field serialization path.

packages/opencode/test/config/config.test.ts
```diff
@@ -558,6 +760,106 @@ test("handles command configuration", async () => {
...
+    directory: tmp.path,
+    fn: async () => {
+      const config = await load()
+      expect(config.command?.["dup"]).toEqual({
+        description: "Singular command",
+        template: "Run the singular command",
+      })
+    },
+  })
+})
```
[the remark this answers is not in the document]
Astro-Han: Fixed in c4cc852cf. The command loader now sorts discovered command files with an explicit rank, preferring `command/` over `commands/` before duplicate detection. The test no longer depends on glob brace expansion order.

packages/opencode/src/config/index.ts
Astro-Han: Documented in the PR body under Compatibility Notes. Direct module imports remain unaffected, while `@/config` consumers should treat these as namespace barrels.

packages/opencode/src/config/agent.ts
Astro-Han: Leaving the per-file ordering uncommented. This namespace self-export pattern is consistent across the config modules in this PR, and adding the same ordering note to each module would be noise. The PR body now documents the namespace barrel compatibility boundary.

packages/opencode/src/config/managed.ts
Astro-Han: Documented and tightened in c4cc852cf. The PR body now notes that `parseManagedPlist()` returns sanitized JSON text for normal parser flow, and malformed JSON now propagates instead of becoming an empty config.

packages/opencode/src/config/config.ts
Astro-Han: Acknowledged. No change needed; this ordering is intentional for PawWork-first config updates.

packages/opencode/src/config/config.ts
Astro-Han: Acknowledged. No change needed; defaulting to `pawwork.json` is intentional PawWork compatibility behavior.

packages/opencode/src/config/console-state.ts
Astro-Han: Acknowledged. No change needed; this is the intended gradual migration pattern for Effect Schema plus Zod compatibility.

packages/opencode/src/config/dependency.ts
Astro-Han: Acknowledged. No change needed; `usesConfigDependencies` and `needsConfigDependencies` intentionally model different dependency states.

packages/opencode/src/config/permission.ts
Astro-Han: Documented in the PR body under Compatibility Notes. Callers should use `ConfigPermission.Action` or the namespace alias rather than the old `Config.PermissionAction` surface.

packages/opencode/src/config/error.ts
Astro-Han: Documented in the PR body under Compatibility Notes. Callers should import from `./error` directly or use `ConfigError.JsonError` and `ConfigError.InvalidError`.

packages/opencode/src/config/parse.ts
Astro-Han: Acknowledged. No change needed; this is the intended module split and keeps the behavior the same while moving the error class out of `ConfigPaths`.

Astro-Han:
@coderabbitai review

packages/opencode/src/config/variable.ts
```diff
@@ -0,0 +1,98 @@
...
+    const optional = rawName.endsWith("?")
+    const varName = optional ? rawName.slice(0, -1) : rawName
+    const value = process.env[varName]
+    if (value !== undefined) return value
+    if (optional || missing === "empty") return ""
+    throw new InvalidError({
+      path: source(input),
+      message: `missing environment variable: "${token}"`,
+    })
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 3602dabc9. Substitution now uses the JSONC scanner to detect line and block comment tokens before expanding `{env:...}` or `{file:...}`. Added a regression test covering full-line comments, inline comments, and block comments.

PR #103, opened 2026-04-21
chore: remove unused TUI surface

### LINKED ISSUE #91: [Feature] Remove unused TUI and Tauri compatibility code
Keep PawWork GUI-only by removing unused TUI and Tauri compatibility code that no longer belongs to the product surface.

## What do you do today?
The repository still carries terminal UI and Tauri-era code even though PawWork now ships as an Electron GUI application. This increases the amount of code that needs to be checked during upstream sync and makes future cherry-picks harder to review because unrelated TUI/Tauri changes can appear in the diff.

PawWork keeps only the GUI and embedded server paths it actually uses. Tauri compatibility leftovers are removed where they are no longer needed. TUI code is removed through a deliberate cleanup that first disconnects build entries, server routes, generated SDK surfaces, plugin exports, config files, and tests so the Electron GUI path remains working.

Both

## Extra context
Initial read-only check found Tauri leftovers in the Electron migration/titlebar/CORS compatibility areas and TUI references in the opencode build script, server /tui routes, config, plugin types, tests, and OpenTUI dependencies. Suggested split: first remove low-risk Tauri compatibility leftovers, then remove TUI in a second scoped change after confirming the GUI does not call /tui APIs and replacing any TUI-only event usage such as MCP toast publishing.

Astro-Han:
## Summary

Remove the unused TUI surface now that PawWork is GUI-only:

- deleted the opencode TUI runtime tree, TUI config files, server TUI routes, schema output, parser worker wiring, and OpenTUI build entries
- narrowed plugin install/load/meta handling to server plugins only, with TUI-only and `oc-themes` packages treated as unsupported for install
- regenerated OpenAPI and SDK outputs without `/tui` routes or TUI client helpers

## Why

The Tauri compatibility cleanup from #93 is already merged. This completes the second cleanup slice from #91 by removing the unused TUI code path while keeping Electron GUI, embedded server, and PTY paths intact.

## Related Issue

Closes #91

## How To Verify

```bash
bun test --timeout 30000
bun test test/config/config.test.ts test/plugin/loader-shared.test.ts test/plugin/install.test.ts test/plugin/install-concurrency.test.ts test/server/no-tui-routes.test.ts test/script/schema.test.ts test/mcp/no-tui-event.test.ts
bun run typecheck
bun run build
bun run smoke:ci
bun run script/build.ts --single --skip-install --skip-embed-web-ui
```

Also ran package-level typechecks/builds for `packages/opencode`, `packages/plugin`, `packages/sdk/js`, `packages/app`, and `packages/desktop-electron`, plus a final residual scan for TUI/OpenTUI symbols.

## Screenshots or Recordings

Not applicable. This removes unused runtime/API/build surfaces and does not add visible UI behavior.

## Checklist


Astro-Han (commented):
Non-inline review body cleared. There are no remaining actionable inline findings for this PR.

packages/opencode/test/script/schema.test.ts
```diff
@@ -0,0 +1,25 @@
...
+import fs from "fs/promises"
+import path from "path"
+import { tmpdir } from "../fixture/fixture"
+
+test("schema generator writes only the main config schema", async () => {
+  await using tmp = await tmpdir()
+  const configFile = path.join(tmp.path, "config.json")
+  const tuiFile = path.join(tmp.path, "tui.json")
+
+  const proc = Bun.spawn(["bun", "script/schema.ts", configFile, tuiFile], {
```
Astro-Han: [P3] Make the ignored second argument explicit

The generator no longer consumes a second output path, but this test still passes `tuiFile` as the second CLI argument and then asserts that file is not written. That is a useful guard, but the variable name still reads like a supported TUI schema output. Please rename it to something like `extraIgnoredFile` so the test documents that the argument is intentionally ignored after TUI schema generation was removed.

Astro-Han (commented):
Migrated the remaining actionable top-level review note into an inline comment.

packages/opencode/test/script/schema.test.ts
Astro-Han: Fixed in the latest push. Renamed `tuiFile` to `extraIgnoredFile` and changed the dummy path to `ignored.json` so the test reads as an ignored extra argument guard.

Astro-Han (commented):
[P2] Minor test naming issues found during strict review.

**1. packages/opencode/test/plugin/install.test.ts:151**

Test name `preserves JSONC comments when adding plugins to server and tui config` still implies modification of tui config. The test verifies that **existing tui.jsonc is left untouched** when installing a package that exposes both server and tui exports.

Consider renaming to: `preserves JSONC comments when adding plugins to server config, leaves existing tui config untouched` or simply `preserves JSONC comments when adding plugins to server config (tui exports ignored)`.

**2. packages/opencode/test/plugin/install.test.ts:60**

The `plugin()` fixture helper creates mock package.json exports including `./tui`. This is valid for testing the "package exposes tui exports but we must not process them" scenario. However, the helper lacks a comment documenting this intent.

Adding a brief note like:
```ts
// Fixture helper for testing packages that expose server and/or tui exports.
// Used to verify: (1) server exports are processed, (2) tui exports are ignored.
```
Would make the test's purpose clearer for future readers.

---

All CI passing. Core cleanup is complete and thorough. These are minor documentation polish items.

packages/opencode/test/plugin/install.test.ts
```diff
@@ -64,6 +64,7 @@ async function plugin(
   },
   themes?: string[],
 ) {
+  // Fixtures may expose tui exports to verify that plugin install ignores them.
```
Astro-Han: 注释提到 "tui exports" 但 TUI 已被移除，建议改为更准确的描述：

```typescript
// Fixtures may expose legacy tui exports to verify that plugin install correctly rejects unsupported targets.
```

这样更明确表达了 fixture 的目的是测试"拒绝不支持的 target"这一行为，而不是验证"忽略"。

packages/opencode/bunfig.toml
```diff
@@ -1,7 +1,5 @@
-preload = ["@opentui/solid/preload"]
-
```
Astro-Han, line 2: 变更后只剩一个 preload 条目。请确认 bun test 仍能正常运行 —— preload 用于测试环境的 mock runtime，移除 @opentui/solid/preload 后需要验证测试运行时没有引入意外依赖。

Astro-Han (commented):
## Code Review Report

### Overall Assessment

The PR structure is clear, and the commit split is reviewable. Each commit has one reversible intent. The change correctly removes the TUI runtime and related dependencies.

### P0 (Blocking)

None.

### P1 (Should Fix)

See the inline comments below.

### P2 (Notes)

1. The bun.lock diff is large because it removes the OpenTUI dependency tree. This is expected.
2. The new test names are clear, including no-tui-routes.test.ts and no-tui-event.test.ts.
3. The OpenAPI diff is large because the TUI API surface was removed. This is expected.

### Verification Points (All Passed)

| Check | Result |
|-------|--------|
| OpenTUI dependencies removed | Passed |
| CLI entry points have no TUI residue | Passed |
| SDK API removed correctly | Passed |
| OpenAPI has no /tui endpoints | Passed |
| Plugin exports narrowed correctly | Passed |
| Tests cover the behavior change | Passed |
| Config.Keybinds removed correctly | Passed |
| keybind.ts removed correctly | Passed |

### Recommendation

After fixing the P1 comment issue, this should be ready to merge. The scope is controlled, the tests cover the behavior, and I did not find remaining TUI references that should block the PR.

packages/opencode/test/plugin/install.test.ts
Astro-Han: Fixed in the latest push. Updated the fixture helper comment to call these legacy tui exports and clarify that they verify unsupported targets are rejected.

packages/opencode/bunfig.toml
Astro-Han: Verified after rebasing onto the latest dev. `bun test --timeout 30000` in packages/opencode completed successfully with 1957 pass, 8 skip, 1 todo, 0 fail, so the remaining bunfig preload is sufficient and removing the OpenTUI preload did not break test runtime startup.

Astro-Han (commented):
Overall this is a well-executed cleanup PR. The changes systematically remove TUI surface across packages/opencode, packages/plugin, and packages/sdk. The addition of regression tests (`no-tui-routes.test.ts`, `no-tui-event.test.ts`) is good practice.

A few minor observations for consideration:

---

**1. `patchName` retains a now-meaningless parameter**

`packages/opencode/src/plugin/install.ts:329-330`

Since `Kind` type now only contains `"server"`, the `_kind` parameter serves no purpose. The underscore prefix correctly indicates it's unused, but conceptually the function could simplify to:
```ts
function patchName(): "opencode" {
  return "opencode"
}
```
Or even just inline `"opencode"` at the call site. Not a bug, just a small conceptual redundancy left over from the refactor.

---

**2. `readV1Plugin` tui check order change**

`packages/opencode/src/plugin/shared.ts:283-300` (patch)

The original code checked `typeof tui !== "function"` before rejecting. After the patch, invalid tui exports (non-function) get the error "unsupported tui export" instead of "invalid tui export". This is arguably more appropriate since TUI is no longer a supported target at all, so "unsupported" is semantically correct. Just noting the change in error message semantics.

---

**3. MCP auth no longer notifies via toast**

`packages/opencode/src/mcp/index.ts:335-346` (patch)

The MCP OAuth failure handling now returns `Effect.succeed(undefined)` instead of publishing `TuiEvent.ToastShow`. This is expected given TUI removal. Confirming that GUI has alternative notification mechanisms for these auth warnings (or that they're handled differently in the GUI client).

---

**4. Test fixture comment explains legacy behavior**

`packages/opencode/test/plugin/install.test.ts:67`

Good addition of the comment explaining why fixtures still generate tui exports:
```ts
// Fixtures may expose legacy tui exports to verify that plugin install rejects unsupported targets.
```

---

**5. Schema generator ignores extra argument**

`packages/opencode/script/schema.ts:51-59` (patch)

The script now accepts but ignores the second `tuiFile` argument. The test in `schema.test.ts` verifies this behavior by passing an extra file that should not be written. Clean approach.

---

No blocking issues. The PR accomplishes its stated goal of removing unused TUI surface while preserving server plugin functionality.

packages/opencode/src/plugin/install.ts
```diff
@@ -13,10 +13,10 @@ import { Filesystem } from "@/util/filesystem"
 import { Flock } from "@/util/flock"
 import { isRecord } from "@/util/record"
```
Astro-Han, line 15: Since Kind now only has "server", _kind parameter is unused. Could simplify to `patchName(): "opencode"` or inline. Minor conceptual redundancy from refactor.

Astro-Han (commented):
Overall: well-executed cleanup. A few minor observations (P2, non-blocking).

packages/opencode/src/plugin/shared.ts
```diff
@@ -107,7 +107,6 @@ function resolvePackageEntrypoint(spec: string, kind: PluginKind, pkg: PluginPac
     if (raw) return resolvePackagePath(spec, raw, kind, pkg)
   }
 
-  if (kind !== "server") return
   const main = packageMain(pkg)
   if (!main) return
   return resolvePackagePath(spec, main, kind, pkg)
```
Astro-Han, line 112: Error message semantics changed: invalid tui exports now get "unsupported" instead of "invalid". Semantically appropriate since TUI is no longer supported. Noting change for completeness.

packages/opencode/src/mcp/index.ts
```diff
@@ -333,25 +332,11 @@ export namespace MCP {
                     status: "needs_client_registration" as const,
                     error: "Server does not support dynamic client registration. Please provide clientId in config.",
                   }
-                  return bus
-                    .publish(TuiEvent.ToastShow, {
-                      title: "MCP Authentication Required",
```
Astro-Han, line 338: MCP OAuth failure now returns `Effect.succeed(undefined)` instead of publishing toast. Expected given TUI removal. Confirm GUI has alternative notification mechanisms for auth warnings.

Astro-Han: Confirmed. The GUI does not depend on the old TUI toast path here: MCP auth states are exposed through `MCP.status()` and the `/mcp` server route, then consumed by the app status UI. `packages/app/src/context/global-sync/bootstrap.ts` loads `sdk.mcp.status()`, `packages/app/src/components/status-popover.tsx` treats non-connected/non-disabled MCP states as issues, and `packages/app/src/components/session/session-status-connections.tsx` surfaces `needs_auth` and `needs_client_registration` as warning states. So no code change is needed for this PR.

packages/opencode/src/plugin/install.ts
Astro-Han: Fixed in the latest push. I removed the `patchName` helper and inlined the fixed `opencode` config name in `patchOne`, since plugin install now only patches the server config.

packages/opencode/src/plugin/shared.ts
Astro-Han: No code change here. I agree the new `unsupported target` wording is the intended behavior now that TUI exports are legacy and no longer accepted. Keeping this as a semantic note.

Astro-Han (commented):
Strict nitpicking review for PR #103.

packages/opencode/test/plugin/install.test.ts
```diff
@@ -206,7 +205,7 @@ describe("plugin.install.task", () => {
     const serverJson = parseJsonc(serverText) as { plugin?: unknown[] }
```
Astro-Han, line 206: The test "preserves JSONC comments when adding server plugin and ignoring tui config" reads an existing tui.jsonc file and expects it to contain only the seed plugin. This is correct behavior, but the test name says "ignoring tui config" which could be misleading. The test actually verifies that the pre-existing tui config is left untouched (not ignored). Consider renaming to "preserves JSONC comments when adding server plugin, leaves pre-existing tui config unchanged" for clarity.

packages/opencode/test/plugin/install.test.ts
```diff
@@ -499,7 +497,7 @@ describe("plugin.install.task", () => {
     const serverJson = await read(server)
```
Astro-Han, line 498: The test "force replaces version only in server config" reads tuiJson and expects it to remain unchanged (acme@1.0.0 stays). This is correct behavior, but the test name says "only in server config" which implies tui config is not read at all. The test actually reads tui config to verify it was not modified. Consider renaming to "force replaces version in server config, leaves tui config unchanged" for accuracy.

Astro-Han (commented):
Additional nitpicking review comments.

packages/opencode/test/plugin/install-concurrency.test.ts
```diff
@@ -86,7 +86,7 @@ describe("plugin.install.concurrent", () => {
     expectPlugins(cfg.plugin, all)
   }, 25_000)
 
-  test("serializes concurrent server+tui config updates across processes", async () => {
+  test("serializes concurrent updates for packages that expose server and tui", async () => {
```
Astro-Han: The test name "serializes concurrent updates for packages that expose server and tui" still mentions "tui" prominently. While the test verifies that tui.jsonc is not created (correct), the name could be clearer. The test actually verifies serialization of config updates when multiple processes install packages that expose both targets. Consider renaming to "serializes concurrent updates for packages exposing server exports (tui exports ignored)" to emphasize the actual behavior being tested.

Astro-Han (commented):
More nitpicking review comments.

packages/opencode/src/plugin/shared.ts
```diff
@@ -287,17 +248,11 @@ export function readV1Plugin(
   if (server !== undefined && typeof server !== "function") {
     throw new TypeError(`Plugin ${spec} has invalid server export`)
   }
-  if (tui !== undefined && typeof tui !== "function") {
-    throw new TypeError(`Plugin ${spec} has invalid tui export`)
-  }
-  if (server !== undefined && tui !== undefined) {
-    throw new TypeError(`Plugin ${spec} must default export either server() or tui(), not both`)
-  }
   if (kind === "server" && server === undefined) {
```
Astro-Han, line 251: In readV1Plugin, after removing the explicit tui type check, a non-function tui export now throws "Plugin  has unsupported tui export". This is semantically correct, but the error message could be more helpful. Consider adding context like "Plugin  has unsupported tui export - TUI plugins are no longer supported in PawWork" so users understand why their plugin is rejected.

Astro-Han (commented):
Code logic nitpicking.

packages/opencode/src/cli/cmd/plug.ts
```diff
@@ -115,9 +115,7 @@ export function createPlugTask(input: PlugInput, dep: PlugDeps = defaultPlugDeps
       if (manifest.code === "manifest_no_targets") {
         inspect.stop("No plugin targets found", 1)
         dep.log.error(`"${mod}" does not expose plugin entrypoints in package.json`)
```
Astro-Han, line 117: The error message update from 'Expected one of: exports["./tui"], exports["./server"], package.json main for server, or package.json["oc-themes"] for tui themes.' to 'Expected one of: exports["./server"] or package.json main.' is correct. However, the message could be slightly more informative for debugging. Consider: 'Expected exports["./server"] entrypoint or package.json main field with a server() export.' to hint at the required export structure.

Astro-Han (commented):
Final nitpicking comment.

packages/opencode/test/plugin/install.test.ts
```diff
@@ -64,6 +64,7 @@ async function plugin(
   },
   themes?: string[],
 ) {
+  // Fixtures may expose legacy tui exports to verify that plugin install rejects unsupported targets.
```
Astro-Han: The fixture helper comment "Fixtures may expose legacy tui exports to verify that plugin install rejects unsupported targets" is good, but it could mention that the helper also creates both server and tui export entries in package.json, not just tui. The fixture is used in tests that install packages exposing both server and tui. Consider expanding to: "Fixture helper creates mock packages exposing server and/or legacy tui exports. Used to verify: (1) server exports are processed correctly, (2) tui exports are rejected as unsupported targets." for completeness.

Astro-Han (commented):
Test fixture documentation nitpick.

packages/opencode/test/plugin/install.test.ts
Astro-Han: Fixed in the latest push. Renamed the test to `force replaces version in server config, leaves tui config unchanged` to make the unchanged tui config check explicit.

packages/opencode/test/plugin/install.test.ts
Astro-Han: Fixed in the latest push. Renamed the test to `preserves JSONC comments when adding server plugin, leaves pre-existing tui config unchanged` so it describes the untouched legacy config assertion.

packages/opencode/test/plugin/install-concurrency.test.ts
Astro-Han: Fixed in the latest push. Renamed the concurrency test to `serializes concurrent updates for packages exposing server exports with tui exports ignored`, keeping the focus on server config serialization and unsupported tui exports.

packages/opencode/src/plugin/shared.ts
Astro-Han: Fixed in the latest push. The unsupported tui export error now adds context: `TUI plugins are no longer supported in PawWork`.

packages/opencode/src/cli/cmd/plug.ts
Astro-Han: Fixed in the latest push. Updated the no-targets hint to `Expected exports["./server"] entrypoint or package.json main field with a server() export.`

packages/opencode/test/plugin/install.test.ts
Astro-Han: Fixed in the latest push. Expanded the fixture helper comment to state that it creates mock packages exposing server and/or legacy tui exports, and that the fixture verifies server processing plus unsupported tui export rejection.

packages/opencode/src/plugin/shared.ts
Astro-Han, line 251: In readV1Plugin, the order of checks has changed. Previously, a non-function tui export would throw "has invalid tui export" before the "unsupported" check. Now, any tui export (function or not) throws "unsupported tui export". This is semantically correct for the TUI removal, but the error message "TUI plugins are no longer supported in PawWork" could be slightly misleading - PawWork is the GUI app, the plugin system runs in the opencode runtime. Consider "TUI plugins are no longer supported (PawWork is now GUI-only)" or just "TUI plugins are no longer supported in this runtime".

Astro-Han (commented):
Strict nitpicking review round 2.

Astro-Han (commented):
More nitpicking.

packages/opencode/test/plugin/loader-shared.test.ts
```diff
@@ -1161,7 +977,7 @@ export default {
         scope: "local" as const,
```
Astro-Han, line 977: The test "retries failed file plugins once after wait and keeps order" now uses kind: "server" instead of "tui". This is correct, but the test description in the test name could be updated to reflect the actual behavior being tested - specifically that the retry logic applies to server plugins now. The test itself is correct, but naming it "retries failed file server plugins once after wait and keeps order" would be clearer.

packages/opencode/src/config/config.ts
```diff
@@ -83,7 +83,6 @@ function normalizeLoadedConfig(data: unknown, source: string) {
   delete copy.theme
```
Astro-Han, line 83: The deprecation warning "tui keys in opencode config are deprecated; move them to tui.json" was removed along with the TUI config support. However, users who have old opencode.json files with theme/keybinds/tui keys will now have those keys silently ignored without any warning. Consider whether a brief migration notice or changelog entry is needed to inform users that these keys are now simply removed without migration to tui.json (since tui.json itself is no longer supported).

Astro-Han (commented):
More nitpicking comments.

packages/sdk/js/src/v2/server.ts
```diff
@@ -98,37 +89,3 @@ export async function createOpencodeServer(options?: ServerOptions) {
     },
```
Astro-Han, line 89: The createOpencodeTui function was removed from the SDK. This is correct, but the removal could be documented in the SDK's changelog or release notes. Users who previously used createOpencodeTui from @opencode-ai/sdk/v2 will get a runtime error "createOpencodeTui is not defined" rather than a helpful deprecation message. Consider whether a stub function with a helpful error message should be kept temporarily to guide users, or if this is acceptable for a breaking change PR.

Astro-Han (commented):
SDK surface nitpick.

packages/opencode/script/build.ts
```diff
@@ -166,7 +164,6 @@ await $`rm -rf dist`
 
 const binaries: Record<string, string> = {}
 if (!skipInstall) {
```
Astro-Han, line 166: The parser.worker.js resolution logic was removed along with the TUI worker path. The OTUI_TREE_SITTER_WORKER_PATH and OPENCODE_WORKER_PATH defines were also removed. This is correct for TUI removal. However, the build script still references pkg.dependencies for @parcel/watcher. The dependency list in package.json was updated, but I notice the build script references pkg.dependencies to get version numbers. Double-check that all remaining dependency references in this script still exist in the updated package.json after removing @opentui/core.

Astro-Han (commented):
Build script nitpick.

packages/opencode/src/plugin/meta.ts
```diff
@@ -130,7 +122,6 @@ export namespace PluginMeta {
       time_changed: prev?.time_changed ?? now,
```
Astro-Han, line 122: The PluginMeta.Entry type no longer has a themes field, and the setTheme function was removed. This is correct for TUI removal. However, the existing plugin-meta.json files (if any users have them) may still contain the themes field. The code now silently ignores this field when reading existing metadata. Consider whether this is acceptable or if a migration/cleanup step should be considered to remove obsolete theme data from existing metadata files. The fingerprint computation was also updated - plugins that previously had themes will now have a different fingerprint and be marked as "updated" on next load.

Astro-Han (commented):
Plugin meta nitpick.

packages/opencode/test/plugin/install.test.ts
```diff
@@ -472,12 +470,13 @@ describe("plugin.install.task", () => {
     )
 
     const ok = await run(ctx(tmp.path))
```
Astro-Han, line 472: The test "ignores oc-themes when server target exists" verifies that oc-themes is not processed when a server target exists. However, the test name could be clearer about the actual behavior: oc-themes are no longer processed at all, regardless of whether a server target exists. The test happens to use a package with both server and oc-themes, but it could also test the case where oc-themes alone exist (covered elsewhere) or where oc-themes exist without server (should return false). The current test name is accurate for the test case, but consider renaming to "installs server plugin when package has both server target and oc-themes (oc-themes ignored)" for precision.

Astro-Han (commented):
Test coverage nitpick.

packages/opencode/src/flag/flag.ts
```diff
@@ -19,7 +19,6 @@ export namespace Flag {
   export const OPENCODE_GIT_BASH_PATH = process.env["OPENCODE_GIT_BASH_PATH"]
   export const OPENCODE_CONFIG = process.env["OPENCODE_CONFIG"]
   export declare const OPENCODE_PURE: boolean
-  export declare const OPENCODE_TUI_CONFIG: string | undefined
   export declare const OPENCODE_CONFIG_DIR: string | undefined
   export declare const OPENCODE_PLUGIN_META_FILE: string | undefined
```
Astro-Han, line 23: OPENCODE_TUI_CONFIG was removed from Flag namespace. This is correct. However, I notice OPENCODE_CONFIG_CONTENT is still present (used for passing config via env var). OPENCODE_CONFIG_CONTENT was previously used by createOpencodeTui in the SDK as well. The SDK removal removed that usage, but OPENCODE_CONFIG_CONTENT is still declared here. Double-check that OPENCODE_CONFIG_CONTENT is still needed by other code paths (likely for embedded server or GUI use cases), otherwise it could be considered a potential cleanup candidate in a future PR.

Astro-Han (commented):
Flag removal nitpick.

packages/opencode/src/server/instance/index.ts
```diff
@@ -39,7 +38,6 @@ export const InstanceRoutes = (upgrade: UpgradeWebSocket): Hono =>
     .route("/", FileRoutes())
```
Astro-Han, line 38: The TuiRoutes import and route mounting were removed. This is correct. However, I notice the route order: FileRoutes, EventRoutes, McpRoutes are mounted in sequence. Previously TuiRoutes was mounted after McpRoutes. The route order doesn't matter much since routes are path-specific, but for consistency and documentation, consider whether the route mounting order should be documented or if any route collision edge cases existed with /tui prefix that are now resolved.

Astro-Han (commented):
Server routes nitpick.

packages/opencode/tsconfig.json
```diff
@@ -3,14 +3,12 @@
   "extends": "@tsconfig/bun/tsconfig.json",
   "compilerOptions": {
     "jsx": "preserve",
-    "jsxImportSource": "@opentui/solid",
     "lib": ["ESNext", "DOM", "DOM.Iterable"],
     "types": [],
     "noUncheckedIndexedAccess": false,
```
Astro-Han, line 8: The @tui/* path alias was removed from tsconfig paths. This is correct. However, I notice the @/* path still exists. Double-check that no remaining source files still have import statements using @tui/* paths - the test file cleanup removed the tui tests, but there might be residual imports in other files that weren't deleted. TypeScript would catch this at compile time, but a quick verification would confirm completeness.

Astro-Han (commented):
tsconfig nitpick.

packages/plugin/package.json
```diff
@@ -10,8 +10,7 @@
   },
```
Astro-Han, line 10: The ./tui export was removed from packages/plugin exports. This is correct. However, the peerDependencies for @opentui/core and @opentui/solid were removed with peerDependenciesMeta marking them as optional. This is correct, but I notice the devDependencies for these packages were also removed. Double-check that packages/plugin no longer needs these dependencies for any remaining functionality (e.g., type declarations or build-time checks). The remaining dependencies (effect, zod) are sufficient for the server plugin target.

Astro-Han (commented):
Package exports nitpick.

package.json
```diff
@@ -27,8 +27,6 @@
       "@types/cross-spawn": "6.0.6",
```
Astro-Han, line 27: The workspace catalog entries for @opentui/core and @opentui/solid were removed. This is correct. However, I notice other catalog entries like @hono/zod-validator, @octokit/rest, etc. are still present. Double-check that the catalog is now correctly pruned to only include dependencies that are actually used across the workspace. The bun.lock diff shows the opentui packages being removed from the lockfile, which confirms the catalog removal is consistent with the dependency changes.

Astro-Han (commented):
Root package.json nitpick.

packages/opencode/src/plugin/shared.ts
Astro-Han: Fixed in the latest push. Updated the error context to `TUI plugins are no longer supported in this runtime`, so the message stays tied to the opencode runtime rather than the GUI app name.

packages/sdk/js/src/v2/server.ts
Astro-Han: No code change here. Keeping a `createOpencodeTui` stub would preserve a removed SDK API surface and conflict with the generated SDK/OpenAPI cleanup in this PR. I would document this as part of the breaking-change release notes rather than keep a temporary runtime shim.

packages/opencode/script/build.ts
Astro-Han: Checked. The build script has one remaining `pkg.dependencies[...]` lookup, for `@parcel/watcher`, and `packages/opencode/package.json` still declares `@parcel/watcher` in dependencies. No missing dependency reference remains after removing OpenTUI.

packages/opencode/src/config/config.ts
Astro-Han: No code change here. `normalizeLoadedConfig` already strips legacy `theme`, `keybinds`, and `tui` keys rather than migrating them to `tui.json`. Adding a runtime warning would keep a TUI-specific migration path alive after removal. This is better covered in release notes or a changelog entry for the breaking TUI removal.

packages/opencode/test/plugin/loader-shared.test.ts
Astro-Han: Fixed in the latest push. Renamed the test to `retries failed file server plugins once after wait and keeps order`.

packages/opencode/src/plugin/meta.ts
Astro-Han: No code change here. Existing metadata with a legacy `themes` field is acceptable: `touchMany` rewrites `store[item.id]` with the new `Entry` shape on the next load, so obsolete theme data is naturally dropped for touched plugins. The one-time `updated` fingerprint state is expected because theme metadata no longer participates in the plugin identity.

packages/opencode/src/flag/flag.ts
Astro-Han: Checked. `OPENCODE_CONFIG_CONTENT` is still used by the opencode config loader, plugin source handling, both SDK server helpers, and config tests. It is still needed for embedded server and GUI-style config injection, so no cleanup in this PR.

packages/opencode/tsconfig.json
Astro-Han: Checked. `rg "@tui/" packages package.json` returns no matches, and `packages/opencode` typecheck passes. The remaining `@/*` alias is still used by normal opencode source imports.

packages/opencode/src/server/instance/index.ts
Astro-Han: No code change here. The remaining routes are mounted on distinct prefixes except the existing root file/event mounts, and removing `/tui` removes that entire path prefix. I found no route collision that needs documentation or ordering changes in this PR.

packages/opencode/test/plugin/install.test.ts
Astro-Han: Fixed in the latest push. Renamed the test to `installs server plugin when package has server target and oc-themes, ignores oc-themes`, while the oc-themes-only unsupported case stays covered by the neighboring test.

packages/plugin/package.json
Astro-Han: Checked. `packages/plugin/src` has no `@opentui` imports; remaining imports are from `@opencode-ai/sdk`, `@opencode-ai/sdk/v2`, `effect`, and `zod`. The package typecheck passes with the remaining dependencies, so the OpenTUI peer/dev dependencies are no longer needed.

package.json
Astro-Han: Checked. This PR removes only the OpenTUI catalog entries, and those are the only catalog entries removed versus `origin/dev`. A catalog audit still reports pre-existing unused entries: `@types/semver`, `dompurify`, `@playwright/test`, and `@solidjs/start`. Since those are unrelated to TUI removal and pre-existing, I am leaving broader catalog pruning out of this PR.

PR #110, opened 2026-04-21
feat: add dynamic right-panel shell tabs

Astro-Han:
## Summary

- Add dynamic shell tab state for the session right panel, with Status pinned and Review, Files, Terminal, and Context opened on demand.
- Migrate legacy right-panel state into the new shell-tab model while preserving existing active review and file-tab behavior.
- Render dynamic shell tabs with close, reorder, and add-tab controls, and route callers through the shell-tab API.

## Why

The right panel previously showed fixed shell tabs even when a user only needed one surface. Dynamic shell tabs make the panel match the active workflow while keeping old persisted layouts compatible.

## Related Issue

None.

## How To Verify

```bash
cd packages/app
bun run typecheck
bun run test:unit
```

Local screenshot smoke was also run with a temporary E2E spec. The captured screenshot is available in the Codex session at `/tmp/pawwork-dynamic-shell-tabs-e2e.png`.

## Screenshots or Recordings

Local E2E screenshot reviewed in Codex session: `/tmp/pawwork-dynamic-shell-tabs-e2e.png`.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Draggable/reorderable right‑panel tabs (status stays pinned), plus an "Add tab" menu and new "context" tab; middle‑click/menu tab close support; improved desktop terminal toggle.

* **Bug Fixes**
  * More consistent side‑panel/tab behavior when opening files, comments, or toggling terminals; improved tab selection fallbacks.

* **Refactor**
  * Simplified tab/panel management and persisted-state migration; added i18n key for "Add tab" and a small record-type utility.

* **Tests**
  * Expanded unit and e2e coverage for tab migration, behavior, and side‑panel flows.


packages/app/src/pages/session/right-panel-tabs.ts
```diff
@@ -1,10 +1,143 @@
...
+
+export const toggleShellTab = (
+  state: ShellTabState,
+  target: RightPanelTab,
+  panelOpen: boolean,
+): { state: ShellTabState; closePanel: boolean } => {
+  if (target === "status") return { state: openShellTab(state, target), closePanel: false }
+  if (state.sidePanelTab === target && panelOpen) return { state: closeShellTab(state, target), closePanel: false }
+  return { state: openShellTab(state, target), closePanel: false }
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed in dc00344ac. `toggleShellTab` now returns `closePanel: true` when the target tab is already active and the panel is open, preserving the tab in `openShellTabs`. Added a regression test for this behavior.

packages/app/src/pages/session/session-side-panel.tsx
```diff
@@ -103,7 +117,7 @@ export function SessionSidePanel(props: {
   }
 
   const openReviewPanel = () => {
-    if (!view().reviewPanel.opened()) view().reviewPanel.open()
+    if (!view().sidePanel.opened()) view().sidePanel.open()
   }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in dc00344ac. Opening a review file now calls `sidePanel.openTab("review")`, so the panel opens and Review becomes active. Added a focused regression test for the helper.

packages/app/src/context/layout.tsx
```diff
@@ -188,6 +219,13 @@ export const { use: useLayout, provider: LayoutProvider } = createSimpleContext(
         }
       })()
 
+      const rightPanel = value.rightPanel
+      const migratedRightPanel = (() => {
+        if (!isRecord(rightPanel)) return rightPanel
+        if (typeof rightPanel.opened === "boolean") return rightPanel
+        return { ...rightPanel, opened: legacyRightPanelOpened(rightPanel, review, fileTree) }
+      })()
```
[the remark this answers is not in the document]
Astro-Han: Fixed in f41cb5e52. `migrateStoredLayout` now creates a default `rightPanel` object when persisted `rightPanel` is missing and carries over `review.panelOpened` or `fileTree.opened`. Added a regression test for the missing-rightPanel legacy case.

packages/app/src/pages/session/migrate-session-view.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (entry.contextInSessionTabs && !candidate.includes("context")) {
+      candidate.push("context")
+    }
+
+    const activeCandidate: RightPanelTab | undefined =
+      entry.sessionTabsRaw?.active === "context"
+        ? "context"
+        : entry.sessionTabsRaw?.active === "changes" || entry.sessionTabsRaw?.active === "review"
+          ? "review"
+          : (entry.migratedSidePanelTab ?? (entry.reviewInSessionTabs ? "review" : undefined))
```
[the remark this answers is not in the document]
Astro-Han: Addressed in f41cb5e52. Extracted the active shell tab resolution into `resolveActiveCandidate` with the same precedence: context, review or changes, then migrated side panel tab or review presence.

packages/app/src/pages/session/right-panel-tabs.ts
```diff
@@ -1,10 +1,143 @@
...
+
+export const moveShellTab = (state: ShellTabState, target: RightPanelTab, to: number): ShellTabState => {
+  if (target === "status") return state
+
+  const from = state.openShellTabs.indexOf(target)
+  if (from === -1) return state
+
+  const next = [...state.openShellTabs]
+  next.splice(to, 0, next.splice(from, 1)[0])
+  return normalizeShellTabs({ openShellTabs: next, sidePanelTab: state.sidePanelTab })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in f41cb5e52. `moveShellTab` now clamps the destination index to keep moves after the pinned Status tab. Added a regression test for negative indexes.

packages/app/src/pages/session/session-side-panel.tsx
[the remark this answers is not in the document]
Astro-Han: Already addressed in dc00344ac. Current head routes this through `openReviewShellTab(view().sidePanel)`, which calls `sidePanel.openTab("review")` so opening a file activates the Review shell. The regression test is in `session-side-panel.test.tsx`.

packages/app/src/pages/session/terminal-shell-tab.ts
```diff
@@ -0,0 +1,21 @@
...
+  open: () => void
+  close: () => void
+}
+
+export function toggleDesktopTerminal(view: DesktopTerminalView, terminal: DesktopTerminal) {
+  const open = view.sidePanel.opened() && view.sidePanel.tab() === "terminal"
+  view.sidePanel.toggleTab("terminal")
+  if (open) terminal.close()
+  else terminal.open()
+}
```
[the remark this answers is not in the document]
Astro-Han: Addressed in f41cb5e52. Kept the immediate terminal open and close calls because they provide command feedback before the side-panel synchronization effect runs, and added a short comment noting the effect remains the idempotent global sync path.

Astro-Han (commented):
## Nitpick Review

All CI checks pass. I addressed the low-risk consistency items that were worth folding into this PR:

- `migrate-session-view.ts` now reuses `same()` instead of keeping a local `arraysEqual` helper.
- `right-panel-tabs.ts` now documents why `coerceLegacySidePanelTab` can return `undefined` while `migrateLegacyRightPanelTab` returns a concrete fallback.

Remaining observations are intentionally left as follow-up-level cleanup, not blockers:

- `isPlainObject` could eventually be consolidated with other local `isRecord` helpers, but introducing a shared utility would broaden this PR.
- `RightPanelShellIconName` is a small convenience alias. Keeping it is clearer than inlining the union everywhere.
- `resolveActiveCandidate` remains private and is covered through `migrateSessionView` behavior tests. Exporting it only for direct tests would expand the module API without much value.

The implementation is ready to merge.

Astro-Han:
@coderabbitai review

packages/app/src/pages/session/migrate-session-view.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export interface MigrationOutput {
+  sessionView: Record<string, unknown>
+  sessionTabs: Record<string, unknown>
+  changed: boolean
+}
+
+const isPlainObject = (value: unknown): value is Record<string, unknown> =>
+  typeof value === "object" && value !== null && !Array.isArray(value)
```
Astro-Han: The `isPlainObject` helper is defined locally here, but there's an identical `isRecord` helper in `layout.tsx` (added in this PR). For future maintainability, consider consolidating these into a shared utility (e.g., `@/utils/is-record`). Not blocking for this PR since the duplication is contained within the migration module, but worth a follow-up cleanup.

packages/app/src/pages/session/migrate-session-view.ts
```diff
@@ -0,0 +1,134 @@
...
+
+const fileTabsOnly = (tabs: readonly unknown[]): string[] =>
+  tabs.filter((tab): tab is string => typeof tab === "string" && tab.startsWith("file://"))
+
+const resolveActiveCandidate = (entry: LegacyEntry): RightPanelTab | undefined => {
+  const rawActive = entry.sessionTabsRaw?.active
```
Astro-Han, line 24: The `resolveActiveCandidate` function handles precedence for determining the active tab during migration. While it's covered indirectly via `migrateSessionView` tests, exporting it (or adding direct unit tests) would make the precedence rules more explicit and easier to verify independently. The current test coverage relies on integration-style tests that don't directly assert the resolution logic.

packages/app/src/pages/session/session-side-panel.tsx
```diff
@@ -50,22 +58,33 @@ export function shouldShowReviewFileOpenButton(activeTab: string | undefined, ha
...
+      </Match>
+      <Match when={props.icon.kind === "icon" && props.icon.name === "status"}>
         <Icon name="status" size="small" class="text-text-weaker" />
       </Match>
-      <Match when={props.icon === "folder"}>
+      <Match when={props.icon.kind === "icon" && props.icon.name === "folder"}>
         <Icon name="folder" size="small" class="text-text-weaker" />
       </Match>
-      <Match when={props.icon === "review"}>
+      <Match when={props.icon.kind === "icon" && props.icon.name === "review"}>
```
Astro-Han, line 84: The `openReviewShellTab` helper is a single-line wrapper that calls `sidePanel.openTab("review")`. If this helper exists purely for readability at call sites, consider adding a brief docstring explaining its purpose. Otherwise, inlining the call might be clearer since the function name already describes what it does.

packages/app/src/pages/session/terminal-shell-tab.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+
+type DesktopTerminal = {
+  open: () => void
+  close: () => void
+}
+
+export function toggleDesktopTerminal(view: DesktopTerminalView, terminal: DesktopTerminal) {
+  const open = view.sidePanel.opened() && view.sidePanel.tab() === "terminal"
```
Astro-Han, line 17: The comment explains the immediate-call intention, but the wording could be clearer: "Apply immediately for command feedback" doesn't specify what feedback. Consider: "Terminal open/close is called immediately for responsive UI feedback; the createEffect in session-side-panel.tsx ensures state stays synchronized idempotently." This makes the dual-update pattern explicit.

packages/app/src/pages/session/helpers.ts
```diff
@@ -59,14 +57,15 @@ export const createSessionTabs = (input: TabsInput) => {
...
   const closableTab = createMemo(() => {
-    const active = activeTab()
-    if (active === "context") return active
-    if (!openedTabs().includes(active)) return
-    return active
+    const active = input.tabs().active()
+    const normalizedActive = active ? input.normalizeTab(active) : undefined
+    if (normalizedActive && input.pathFromTab(normalizedActive)) return normalizedActive
+    const current = activeTab()
+    if (openedTabs().includes(current)) return current
```
Astro-Han, line 67: The `closableTab` memo now has more complex logic: it first checks the raw active tab for valid file paths, then falls back to `activeTab()`. Consider adding a brief comment explaining the priority order: "File tabs take precedence for closing; only fall back to shell tabs when no file tab is active." This clarifies the behavior change from the previous simpler implementation.

packages/app/src/pages/session/right-panel-tabs.ts
```diff
@@ -1,10 +1,146 @@
...
+
+export const isRightPanelTab = (value: unknown): value is RightPanelTab =>
+  typeof value === "string" && (RIGHT_PANEL_TAB_VALUES as readonly string[]).includes(value)
+
+// Used when reading legacy persisted state where invalid input should remain unset.
+export const coerceLegacySidePanelTab = (value: unknown): RightPanelTab | undefined => {
+  if (value === "changes") return "review"
+  return isRightPanelTab(value) ? value : undefined
+}
+
```
Astro-Han, line 66: The clamping logic here correctly prevents moving tabs before the pinned "status" tab. However, the `Math.min(Math.max(to, 1), state.openShellTabs.length - 1)` calculation doesn't have an explanatory comment. Adding: "Clamp to valid range: minimum index 1 keeps tabs after pinned status, maximum is last position" would make the constraint explicit for future readers.

packages/app/src/pages/session/close-session-tab.ts
```diff
@@ -0,0 +1,30 @@
+import type { RightPanelTab } from "@/pages/session/right-panel-tabs"
+
+type CloseSessionTabInput = {
+  closableTab: () => string | undefined
+  closeFileTab: (tab: string) => void
+  sidePanelOpened: () => boolean
+  sidePanelTab: () => RightPanelTab
+  closeShellTab: (tab: RightPanelTab) => void
+}
+
```
Astro-Han, line 10: The `closeSessionTab` function has a clear priority order but lacks a docstring. Adding: "Closes the active closable tab: file tabs first, then non-status shell tabs. Returns true if a tab was closed, false otherwise." would make the behavior contract explicit and match the documentation style of other helpers in this codebase.

packages/app/src/pages/session/migrate-session-view.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect((out.sessionTabs as any).a).toEqual({ all: ["file://x"], active: undefined })
+  })
+
+  test("normalizes damaged openShellTabs", () => {
+    const out = migrateSessionView(
+      { a: { scroll: {}, openShellTabs: ["review", "status", "files", "files"], sidePanelTab: "review" } },
+      {},
+    )
+    expect((out.sessionView as any).a.openShellTabs).toEqual(["status", "review", "files"])
+    expect((out.sessionView as any).a.sidePanelTab).toBe("review")
```
Astro-Han, line 148: Test "is idempotent on migrated data" correctly verifies that running migration twice produces identical output. Consider also adding a test for the edge case where `openShellTabs` is a valid array but `sidePanelTab` is an invalid value (e.g., "bogus") - the current tests cover "falls back to status when sidePanelTab not in list" but not the combined scenario of valid tabs + invalid active.

packages/app/src/pages/session/migrate-session-view.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    const rawOpenShellTabs = Array.isArray(entry.rawOpenShellTabs) ? entry.rawOpenShellTabs : undefined
+    const viewShapeChanged =
+      !entry.sessionViewRaw ||
+      !rawOpenShellTabs ||
+      !same(rawOpenShellTabs, normalized.openShellTabs) ||
+      entry.sessionViewRaw.sidePanelTab !== normalized.sidePanelTab
+    const tabsShapeChanged = sessionTabs[key] !== entry.sessionTabsRaw
+
```
Astro-Han: The `tabsShapeChanged` detection checks if `sessionTabs[key] !== entry.sessionTabsRaw`, but this reference comparison may miss cases where the array contents changed but the object reference is the same. Consider using `same()` for deeper comparison here, similar to how `viewShapeChanged` uses `same()` for `openShellTabs` comparison.

packages/app/src/pages/session/right-panel-tabs.ts
```diff
@@ -1,10 +1,146 @@
...
 
+export const RIGHT_PANEL_TAB_VALUES: readonly RightPanelTab[] = [
+  "status",
+  "files",
+  "review",
+  "terminal",
+  "context",
+] as const
+
+export type RightPanelShellIconName = "status" | "folder" | "review" | "terminal"
```
Astro-Han, line 11: The `RightPanelShellIconName` type alias duplicates the icon name union that's partially defined in `ShellTabIcon`. This circular dependency is fine for type safety, but consider renaming `RightPanelShellIconName` to `ShellIconName` for brevity since the "RightPanel" prefix is redundant given the module context.

packages/app/src/pages/session/migrate-session-view.ts
Astro-Han: Fixed in the latest push. Consolidated the duplicate object guard into `@/utils/is-record` and switched both migration and layout code to use it.

packages/app/src/pages/session/migrate-session-view.ts
Astro-Han: Fixed in the latest push. `resolveActiveCandidate` is now exported and has direct precedence tests for context, review/changes, stored side-panel fallback, and inferred review fallback.

packages/app/src/pages/session/session-side-panel.tsx
Astro-Han: Fixed in the latest push. Updated the helper docstring to clarify that this names the file-opening transition where Review must be active before file-specific content is shown.

packages/app/src/pages/session/helpers.ts
Astro-Han: Fixed in the latest push. Added a short comment documenting the close priority: active file tab first, then shell tab fallback.

packages/app/src/pages/session/terminal-shell-tab.ts
Astro-Han: Fixed in the latest push. Reworded the comment to explicitly call out responsive terminal UI feedback and the idempotent synchronization effect in `session-side-panel.tsx`.

packages/app/src/pages/session/right-panel-tabs.ts
Astro-Han: Fixed in the latest push. Added the clamp comment explaining index 1 keeps moves after pinned Status and the upper bound is the last open tab position.

packages/app/src/pages/session/migrate-session-view.ts
Astro-Han: Fixed in the latest push. Replaced the reference comparison with an explicit `tabsShapeChanged` flag set at the point where migration rewrites the session tabs entry.

packages/app/src/pages/session/close-session-tab.ts
Astro-Han: Fixed in the latest push. Added a docstring to `closeSessionTab` documenting the close order and return contract.

packages/app/src/pages/session/right-panel-tabs.ts
Astro-Han: Leaving this as-is for this PR. `RightPanelShellIconName` is more explicit at import sites outside `right-panel-tabs.ts`, where a generic `ShellIconName` could be confused with other shell UI concepts. No functional issue here.

packages/app/src/pages/session/migrate-session-view.test.ts
Astro-Han: Fixed in the latest push. Added a migration regression test for valid `openShellTabs` with an invalid `sidePanelTab`, asserting the active tab resets to Status while preserving the valid shell tabs.

packages/app/src/components/prompt-input.tsx
```diff
@@ -205,16 +205,14 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
...
       view().sidePanel.explorer.setTab("changes")
       tabs().setActive("review")
       queueCommentFocus()
       return
     }
 
-    if (!view().sidePanel.opened()) view().sidePanel.open()
-    view().sidePanel.setTab("changes")
+    view().sidePanel.openTab("review")
     view().sidePanel.explorer.setTab("all")
```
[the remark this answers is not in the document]
Astro-Han: Fixed in the latest push. Hoisted `view().sidePanel.openTab("review")` before the branch so the shared action runs once.

packages/app/e2e/files/file-tree.spec.ts
```diff
@@ -14,16 +13,11 @@ test("@smoke review tab no longer renders the legacy file-tree sub-panel", async
...
-    const rightToggle = page.locator(`${titlebarRightSelector} button`).first()
-    const rightPanel = page.locator("#right-panel")
-    const shellTabList = rightPanel.getByRole("tablist").first()
-
-    await expect(rightToggle).toBeVisible()
-    await rightToggle.click()
-    await expect(rightPanel).toHaveAttribute("aria-hidden", "false")
+    const shellTabList = page.getByRole("tablist").first()
+    await shellTabList.locator("button").last().click()
+    await page.getByRole("menuitem", { name: "Review" }).click()
```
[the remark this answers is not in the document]
Astro-Han: Fixed in the latest push. Scoped the tablist lookup through the named right utility panel landmark before selecting the shell tab list. Verified with `bun test:e2e:local -- e2e/files/file-tree.spec.ts`.

packages/app/src/pages/session/migrate-session-view.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (entry.sessionTabsRaw && (!allHasOnlyFiles || !activeIsValidInnerTab)) {
+      const all = fileTabsOnly(rawAll)
+      const active =
+        rawActive === "changes" || rawActive === "review"
+          ? "review"
+          : rawActive && rawActive.startsWith("file://") && all.includes(rawActive)
+            ? rawActive
+            : undefined
+      sessionTabs[key] = { ...entry.sessionTabsRaw, all, active }
+      tabsShapeChanged = true
```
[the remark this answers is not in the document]
Astro-Han: Fixed in the latest push. File active tabs are now only valid when present in the filtered file tab list. Added a regression test for `{ all: ["file://x"], active: "file://missing" }`.

packages/app/src/context/layout.tsx
```diff
@@ -148,103 +170,123 @@ const normalizeStoredSessionTabs = (key: string, tabs: SessionTabs) => {
...
+  })()
+
+  const rightPanel = value.rightPanel
+  const migratedRightPanel = (() => {
+    if (typeof rightPanel === "boolean") return rightPanel
+    const opened = legacyRightPanelOpened(rightPanel, review, fileTree)
+    if (!isRecord(rightPanel)) return { width: DEFAULT_RIGHT_PANEL_WIDTH, opened }
+    if (typeof rightPanel.opened === "boolean") return rightPanel
+    return { ...rightPanel, opened }
+  })()
```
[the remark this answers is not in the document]
Astro-Han: Fixed in the latest push. Legacy boolean `rightPanel` now migrates into `{ width, opened }` and preserves the boolean value. Added a regression test for `rightPanel: false`.

packages/app/src/pages/session/session-side-panel.tsx
```diff
@@ -187,6 +214,19 @@ export function SessionSidePanel(props: {
...
+    const { draggable, droppable } = event
+    if (!draggable || !droppable) return
+
+    const from = draggable.id.toString()
+    const to = droppable.id.toString()
+    if (!isRightPanelTab(from) || !isRightPanelTab(to)) return
+
+    const toIndex = view().sidePanel.openTabs().indexOf(to)
+    if (toIndex === -1) return
+    view().sidePanel.moveTab(from, toIndex)
```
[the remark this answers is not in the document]
Astro-Han: Fixed in the latest push. Shell drag now computes the target through `getTabReorderIndex`, matching the file-tab reorder path and avoiding direct tablist index handling in the component.

packages/app/src/pages/session/helpers.ts
```diff
@@ -59,14 +57,16 @@ export const createSessionTabs = (input: TabsInput) => {
...
-    if (active === "context") return active
-    if (!openedTabs().includes(active)) return
-    return active
+    // File tabs take close priority; shell tabs only apply when no file tab is active.
+    const active = input.tabs().active()
+    const normalizedActive = active ? input.normalizeTab(active) : undefined
+    if (normalizedActive && input.pathFromTab(normalizedActive)) return normalizedActive
+    const current = activeTab()
+    if (openedTabs().includes(current)) return current
+    return undefined
```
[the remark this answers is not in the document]
Astro-Han: Fixed in the latest push. `closableTab()` now requires the normalized active file tab to still exist in `openedTabs()` before returning it. Added a regression test for stale active file tabs.

packages/app/src/pages/session/close-session-tab.ts
```diff
@@ -0,0 +1,31 @@
...
+export function closeSessionTab(input: CloseSessionTabInput): boolean {
+  const fileTab = input.closableTab()
+  if (fileTab) {
+    input.closeFileTab(fileTab)
+    return true
+  }
+
+  const shellTab = input.sidePanelTab()
+  if (!input.sidePanelOpened() || shellTab === "status") return false
+  input.closeShellTab(shellTab)
```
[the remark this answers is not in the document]
Astro-Han: Fixed in the latest push. `close-session-tab.ts` now imports `RIGHT_PANEL_TAB_META` and uses the canonical `closable` metadata for both command enablement and shell-tab close behavior.

PR #138, opened 2026-04-22
fix: sync runtime import handling

Astro-Han:
## Summary
Sync the #92 Runtime Resource lane slice for upstream `92c005866`, normalizing local provider and local tool dynamic imports through `file://` URLs.

## Why
Part of #92. Upstream `v1.14.20` fixes local dynamic imports on Windows and Node because Node rejects raw Windows filesystem paths in dynamic `import()`. PawWork needs this for Windows plugin and tool loading reliability.

## Related Issue
Part of #92.

## How To Verify

```bash
(cd packages/opencode && bun test test/provider/provider.test.ts -t "custom provider with npm package|local provider import spec normalizes filesystem paths to file URLs")
(cd packages/opencode && bun test test/tool/registry.test.ts -t "loads tools from .opencode/tool|loads tools from .opencode/tools|local tool import spec normalizes filesystem paths to file URLs")
(cd packages/opencode && bun test test/provider/provider.test.ts test/tool/registry.test.ts)
(cd packages/opencode && bun run typecheck)
git diff --check origin/dev...HEAD
rg -n 'import\(installedPath\)|process\.platform === "win32" \? match' packages/opencode/src/provider/provider.ts packages/opencode/src/tool/registry.ts
```

Expected for the final `rg` command: no output.

## Screenshots or Recordings
Not applicable. No UI changes.

## Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -50,6 +50,10 @@ import { Agent } from "../agent/agent"
 import { Skill } from "../skill"
 import { needsConfigDependencies, usesConfigDependencies } from "../config/dependency"
 
+export function localToolImportSpec(input: string) {
+  return input.startsWith("file://") ? input : pathToFileURL(input).href
+}
+
 export namespace ToolRegistry {
   const log = Log.create({ service: "tool.registry" })
 
@@ -177,7 +181,7 @@ export namespace ToolRegistry {
               ...Permission.disabled(ids, rules),
             ])
             if (ids.length && ids.every((id) => disabled.has(id))) continue
-            const spec = process.platform === "win32" ? match : pathToFileURL(match).href
+            const spec = localToolImportSpec(match)
             const configDir = path.dirname(path.dirname(match))
             const usesDeps = yield* Effect.promise(() => usesConfigDependencies(match))
             if (usesDeps && depsFailed.has(configDir)) continue
--- a/packages/opencode/test/tool/registry.test.ts
+++ b/packages/opencode/test/tool/registry.test.ts
@@ -5,7 +5,7 @@ import { tmpdir } from "../fixture/fixture"
 import { writeMockConfigInstall } from "../shared/mock-npm-install"
 import { withConfigDepsLock } from "../shared/config-deps-lock"
 import { Instance } from "../../src/project/instance"
-import { ToolRegistry } from "../../src/tool/registry"
+import { localToolImportSpec, ToolRegistry } from "../../src/tool/registry"
 import { Npm } from "../../src/npm"
 
 afterEach(async () => {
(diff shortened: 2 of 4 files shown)


packages/opencode/src/tool/registry.ts
```diff
@@ -50,6 +50,10 @@ import { Agent } from "../agent/agent"
 import { Skill } from "../skill"
 import { needsConfigDependencies, usesConfigDependencies } from "../config/dependency"
 
+export function localToolImportSpec(input: string) {
+  return pathToFileURL(input).href
```
Astro-Han: **Missing `file://` prefix check**

`localProviderImportSpec` checks for `startsWith("file://")` before calling `pathToFileURL`, but this function does not. The two functions have similar names and similar purpose, yet behave differently.

While `match` currently always comes from `Glob.scanSync` with `absolute: true`, this inconsistency is a latent bug risk. If future refactoring changes the input source, `pathToFileURL("file://...")` produces malformed URLs:

```js
pathToFileURL("file:///tmp/test.js").href
// → "file:///current/dir/file:/tmp/test.js" (wrong!)```

Suggestion: either add the same check for consistency, or document why it's intentionally omitted.

---
Compare with `localProviderImportSpec` in provider.ts which handles this correctly:
```ts
export function localProviderImportSpec(input: string) {
  return input.startsWith("file://") ? input : pathToFileURL(input).href
}
```

packages/opencode/test/tool/registry.test.ts
```diff
@@ -93,6 +93,14 @@ describe("tool.registry", () => {
     })
   })
 
+  test("local tool import spec normalizes filesystem paths to file URLs", () => {
+    const toolPath = path.resolve("pawwork-tools", "marked.ts")
+    expect(localToolImportSpec(toolPath)).toStartWith("file://")
+    expect(localToolImportSpec(toolPath)).not.toBe(toolPath)
```
Astro-Han, line 99: **Test coverage gap: missing `file://` input test**

The `localProviderImportSpec` tests include a case for `file://` URL input:
```ts
expect(localProviderImportSpec("file:///tmp/pawwork-provider/index.js")).toBe(
  "file:///tmp/pawwork-provider/index.js",
)
```

But this test suite lacks the equivalent case for `localToolImportSpec`. Even though such input is not currently expected, the test would document the function's behavior boundary and catch regressions if the input assumptions change.

Consider adding:
```ts
expect(localToolImportSpec("file:///tmp/tool.ts")).toBe("file:///tmp/tool.ts")
```

Astro-Han (commented):
## Review Summary

Code mostly looks good for the upstream sync, but found an inconsistency between the two import spec functions that should be addressed for long-term maintainability.

### Key Issue: Function Behavior Inconsistency

`localProviderImportSpec` checks for `file://` prefix to avoid double-conversion, but `localToolImportSpec` does not. While `match` is currently always a filesystem path (from Glob.scanSync), this inconsistency could bite us during future refactoring.

If `pathToFileURL` receives an already-formatted `file://` URL, it treats it as a literal file path and produces incorrect output.

packages/opencode/test/tool/registry.test.ts
Astro-Han: Fixed in e0c2c2364. Added `file:///tmp/tool.ts` coverage to document the idempotent URL case.

packages/opencode/src/tool/registry.ts
Astro-Han: Fixed in e0c2c2364. `localToolImportSpec` now preserves existing `file://` inputs before falling back to `pathToFileURL`.

PR #140, opened 2026-04-22
refactor: sync session runtime lane

Astro-Han:
## Summary
- Sync the #92 Session lane for retained recent turns during compaction, including `tail_start_id` on compaction parts and compacted message filtering for successful summaries and pending compactions.
- Make tool-output pruning explicit opt-in via `compaction.prune`, and update config, SDK, and OpenAPI metadata.
- Fix v2 session entry reduction so interleaved tool events route by `callID`, retry attempts are recorded with structured error context, and session error schema access is deferred with `z.lazy`.

## Why
Part of #92. This lands the session runtime slice after the Base PR #102, while keeping Runtime Resource and Desktop work out of scope.

## Compatibility Notes
- Existing configs that still want automatic old tool-output pruning should set `compaction.prune: true`.
- v2 `retried` events now store structured retry errors, and legacy string retry errors are still accepted and normalized while reducing session entries.

## Related Issue
Part of #92.

## How To Verify

```bash
bun test test/session/compaction.test.ts test/session/messages-pagination.test.ts test/session/session-entry.test.ts test/session/session.test.ts
cd packages/opencode && bun run typecheck
cd packages/sdk/js && bun run typecheck
```

Latest local results after autosquash:

- Focused session tests: 115 pass, 0 fail
- `packages/opencode` typecheck: pass
- `packages/sdk/js` typecheck: pass
- `git diff --check origin/dev...HEAD`: pass

Note: full `bun turbo typecheck` currently fails in `@opencode-ai/desktop-electron` on `@opencode-ai/util/file-extensions` module resolution. This is outside this PR diff, which does not touch `packages/desktop-electron` or `packages/util`.

## Screenshots or Recordings

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Intelligent tail-budgeting during compaction to preserve recent conversation turns.
  * Added structured assistant retry records with attempt counts and rich error details.
  * Compaction parts can mark a tail boundary (tail_start_id) to retain recent messages verbatim.

* **Bug Fixes**
  * Fixed tool state routing for interleaved concurrent tool calls.
  * Corrected overflow/headroom calculation for content management.

* **Configuration**
  * compaction.prune now defaults to false; set true to enable pruning.

* **Tests**
  * Added comprehensive compaction, pagination, retry, and schema validation tests.

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -255,7 +255,7 @@ const InfoSchema = Schema.Struct({
         description: "Enable automatic compaction when context is full (default: true)",
       }),
       prune: Schema.optional(Schema.Boolean).annotate({
-        description: "Enable pruning of old tool outputs (default: true)",
+        description: "Prune old tool outputs (default: `false`). Set `true` to enable.",
       }),
       tail_turns: Schema.optional(NonNegativeInt).annotate({
         description:
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -2,7 +2,7 @@ import { BusEvent } from "@/bus/bus-event"
 import { Bus } from "@/bus"
 import * as Session from "./session"
 import { SessionID, MessageID, PartID } from "./schema"
-import { Provider } from "../provider"
+import { Provider, ProviderTransform } from "../provider"
 import { MessageV2 } from "./message-v2"
 import z from "zod"
 import { Token } from "../util/token"
@@ -32,6 +32,12 @@ export const Event = {
 export const PRUNE_MINIMUM = 20_000
 export const PRUNE_PROTECT = 40_000
 const PRUNE_PROTECTED_TOOLS = ["skill"]
+// Keep the last exchange pair by default so a resumed session keeps the active request
+// and response intact. The token floor handles small contexts, while the ceiling keeps
+// retained text from crowding out the generated summary on larger models.
+const DEFAULT_TAIL_TURNS = 2
+const MIN_TAIL_TOKENS = 2_000
+const MAX_TAIL_TOKENS = 8_000
 export const CreateInput = z.object({
   sessionID: SessionID.zod,
   agent: z.string(),
@@ -67,6 +73,20 @@ export interface Interface {
 
 export class Service extends Context.Service<Service, Interface>()("@opencode/SessionCompaction") {}
 
+function usable(input: { cfg: Config.Info; model: Provider.Model }) {
+  const reserved = input.cfg.compaction?.reserved ?? Math.min(20_000, ProviderTransform.maxOutputTokens(input.model))
+  return input.model.limit.input
+    ? Math.max(0, input.model.limit.input - reserved)
+    : Math.max(0, input.model.limit.context - ProviderTransform.maxOutputTokens(input.model))
+}
+
+function tailBudget(input: { cfg: Config.Info; model: Provider.Model }) {
+  return (
+    input.cfg.compaction?.preserve_recent_tokens ??
+    Math.min(MAX_TAIL_TOKENS, Math.max(MIN_TAIL_TOKENS, Math.floor(usable(input) * 0.25)))
+  )
+}
+
 export const layer: Layer.Layer<
   Service,
   never,
@@ -93,6 +113,61 @@ export const layer: Layer.Layer<
       model: Provider.Model
     }) {
       return overflow({ cfg: yield* config.get(), tokens: input.tokens, model: input.model })
+    })
+
+    const estimate = Effect.fn("SessionCompaction.estimate")(function* (input: {
+      messages: MessageV2.WithParts[]
+      model: Provider.Model
+    }) {
+      const msgs = yield* MessageV2.toModelMessagesEffect(input.messages, input.model, { stripMedia: true })
+      // JSON adds structural overhead, so this intentionally overestimates instead of risking overflow.
+      return Token.estimate(JSON.stringify(msgs))
+    })
+
+    const select = Effect.fn("SessionCompaction.select")(function* (input: {
+      messages: MessageV2.WithParts[]
+      cfg: Config.Info
+      model: Provider.Model
+    }) {
+      const limit = input.cfg.compaction?.tail_turns ?? DEFAULT_TAIL_TURNS
+      if (limit <= 0) return { head: input.messages, tail_start_id: undefined }
+      const budget = tailBudget({ cfg: input.cfg, model: input.model })
+      const userStarts = input.messages.flatMap((msg, idx) => (msg.info.role === "user" ? [idx] : []))
+      const turns = input.messages.flatMap((msg, idx) =>
+        msg.info.role === "user" && !msg.parts.some((part) => part.type === "compaction") ? [idx] : [],
+      )
+      if (!turns.length) return { head: input.messages, tail_start_id: undefined }
+
+      let total = 0
+      let start = input.messages.length
+      let kept = 0
+
+      for (let i = turns.length - 1; i >= 0 && kept < limit; i--) {
+        const idx = turns[i]
+        const userIdx = userStarts.indexOf(idx)
+        const end = userIdx >= 0 && userIdx + 1 < userStarts.length ? userStarts[userIdx + 1] : input.messages.length
+        const size = yield* estimate({
+          messages: input.messages.slice(idx, end),
+          model: input.model,
+        })
+        if (kept === 0 && size > budget) {
+          // The latest turn must stay intact or be summarized with the rest.
+          // Keeping a partial turn would separate tool calls from their prompt.
+          log.info("tail fallback", { budget, size })
+          return { head: input.messages, tail_start_id: undefined }
+        }
+        if (total + size > budget) break
+        total += size
+        start = idx
+        kept++
+      }
+
+      // If the tail starts at the first message, there is no older head to summarize.
+      if (kept === 0 || start === 0) return { head: input.messages, tail_start_id: undefined }
+      return {
+        head: input.messages.slice(0, start),
+        tail_start_id: input.messages[start]?.info.id,
+      }
     })
 
     // goes backwards through parts until there are PRUNE_PROTECT tokens worth of tool
@@ -157,6 +232,7 @@ export const layer: Layer.Layer<
         throw new Error(`Compaction parent must be a user message: ${input.parentID}`)
       }
       const userMessage = parent.info
+      const compactionPart = parent.parts.find((part): part is MessageV2.CompactionPart => part.type === "compaction")
 
       let messages = input.messages
       let replay:
@@ -187,6 +263,15 @@ export const layer: Layer.Layer<
       const model = agent.model
         ? yield* provider.getModel(agent.model.providerID, agent.model.modelID)
         : yield* provider.getModel(userMessage.model.providerID, userMessage.model.modelID)
+      const cfg = yield* config.get()
+      // Drop the new compaction marker itself; otherwise it appears in the final turn's
+      // slice and inflates the retained-tail token estimate.
+      const history = compactionPart ? messages.filter((message) => message.info.id !== input.parentID) : messages
+      const selected = yield* select({
+        messages: history,
+        cfg,
+        model,
+      })
       // Allow plugins to inject context or replace compaction prompt.
       const compacting = yield* plugin.trigger(
         "experimental.session.compacting",
@@ -224,7 +309,7 @@ When constructing the summary, try to stick to this template:
 ---`
 
       const prompt = compacting.prompt ?? [defaultPrompt, ...compacting.context].join("\n\n")
-      const msgs = structuredClone(messages)
+      const msgs = structuredClone(selected.head)
       yield* plugin.trigger("experimental.chat.messages.transform", {}, { messages: msgs })
       const modelMessages = yield* MessageV2.toModelMessagesEffect(msgs, model, { stripMedia: true })
       const ctx = yield* InstanceState.context
@@ -285,6 +370,14 @@ When constructing the summary, try to stick to this template:
         processor.message.finish = "error"
         yield* session.updateMessage(processor.message)
         return "stop"
+      }
+
+      if (compactionPart && compactionPart.tail_start_id !== selected.tail_start_id) {
+        // Repeated compactions can move the retained-tail boundary forward as older turns accumulate.
+        yield* session.updatePart({
+          ...compactionPart,
+          tail_start_id: selected.tail_start_id,
+        })
       }
 
       if (result === "continue" && input.auto) {
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -173,7 +173,7 @@ export const layer: Layer.Layer<
     // calls, then erases output of older tool calls to free context space
     const prune = Effect.fn("SessionCompaction.prune")(function* (input: { sessionID: SessionID }) {
       const cfg = yield* config.get()
-      if (cfg.compaction?.prune === false) return
+      if (cfg.compaction?.prune !== true) return
       log.info("pruning")
 
       const msgs = yield* session
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -75,9 +75,8 @@ export class Service extends Context.Service<Service, Interface>()("@opencode/Se
 
 function usable(input: { cfg: Config.Info; model: Provider.Model }) {
   const reserved = input.cfg.compaction?.reserved ?? Math.min(20_000, ProviderTransform.maxOutputTokens(input.model))
-  return input.model.limit.input
-    ? Math.max(0, input.model.limit.input - reserved)
-    : Math.max(0, input.model.limit.context - ProviderTransform.maxOutputTokens(input.model))
+  const base = input.model.limit.input ?? input.model.limit.context
+  return Math.max(0, base - reserved)
 }
 
 function tailBudget(input: { cfg: Config.Info; model: Provider.Model }) {
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -210,6 +210,9 @@ export const CompactionPart = PartBase.extend({
   type: z.literal("compaction"),
   auto: z.boolean(),
   overflow: z.boolean().optional(),
+  tail_start_id: MessageID.zod
+    .describe("Message ID of the first retained message kept verbatim after compaction.")
+    .optional(),
 }).meta({
   ref: "CompactionPart",
 })
@@ -926,14 +929,26 @@ export function get(input: { sessionID: SessionID; messageID: MessageID }): With
 export function filterCompacted(msgs: Iterable<WithParts>) {
   const result = [] as WithParts[]
   const completed = new Set<string>()
+  let tailStartID: MessageID | undefined
+  // Read newest to oldest. The first tail marker wins, even before its summary finishes.
   for (const msg of msgs) {
     result.push(msg)
-    if (msg.info.role === "user" && completed.has(msg.info.id) && msg.parts.some((part) => part.type === "compaction"))
+    const compactionPart = msg.parts.find((part): part is CompactionPart => part.type === "compaction")
+    tailStartID ??= compactionPart?.tail_start_id
+    if (tailStartID && msg.info.id === tailStartID) break
+    if (msg.info.role === "user" && completed.has(msg.info.id) && compactionPart) {
+      if (tailStartID) continue
       break
+    }
     if (msg.info.role === "assistant" && msg.info.summary && msg.info.finish && !msg.info.error)
       completed.add(msg.info.parentID)
   }
   result.reverse()
+  if (tailStartID) {
+    // After reversing to chronological order, drop everything before the retained tail.
+    const idx = result.findIndex((msg) => msg.info.id === tailStartID)
+    if (idx >= 0) return result.slice(idx)
+  }
   return result
 }
 
--- a/packages/opencode/src/session/overflow.ts
+++ b/packages/opencode/src/session/overflow.ts
@@ -15,8 +15,6 @@ export function isOverflow(input: { cfg: Config.Info; tokens: MessageV2.Assistan
 
   const reserved =
     input.cfg.compaction?.reserved ?? Math.min(COMPACTION_BUFFER, ProviderTransform.maxOutputTokens(input.model))
-  const usable = input.model.limit.input
-    ? input.model.limit.input - reserved
-    : context - ProviderTransform.maxOutputTokens(input.model)
+  const usable = Math.max(0, (input.model.limit.input ?? context) - reserved)
   return count >= usable
 }
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -258,7 +258,8 @@ export const Event = {
     "session.error",
     z.object({
       sessionID: SessionID.zod.optional(),
-      error: MessageV2.Assistant.shape.error,
+      // z.lazy avoids eager access to MessageV2.Assistant while session.ts is still initializing.
+      error: z.lazy(() => MessageV2.Assistant.shape.error),
     }),
   ),
 }
--- a/packages/opencode/src/v2/session-entry.ts
+++ b/packages/opencode/src/v2/session-entry.ts
@@ -96,6 +96,29 @@ export namespace SessionEntry {
     text: Schema.String,
   }) {}
 
+  export class AssistantRetry extends Schema.Class<AssistantRetry>("Session.Entry.Assistant.Retry")({
+    attempt: Schema.Number,
+    error: SessionEvent.RetryError,
+    time: Schema.Struct({
+      created: Schema.DateTimeUtc,
+    }),
+  }) {
+    static fromEvent(event: SessionEvent.Retried) {
+      const error =
+        typeof event.error === "string"
+          // Legacy retry events stored only the message, so retryability cannot be recovered.
+          ? new SessionEvent.RetryError({ message: event.error, isRetryable: true })
+          : event.error
+      return new AssistantRetry({
+        attempt: event.attempt ?? 0,
+        error,
+        time: {
+          created: event.timestamp,
+        },
+      })
+    }
+  }
+
   export const AssistantContent = Schema.Union([AssistantText, AssistantReasoning, AssistantTool])
   export type AssistantContent = Schema.Schema.Type<typeof AssistantContent>
 
@@ -103,6 +126,7 @@ export namespace SessionEntry {
     ...Base,
     type: Schema.Literal("assistant"),
     content: AssistantContent.pipe(Schema.Array),
+    retries: AssistantRetry.pipe(Schema.Array, Schema.optional),
     cost: Schema.Number.pipe(Schema.optional),
     tokens: Schema.Struct({
       input: Schema.Number,
@@ -280,6 +304,11 @@ export namespace SessionEntry {
         if (!pendingAssistant) break
         const match = lastReasoning(pendingAssistant)
         if (match) match.text = event.text
+        break
+      }
+      case "retried": {
+        if (!pendingAssistant) break
+        pendingAssistant.retries = [...(pendingAssistant.retries ?? []), AssistantRetry.fromEvent(event)]
         break
       }
       case "step.ended": {
--- a/packages/opencode/src/v2/session-entry.ts
+++ b/packages/opencode/src/v2/session-entry.ts
@@ -146,8 +146,12 @@ export namespace SessionEntry {
     )
     const lastText = (assistant: Mutable<Assistant>) =>
       assistant.content.findLast((x: Mutable<AssistantContent>): x is Mutable<AssistantText> => x.type === "text")
-    const lastTool = (assistant: Mutable<Assistant>) =>
-      assistant.content.findLast((x: Mutable<AssistantContent>): x is Mutable<AssistantTool> => x.type === "tool")
+    // Tool events can interleave, so state updates must target the matching call.
+    const lastTool = (assistant: Mutable<Assistant>, callID: string) =>
+      assistant.content.findLast(
+        (x: Mutable<AssistantContent>): x is Mutable<AssistantTool> =>
+          x.type === "tool" && x.callID === callID,
+      )
     const lastReasoning = (assistant: Mutable<Assistant>) =>
       assistant.content.findLast(
         (x: Mutable<AssistantContent>): x is Mutable<AssistantReasoning> => x.type === "reasoning",
@@ -211,7 +215,7 @@ export namespace SessionEntry {
       }
       case "tool.input.delta": {
         if (!pendingAssistant) break
-        const match = lastTool(pendingAssistant)
+        const match = lastTool(pendingAssistant, event.callID)
         if (match && match.state.status === "pending") match.state.input += event.delta
         break
       }
@@ -220,7 +224,7 @@ export namespace SessionEntry {
       }
       case "tool.called": {
         if (!pendingAssistant) break
-        const match = lastTool(pendingAssistant)
+        const match = lastTool(pendingAssistant, event.callID)
         if (match) {
           match.time.ran = event.timestamp
           match.state = {
@@ -232,7 +236,7 @@ export namespace SessionEntry {
       }
       case "tool.success": {
         if (!pendingAssistant) break
-        const match = lastTool(pendingAssistant)
+        const match = lastTool(pendingAssistant, event.callID)
         if (match && match.state.status === "running") {
           match.state = {
             status: "completed",
@@ -247,7 +251,7 @@ export namespace SessionEntry {
       }
       case "tool.error": {
         if (!pendingAssistant) break
-        const match = lastTool(pendingAssistant)
+        const match = lastTool(pendingAssistant, event.callID)
         if (match && match.state.status === "running") {
           match.state = {
             status: "error",
--- a/packages/opencode/src/v2/session-event.ts
+++ b/packages/opencode/src/v2/session-event.ts
@@ -49,6 +49,17 @@ export namespace SessionEvent {
     source: Source.pipe(Schema.optional),
   }) {}
 
+  export class RetryError extends Schema.Class<RetryError>("Session.Event.Retry.Error")({
+    message: Schema.String,
+    statusCode: Schema.Number.pipe(Schema.optional),
+    isRetryable: Schema.Boolean,
+    responseHeaders: Schema.Record(Schema.String, Schema.String).pipe(Schema.optional),
+    responseBody: Schema.String.pipe(Schema.optional),
+    metadata: Schema.Record(Schema.String, Schema.String).pipe(Schema.optional),
+  }) {}
+  // Accept legacy string errors while new retry events carry structured context.
+  const RetryEventError = Schema.Union([RetryError, Schema.String])
+
   export class Prompt extends Schema.Class<Prompt>("Session.Event.Prompt")({
     ...Base,
     type: Schema.Literal("prompt"),
@@ -382,14 +393,16 @@ export namespace SessionEvent {
   export class Retried extends Schema.Class<Retried>("Session.Event.Retried")({
     ...Base,
     type: Schema.Literal("retried"),
-    error: Schema.String,
+    attempt: Schema.Number.pipe(Schema.optional),
+    error: RetryEventError,
   }) {
-    static create(input: BaseInput & { error: string }) {
+    static create(input: BaseInput & { attempt?: number; error: RetryError | string }) {
       return new Retried({
         id: input.id ?? ID.create(),
         type: "retried",
         timestamp: input.timestamp ?? DateTime.makeUnsafe(Date.now()),
         metadata: input.metadata,
+        ...(input.attempt !== undefined ? { attempt: input.attempt } : {}),
         error: input.error,
       })
     }
--- a/packages/opencode/test/session/compaction.test.ts
+++ b/packages/opencode/test/session/compaction.test.ts
@@ -540,6 +540,23 @@ describe("session.compaction.isOverflow", () => {
     ),
   )
 
+  it.live(
+    "respects reserved override without input caps",
+    provideTmpdirInstance(
+      () =>
+        Effect.gen(function* () {
+          const compact = yield* SessionCompaction.Service
+          const model = createModel({ context: 100_000, output: 10_000 })
+          const tokens = { input: 45_000, output: 5_000, reasoning: 0, cache: { read: 0, write: 0 } }
+          expect(yield* compact.isOverflow({ tokens, model })).toBe(true)
+        }),
+      {
+        config: {
+          compaction: { reserved: 50_000 },
+        },
+      },
+    ),
+  )
 })
 
 describe("session.compaction.create", () => {
--- a/packages/opencode/test/session/compaction.test.ts
+++ b/packages/opencode/test/session/compaction.test.ts
@@ -593,7 +593,7 @@ describe("session.compaction.create", () => {
 
 describe("session.compaction.prune", () => {
   it.live(
-    "compacts old completed tool output",
+    "keeps old completed tool output by default",
     provideTmpdirInstance((dir) =>
       Effect.gen(function* () {
         const compact = yield* SessionCompaction.Service
@@ -676,9 +676,97 @@ describe("session.compaction.prune", () => {
         expect(part?.type).toBe("tool")
         expect(part?.state.status).toBe("completed")
         if (part?.type === "tool" && part.state.status === "completed") {
-          expect(part.state.time.compacted).toBeNumber()
+          expect(part.state.time.compacted).toBeUndefined()
         }
       }),
+    ),
+  )
+
+  it.live(
+    "compacts old completed tool output when prune is explicitly enabled",
+    provideTmpdirInstance((dir) =>
+      Effect.gen(function* () {
+        const compact = yield* SessionCompaction.Service
+        const ssn = yield* SessionNs.Service
+        const session = yield* ssn.create({})
+        const first = yield* ssn.updateMessage({
+          id: MessageID.ascending(),
+          role: "user",
+          sessionID: session.id,
+          agent: "build",
+          model: ref,
+          time: { created: Date.now() },
+        })
+        yield* ssn.updatePart({
+          id: PartID.ascending(),
+          messageID: first.id,
+          sessionID: session.id,
+          type: "text",
+          text: "run a large tool",
+        })
+        const response = yield* ssn.updateMessage({
+          id: MessageID.ascending(),
+          role: "assistant",
+          sessionID: session.id,
+          mode: "build",
+          agent: "build",
+          path: { cwd: dir, root: dir },
+          cost: 0,
+          tokens: {
+            output: 0,
+            input: 0,
+            reasoning: 0,
+            cache: { read: 0, write: 0 },
+          },
+          modelID: ref.modelID,
+          providerID: ref.providerID,
+          parentID: first.id,
+          time: { created: Date.now() },
+          finish: "end_turn",
+        })
+        const part = yield* ssn.updatePart({
+          id: PartID.ascending(),
+          messageID: response.id,
+          sessionID: session.id,
+          type: "tool",
+          callID: crypto.randomUUID(),
+          tool: "bash",
+          state: {
+            status: "completed",
+            input: {},
+            output: "x".repeat(200_000),
+            title: "done",
+            metadata: {},
+            time: { start: Date.now(), end: Date.now() },
+          },
+        })
+        for (const text of ["second", "third"]) {
+          const msg = yield* ssn.updateMessage({
+            id: MessageID.ascending(),
+            role: "user",
+            sessionID: session.id,
+            agent: "build",
+            model: ref,
+            time: { created: Date.now() },
+          })
+          yield* ssn.updatePart({
+            id: PartID.ascending(),
+            messageID: msg.id,
+            sessionID: session.id,
+            type: "text",
+            text,
+          })
+        }
+
+        yield* compact.prune({ sessionID: session.id })
+
+        const messages = yield* ssn.messages({ sessionID: session.id })
+        const updated = messages.flatMap((message) => message.parts).find((candidate) => candidate.id === part.id)
+        expect(updated?.type).toBe("tool")
+        if (updated?.type !== "tool" || updated.state.status !== "completed") return
+        expect(updated.state.time.compacted).toBeNumber()
+      }),
+      { config: { compaction: { prune: true } } },
     ),
   )
 
--- a/packages/opencode/test/session/compaction.test.ts
+++ b/packages/opencode/test/session/compaction.test.ts
@@ -168,7 +168,19 @@ function layer(result: "continue" | "compact") {
   )
 }
 
-function runtime(result: "continue" | "compact", plugin = Plugin.defaultLayer, provider = ProviderTest.fake()) {
+function cfg(compaction?: Config.Info["compaction"]) {
+  const base = Config.Info.parse({})
+  return Layer.mock(Config.Service)({
+    get: () => Effect.succeed({ ...base, compaction }),
+  })
+}
+
+function runtime(
+  result: "continue" | "compact",
+  plugin = Plugin.defaultLayer,
+  provider = ProviderTest.fake(),
+  config = Config.defaultLayer,
+) {
   const bus = Bus.layer
   return ManagedRuntime.make(
     Layer.mergeAll(SessionCompaction.layer, bus).pipe(
@@ -178,7 +190,7 @@ function runtime(result: "continue" | "compact", plugin = Plugin.defaultLayer, p
       Layer.provide(Agent.defaultLayer),
       Layer.provide(plugin),
       Layer.provide(bus),
-      Layer.provide(Config.defaultLayer),
+      Layer.provide(config),
     ),
   )
 }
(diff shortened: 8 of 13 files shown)


packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +112,60 @@ export const layer: Layer.Layer<
...
+      const budget = tailBudget({ cfg: input.cfg, model: input.model })
+      const turns = input.messages.flatMap((msg, idx) =>
+        msg.info.role === "user" && !msg.parts.some((part) => part.type === "compaction") ? [idx] : [],
+      )
+      if (!turns.length) return { head: input.messages, tail_start_id: undefined }
+
+      let total = 0
+      let start = input.messages.length
+      let kept = 0
+
```
Astro-Han, line 143: The fallback behavior when the first turn exceeds the budget is logged but not documented. Consider adding a comment explaining why the entire tail retention mechanism falls back when the most recent turn alone exceeds the budget. This could confuse future readers who might expect partial retention or a different handling strategy.

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +112,60 @@ export const layer: Layer.Layer<
...
+      for (let i = turns.length - 1; i >= 0 && kept < limit; i--) {
+        const idx = turns[i]
+        const end = i + 1 < turns.length ? turns[i + 1] : input.messages.length
+        const size = yield* estimate({
+          messages: input.messages.slice(idx, end),
+          model: input.model,
+        })
+        if (kept === 0 && size > budget) {
+          log.info("tail fallback", { budget, size })
+          return { head: input.messages, tail_start_id: undefined }
```
Astro-Han, line 156: The `start === 0` condition returns `{ head: input.messages, tail_start_id: undefined }`. When `start` becomes 0, it means the selection process reached the first message, which would effectively include the entire history as the 'tail'. This fallback behavior is correct but lacks a comment explaining the rationale. A reader might wonder why retaining the entire history is treated as 'no tail'.

packages/opencode/src/session/message-v2.ts
```diff
@@ -926,14 +927,30 @@ export function get(input: { sessionID: SessionID; messageID: MessageID }): With
...
   const result = [] as WithParts[]
   const completed = new Set<string>()
+  let tailStartID: MessageID | undefined
   for (const msg of msgs) {
     result.push(msg)
-    if (msg.info.role === "user" && completed.has(msg.info.id) && msg.parts.some((part) => part.type === "compaction"))
+    if (tailStartID && msg.info.id === tailStartID) break
+    const compactionPart = msg.parts.find((part): part is CompactionPart => part.type === "compaction")
+    if (msg.info.role === "user" && completed.has(msg.info.id) && compactionPart) {
+      if (compactionPart.tail_start_id) {
```
Astro-Han: The `tailStartID` tracking logic here is complex: it's set during forward iteration when encountering a compaction with `tail_start_id`, then the loop breaks when matching the ID. The subsequent reverse and slice operations rely on this behavior. Consider adding inline comments explaining the two-phase logic: (1) forward pass captures `tailStartID` from the most recent compaction, (2) reverse + slice retains messages from `tailStartID` onward.

packages/opencode/src/v2/session-entry.ts
```diff
@@ -96,13 +96,32 @@ export namespace SessionEntry {
     text: Schema.String,
   }) {}
```
Astro-Han, line 98: The `lastTool` function now accepts `callID` to route interleaved tool events by their call ID. This change fixes a bug where parallel tool calls could have their states mixed up. However, the function signature change is not commented. Consider documenting why `callID` routing is necessary for correct interleaved tool handling.

packages/opencode/src/v2/session-event.ts
```diff
@@ -49,6 +49,15 @@ export namespace SessionEvent {
...
 
+  export class RetryError extends Schema.Class<RetryError>("Session.Event.Retry.Error")({
+    message: Schema.String,
+    statusCode: Schema.Number.pipe(Schema.optional),
+    isRetryable: Schema.Boolean,
+    responseHeaders: Schema.Record(Schema.String, Schema.String).pipe(Schema.optional),
+    responseBody: Schema.String.pipe(Schema.optional),
+    metadata: Schema.Record(Schema.String, Schema.String).pipe(Schema.optional),
+  }) {}
+
```
Astro-Han, line 62: The `Retried` event schema change from `error: Schema.String` to `error: RetryError` is a breaking change. Existing persisted events with string errors will fail to parse under this new schema. Consider whether a migration step is needed for existing session data, or document that this is acceptable because `Retried` events are transient and not persisted long-term.

packages/opencode/src/config/config.ts
```diff
@@ -255,7 +255,7 @@ const InfoSchema = Schema.Struct({
         description: "Enable automatic compaction when context is full (default: true)",
```
Astro-Han, line 255: Changing `prune` default from `true` to `false` is a behavior change for existing users who relied on automatic tool output pruning. While the PR summary mentions 'explicit opt-in', consider documenting this change in release notes or a migration guide for users who expect pruning to remain active by default.

packages/opencode/test/session/compaction.test.ts
```diff
@@ -580,12 +647,52 @@ describe("session.compaction.prune", () => {
...
+        await rt.runPromise(SessionCompaction.Service.use((compact) => compact.prune({ sessionID: session.id })))
+
+        const messages = await svc.messages({ sessionID: session.id })
+        const updated = messages.flatMap((message) => message.parts).find((candidate) => candidate.id === part.id)
+        expect(updated?.type).toBe("tool")
+        if (updated?.type !== "tool" || updated.state.status !== "completed") return
+        expect(updated.state.time.compacted).toBeNumber()
+      },
+    })
+  })
```
Astro-Han: Using `JSON.stringify(captured?.messages)` for content verification relies on JSON serialization format, which may produce false positives/negatives if message structure changes. Consider directly verifying `captured.messages.some(m => m.content.includes('first request'))` instead of string matching for more robust assertions.

packages/opencode/test/session/compaction.test.ts
```diff
@@ -580,12 +647,52 @@ describe("session.compaction.prune", () => {
...
+        const first = await user(session.id, "run a large tool")
+        const response = await assistant(session.id, first.id, tmp.path)
+        const part = await svc.updatePart({
+          id: PartID.ascending(),
+          messageID: response.id,
+          sessionID: session.id,
+          type: "tool",
+          callID: crypto.randomUUID(),
+          tool: "bash",
+          state: {
```
Astro-Han, line 734: The test sets `preserve_recent_tokens: 6_000` but creates simple short messages that likely don't approach the token budget limit. Consider adding a test case where messages approach or exceed the budget to verify that the `select` function correctly falls back when the first turn exceeds `budget`, or correctly caps retained turns at the budget boundary.

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +112,60 @@ export const layer: Layer.Layer<
...
+      if (kept === 0 || start === 0) return { head: input.messages, tail_start_id: undefined }
+      return {
+        head: input.messages.slice(0, start),
+        tail_start_id: input.messages[start]?.info.id,
+      }
+    })
+
     // goes backwards through parts until there are PRUNE_PROTECT tokens worth of tool
     // calls, then erases output of older tool calls to free context space
     const prune = Effect.fn("SessionCompaction.prune")(function* (input: { sessionID: SessionID }) {
```
Astro-Han, line 174: The `history` calculation excludes the last message when `messages.at(-1)?.info.id === input.parentID`. This handles re-compaction scenarios where the compaction user message appears at the end of the message list, but the logic is not commented. Consider explaining why this exclusion is necessary for correct tail selection during repeated compactions.

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +112,60 @@ export const layer: Layer.Layer<
...
+      model: Provider.Model
+    }) {
+      const msgs = yield* MessageV2.toModelMessagesEffect(input.messages, input.model, { stripMedia: true })
+      return Token.estimate(JSON.stringify(msgs))
+    })
+
+    const select = Effect.fn("SessionCompaction.select")(function* (input: {
+      messages: MessageV2.WithParts[]
+      cfg: Config.Info
+      model: Provider.Model
```
Astro-Han, line 129: This conditional update only writes `tail_start_id` when it differs from the existing value. While this avoids unnecessary writes, consider adding a brief comment explaining that `tail_start_id` may shift during repeated compactions as older history accumulates and the retained tail boundary moves forward.

packages/opencode/src/session/message-v2.ts
```diff
@@ -926,14 +927,30 @@ export function get(input: { sessionID: SessionID; messageID: MessageID }): With
...
+        tailStartID = compactionPart.tail_start_id
+        continue
+      }
       break
+    }
     if (msg.info.role === "assistant" && msg.info.summary && msg.info.finish && !msg.info.error)
       completed.add(msg.info.parentID)
   }
```
Astro-Han, line 945: This fallback assignment uses `??=` to set `tailStartID` if it wasn't captured during forward iteration. This handles cases where the compaction has `tail_start_id` but no successful summary yet (i.e., `completed` set does not contain the compaction's parent ID). The logic is correct but could benefit from a comment explaining when this fallback path is taken.

Astro-Han:
## Review Summary

This PR introduces significant changes to the compaction system with retained recent turns and interleaved tool routing. Overall the implementation appears sound, but I have several nitpick-level concerns:

### Key Issues

1. **Breaking Change - `Retried` Event Schema**: The `error` field changed from `Schema.String` to `RetryError` object. This may break existing persisted session data. Please confirm that `Retried` events are transient and not stored long-term, or document a migration strategy.

2. **Behavior Change - `prune` Default**: Changing from `true` to `false` affects existing users. Consider documenting this in release notes.

### Missing Documentation

Several complex logic blocks lack comments explaining their rationale:
- The `select` function's fallback behavior when first turn exceeds budget
- The `start === 0` edge case in tail selection
- The `history` calculation for re-compaction scenarios  
- The `filterCompacted` two-phase logic (forward capture + reverse slice)

### Test Concerns

- Token budget boundary conditions not tested
- `JSON.stringify` verification is fragile

The inline comments above detail each concern.

Astro-Han:
@coderabbitai review

Astro-Han:
## Code Review Comments

### 1. `packages/opencode/src/session/compaction.ts:35` — Constants lack documentation

The constants `DEFAULT_TAIL_TURNS`, `MIN_TAIL_TOKENS`, and `MAX_TAIL_TOKENS` define defaults for the retained tail behavior but lack documentation explaining why these specific values were chosen. Consider adding brief comments explaining the rationale—for example, why 2 turns is the default, and how the 2,000–8,000 token range balances typical context windows.

### 2. `packages/opencode/src/v2/session-event.ts:402` — Type mismatch in `Retried.create`

The `Retried.create` factory method signature declares `error: RetryError`, but the schema accepts `RetryEventError` (union of `RetryError` and legacy `string`). This means `create` cannot directly construct events from legacy string errors, even though `Schema.decodeUnknownSync` handles them correctly when deserializing. The `AssistantRetry.fromEvent` helper normalizes both cases, so callers constructing events directly should use that pattern. Consider aligning the `create` signature to match the schema or documenting the limitation.

### 3. `packages/opencode/src/session/compaction.ts:120` — Token estimation approximation

The `estimate` function uses `JSON.stringify(msgs)` to approximate token count. This approach counts JSON syntax characters (quotes, brackets, commas) as tokens, which inflates the estimate compared to actual model tokenization. For compaction budgeting this is acceptable as a rough heuristic, but consider documenting this approximation or using a model-specific tokenizer if precision matters for edge cases near budget boundaries.

### 4. `packages/opencode/test/session/compaction.test.ts:837` — Edge case test value

The test uses `preserve_recent_tokens: 1` which means the tail budget is just 1 token. While valid, this is an extreme edge case. Consider whether a slightly larger value (e.g., 10 or 50) would more realistically test the budget exceeded scenario without depending on JSON serialization quirks that might inflate estimates.

### 5. `packages/opencode/src/session/message-v2.ts:944` — Two-stage fallback logic

The fallback scans `result` (chronological order) for any compaction part with `tail_start_id`. Since the forward iteration already captures the newest compaction's marker via `tailStartID ??= compactionPart?.tail_start_id`, this fallback only runs for truly pending compactions without a completed summary. The comment explains this, but the two-stage logic (forward capture + reverse fallback) could be clearer with a brief note explaining why pending compactions require the reverse scan (they lack completed summaries in the `completed` set).

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +112,63 @@ export const layer: Layer.Layer<
...
```
Astro-Han, line 129: The constants `DEFAULT_TAIL_TURNS`, `MIN_TAIL_TOKENS`, and `MAX_TAIL_TOKENS` define defaults for the retained tail behavior but lack documentation explaining why these specific values were chosen. Consider adding brief comments explaining the rationale—for example, why 2 turns is the default, and how the 2,000–8,000 token range balances typical context windows.

Astro-Han (commented):
Overall the implementation is solid and the changes align well with the #92 session sync requirements. Below are a few minor nitpicks and documentation suggestions.

packages/opencode/src/session/compaction.ts
Astro-Han, line 129: The constants `DEFAULT_TAIL_TURNS`, `MIN_TAIL_TOKENS`, and `MAX_TAIL_TOKENS` define defaults for the retained tail behavior but lack documentation explaining why these specific values were chosen. Consider adding brief comments explaining the rationale.

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +112,63 @@ export const layer: Layer.Layer<
...
+    const estimate = Effect.fn("SessionCompaction.estimate")(function* (input: {
+      messages: MessageV2.WithParts[]
+      model: Provider.Model
+    }) {
+      const msgs = yield* MessageV2.toModelMessagesEffect(input.messages, input.model, { stripMedia: true })
+      return Token.estimate(JSON.stringify(msgs))
+    })
+
+    const select = Effect.fn("SessionCompaction.select")(function* (input: {
+      messages: MessageV2.WithParts[]
```
Astro-Han, line 127: The `estimate` function uses `JSON.stringify(msgs)` to approximate token count. This inflates the estimate compared to actual model tokenization. Consider documenting this approximation.

packages/opencode/src/v2/session-event.ts
```diff
@@ -382,14 +393,16 @@ export namespace SessionEvent {
```
Astro-Han, line 381: The `Retried.create` factory method signature declares `error: RetryError`, but the schema accepts `RetryEventError`. This means `create` cannot directly construct events from legacy string errors. Consider aligning the signature or documenting the limitation.

packages/opencode/src/session/message-v2.ts
```diff
@@ -926,14 +929,31 @@ export function get(input: { sessionID: SessionID; messageID: MessageID }): With
...
+    const compactionPart = msg.parts.find((part): part is CompactionPart => part.type === "compaction")
+    tailStartID ??= compactionPart?.tail_start_id
+    if (tailStartID && msg.info.id === tailStartID) break
+    if (msg.info.role === "user" && completed.has(msg.info.id) && compactionPart) {
+      if (tailStartID) continue
       break
+    }
     if (msg.info.role === "assistant" && msg.info.summary && msg.info.finish && !msg.info.error)
       completed.add(msg.info.parentID)
   }
```
Astro-Han, line 945: The fallback scans chronological order. Consider adding a brief note explaining why pending compactions require this reverse scan (they lack completed summaries in the `completed` set).

Astro-Han (commented):
Additional nitpicks.

packages/opencode/test/session/compaction.test.ts
```diff
@@ -222,7 +234,73 @@ function llm() {
...
+        totalUsage: {
+          inputTokens: 1,
+          outputTokens: 1,
+          totalTokens: 2,
+          inputTokenDetails: {
+            noCacheTokens: undefined,
+            cacheReadTokens: undefined,
+            cacheWriteTokens: undefined,
+          },
+          outputTokenDetails: {
```
Astro-Han, line 282: The test uses `preserve_recent_tokens: 1` which means the tail budget is just 1 token. While valid, this is an extreme edge case. Consider using a larger value (e.g., 50) to more realistically test the budget exceeded scenario.

Astro-Han (commented):
Test nitpicks.

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +114,64 @@ export const layer: Layer.Layer<
...
+      cfg: Config.Info
+      model: Provider.Model
+    }) {
+      const limit = input.cfg.compaction?.tail_turns ?? DEFAULT_TAIL_TURNS
+      if (limit <= 0) return { head: input.messages, tail_start_id: undefined }
+      const budget = tailBudget({ cfg: input.cfg, model: input.model })
```
Astro-Han, line 133: **P1: Semantics change may break existing behavior**

The change from `cfg.compaction?.prune === false` to `cfg.compaction?.prune !== true` alters the default behavior. Previously, pruning was enabled by default and only disabled when explicitly set to `false`. Now, pruning is disabled by default and only enabled when explicitly set to `true`.

This is a significant behavioral shift that could affect users who rely on automatic pruning without having explicitly configured it.

packages/opencode/src/session/compaction.ts
```diff
@@ -32,6 +32,11 @@ export const Event = {
 export const PRUNE_MINIMUM = 20_000
 export const PRUNE_PROTECT = 40_000
 const PRUNE_PROTECTED_TOOLS = ["skill"]
+// Keep the last exchange pair by default, capped to a small bounded slice.
+// The token floor handles small contexts, while the ceiling avoids starving the summary.
+const DEFAULT_TAIL_TURNS = 2
+const MIN_TAIL_TOKENS = 2_000
```
Astro-Han, line 39: **P2: Constants should be documented with their rationale**

`MIN_TAIL_TOKENS = 2_000` and `MAX_TAIL_TOKENS = 8_000` are arbitrary thresholds. Consider adding inline comments explaining why these specific values were chosen (e.g., based on empirical testing, typical turn sizes, or context window ratios).

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +114,64 @@ export const layer: Layer.Layer<
...
+
+      for (let i = turns.length - 1; i >= 0 && kept < limit; i--) {
+        const idx = turns[i]
+        const end = i + 1 < turns.length ? turns[i + 1] : input.messages.length
+        const size = yield* estimate({
+          messages: input.messages.slice(idx, end),
+          model: input.model,
+        })
+        if (kept === 0 && size > budget) {
+          // The latest turn must stay intact or be summarized with the rest.
```
Astro-Han, line 153: **P2: Overestimation rationale should be explicit**

The comment "JSON adds structural overhead, so this intentionally overestimates" is good, but could specify the expected overestimation factor (e.g., "typically 10-20% overhead") to help reviewers assess whether tuning is needed.

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +114,64 @@ export const layer: Layer.Layer<
...
+          // Keeping a partial turn would separate tool calls from their prompt.
+          log.info("tail fallback", { budget, size })
+          return { head: input.messages, tail_start_id: undefined }
+        }
+        if (total + size > budget) break
+        total += size
+        start = idx
```
Astro-Han, line 160: **P1: Edge case: limit <= 0 silently disables tail retention**

When `tail_turns` is negative or zero, the early return `{ head: input.messages, tail_start_id: undefined }` is correct but silent. Consider adding `log.info` to inform operators. Additionally, negative values should be rejected at schema level rather than silently treated as disabled.

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +114,64 @@ export const layer: Layer.Layer<
...
+        tail_start_id: input.messages[start]?.info.id,
+      }
+    })
+
     // goes backwards through parts until there are PRUNE_PROTECT tokens worth of tool
     // calls, then erases output of older tool calls to free context space
     const prune = Effect.fn("SessionCompaction.prune")(function* (input: { sessionID: SessionID }) {
       const cfg = yield* config.get()
-      if (cfg.compaction?.prune === false) return
+      if (cfg.compaction?.prune !== true) return
```
Astro-Han, line 176: **P2: Budget overflow check may leave insufficient tail**

The loop `if (total + size > budget) break` stops when adding another turn would exceed budget. If `total` is very small, the tail might be too short to be useful. Consider whether MIN_TAIL_TOKENS adequately protects this, or whether fallback to full summarization is sufficient.

packages/opencode/src/session/compaction.ts
```diff
@@ -187,6 +260,15 @@ export const layer: Layer.Layer<
       const model = agent.model
         ? yield* provider.getModel(agent.model.providerID, agent.model.modelID)
         : yield* provider.getModel(userMessage.model.providerID, userMessage.model.modelID)
+      const cfg = yield* config.get()
+      // Re-compaction passes the compaction request as parentID. Exclude it so the retained tail
+      // is selected from real conversation history instead of the marker message.
+      const history = compactionPart && messages.at(-1)?.info.id === input.parentID ? messages.slice(0, -1) : messages
+      const selected = yield* select({
```
Astro-Han, line 269: **P1: Compaction part detection assumes position**

The condition `messages.at(-1)?.info.id === input.parentID` assumes compaction part is always at end. If race conditions or concurrent updates could move it, this logic would fail. Consider whether architecture guarantees this, or whether a more robust check (finding by parentID) would be safer.

packages/opencode/src/session/compaction.ts
```diff
@@ -287,6 +369,14 @@ When constructing the summary, try to stick to this template:
         return "stop"
       }
 
+      if (compactionPart && compactionPart.tail_start_id !== selected.tail_start_id) {
+        // Repeated compactions can move the retained-tail boundary forward as older turns accumulate.
+        yield* session.updatePart({
+          ...compactionPart,
+          tail_start_id: selected.tail_start_id,
```
Astro-Han, line 378: **P2: Re-compaction boundary update may race**

The check and subsequent `updatePart` could race with concurrent compaction modifying `tail_start_id`. Consider whether session store provides atomicity guarantees, or whether compare-and-swap is needed.

packages/opencode/src/session/message-v2.ts
```diff
@@ -210,6 +210,9 @@ export const CompactionPart = PartBase.extend({
   type: z.literal("compaction"),
   auto: z.boolean(),
   overflow: z.boolean().optional(),
```
Astro-Han, line 212: **P2: Schema description style inconsistency**

`tail_start_id` uses `.describe()` while other fields use `.annotate({ description: ... })`. Consider standardizing style across the schema.

packages/opencode/src/session/message-v2.ts
```diff
@@ -926,14 +929,32 @@ export function get(input: { sessionID: SessionID; messageID: MessageID }): With
 export function filterCompacted(msgs: Iterable<WithParts>) {
   const result = [] as WithParts[]
   const completed = new Set<string>()
+  let tailStartID: MessageID | undefined
+  // Read newest to oldest. The first tail marker wins, even before its summary finishes.
   for (const msg of msgs) {
     result.push(msg)
-    if (msg.info.role === "user" && completed.has(msg.info.id) && msg.parts.some((part) => part.type === "compaction"))
+    const compactionPart = msg.parts.find((part): part is CompactionPart => part.type === "compaction")
```
Astro-Han, line 936: **P1: Complex iteration logic needs review**

The `tailStartID ??=` inside loop and `if (tailStartID && msg.info.id === tailStartID) break` interaction is subtle. Edge cases could incorrectly include/exclude messages. Consider explicit tests for: multiple compactions with overlapping tail_start_id, pending vs completed compactions, tool parts before tail_start_id.

packages/opencode/src/session/message-v2.ts
```diff
@@ -926,14 +929,32 @@ export function get(input: { sessionID: SessionID; messageID: MessageID }): With
...
       break
+    }
     if (msg.info.role === "assistant" && msg.info.summary && msg.info.finish && !msg.info.error)
       completed.add(msg.info.parentID)
   }
   result.reverse()
+  // Pending compactions have no completed summary, so use the newest marker found above.
+  // If none was captured during the newest-first scan, fall back to the chronological result.
+  tailStartID ??= result
+    .flatMap((msg) => msg.parts)
```
Astro-Han: **P3: Fallback logic duplication**

The fallback re-scans after reversing, duplicating the newest-first scan. The comment explains why, but consider consolidating to avoid the second pass.

packages/opencode/src/session/session.ts
```diff
@@ -258,7 +258,7 @@ export const Event = {
     "session.error",
     z.object({
       sessionID: SessionID.zod.optional(),
-      error: MessageV2.Assistant.shape.error,
+      error: z.lazy(() => MessageV2.Assistant.shape.error),
```
Astro-Han, line 262: **P1: z.lazy needs explanatory comment**

`z.lazy(() => MessageV2.Assistant.shape.error)` resolves circular reference but lacks explanation. Future maintainers could accidentally remove it. Add: `// z.lazy required due to circular reference with MessageV2.Assistant`

packages/opencode/src/v2/session-entry.ts
```diff
@@ -146,8 +169,12 @@ export namespace SessionEntry {
     )
```
Astro-Han, line 170: **P1: callID assumption needs verification**

`lastTool(assistant, event.callID)` assumes callID always exists for tool events. Verify schema guarantees. If undefined in legacy/malformed data, fallback `callID === undefined || x.callID === callID` would match last tool incorrectly.

packages/opencode/src/v2/session-entry.ts
```diff
@@ -96,13 +96,36 @@ export namespace SessionEntry {
     text: Schema.String,
   }) {}
 
+  export class AssistantRetry extends Schema.Class<AssistantRetry>("Session.Entry.Assistant.Retry")({
+    attempt: Schema.Number,
+    error: SessionEvent.RetryError,
+    time: Schema.Struct({
```
Astro-Han, line 102: **P2: RetryError normalization assumes non-retryable**

`isRetryable: false` for legacy string errors may be incorrect ("rate limited" is retryable). Consider parsing content or documenting this limitation.

packages/opencode/src/v2/session-event.ts
```diff
@@ -49,6 +49,17 @@ export namespace SessionEvent {
     source: Source.pipe(Schema.optional),
   }) {}
 
+  export class RetryError extends Schema.Class<RetryError>("Session.Event.Retry.Error")({
+    message: Schema.String,
+    statusCode: Schema.Number.pipe(Schema.optional),
+    isRetryable: Schema.Boolean,
+    responseHeaders: Schema.Record(Schema.String, Schema.String).pipe(Schema.optional),
+    responseBody: Schema.String.pipe(Schema.optional),
```
Astro-Han, line 57: **P3: Optional fields need documentation**

`statusCode`, `responseHeaders`, `responseBody`, `metadata` lack usage documentation. Add brief descriptions for when these are populated.

packages/opencode/src/config/config.ts
```diff
@@ -255,7 +255,7 @@ const InfoSchema = Schema.Struct({
...
       }),
       prune: Schema.optional(Schema.Boolean).annotate({
-        description: "Enable pruning of old tool outputs (default: true)",
+        description: "Set true to prune old tool outputs. False or omitted disables automatic pruning.",
```
Astro-Han: **P1: Description change needs migration note**

Users won't know behavior changed from default-enabled to opt-in. Consider changelog/migration note indicating `prune: true` is now required.

packages/opencode/test/session/messages-pagination.test.ts
```diff
@@ -711,6 +715,132 @@ describe("MessageV2.filterCompacted", () => {
     })
   })
 
+  it.live(
+    "keeps retained tail messages after a compaction part with tail_start_id",
+    provideTmpdirInstance(() =>
+      Effect.promise(async () => {
```
Astro-Han: **P2: .live annotation may be unnecessary**

Tests don't appear to require live Effect runtime. Consider regular `test` for reduced overhead.

packages/opencode/test/session/compaction.test.ts
```diff
@@ -580,12 +658,56 @@ describe("session.compaction.prune", () => {
         expect(part?.type).toBe("tool")
         expect(part?.state.status).toBe("completed")
         if (part?.type === "tool" && part.state.status === "completed") {
-          expect(part.state.time.compacted).toBeNumber()
+          expect(part.state.time.compacted).toBeUndefined()
         }
```
Astro-Han, line 680: **P2: Test title style inconsistent**

"compacts old completed tool output when prune is explicitly enabled" uses sentence style. Others use hyphenated titles. Standardize for grepability.

Astro-Han (commented):
## Code Review Summary

This PR introduces significant changes to session compaction behavior with a new tail retention mechanism and retry event tracking. Overall the implementation is solid, but there are several issues requiring attention.

### P1 Issues (Suggest Fixing)

1. **Prune semantics change** (compaction.ts:133) - The shift from `prune === false` to `prune !== true` silently changes default behavior for existing users without migration guidance.

2. **Silent tail_turns edge case** (compaction.ts:158) - Negative/zero values silently disable tail retention without logging.

3. **Compaction part position assumption** (compaction.ts:267) - Logic assumes compaction part is always at array end, may fail under race conditions.

4. **filterCompacted iteration complexity** (message-v2.ts:936) - The interaction between tail markers and compaction logic is subtle; needs explicit edge case tests.

5. **z.lazy missing explanation** (session.ts:261) - Circular reference fix lacks comment explaining necessity.

6. **callID assumption unverified** (session-entry.ts:169) - Tool routing assumes callID always exists; needs schema verification.

7. **Config description lacks migration note** (config.ts:258) - Users won't know behavior changed from default-enabled to opt-in.

### P2 Issues (Record Only)

- Magic constants need rationale documentation
- Budget overflow may leave insufficient tail
- Re-compaction boundary update potential race
- Schema description style inconsistency
- RetryError normalization assumption questionable
- Test helper verbose type checks
- Test .live annotation possibly unnecessary
- Test title style inconsistency

### P3 Issues (Optional)

- Fallback logic duplication
- RetryError optional fields lack documentation

### Recommendation

Address P1 issues before merge, particularly:
1. Add migration note/changelog for prune behavior change
2. Add logging for edge case configurations
3. Verify callID schema guarantees or add fallback handling
4. Add comment explaining z.lazy necessity

packages/opencode/src/session/compaction.ts
Astro-Han: Keeping this behavior as intentional for #92. The PR moves pruning to explicit opt-in so retained-tail compaction does not silently discard tool output. The migration path is already in the PR Compatibility Notes and config/SDK docs: set `compaction.prune: true` to keep the old pruning behavior.

packages/opencode/src/session/compaction.ts
Astro-Han: Fixed. The constants now explain the default retained exchange pair and why the token floor/ceiling bounds exist.

packages/opencode/src/session/compaction.ts
Astro-Han: Adjusted the comment to explain the conservative overestimate. I did not add a fixed percentage because the overhead varies by message shape; claiming a 10 to 20 percent factor would be less accurate than the current bounded behavior and tests.

packages/opencode/src/session/compaction.ts
Astro-Han: Not changing this. Negative `tail_turns` is already rejected by the config schema through `NonNegativeInt`; `0` is the explicit disable value. Adding runtime logs for an intentional disable would make normal config behavior noisy.

packages/opencode/src/session/compaction.ts
Astro-Han: Not changing this. `MIN_TAIL_TOKENS` protects the default budget, and explicit smaller `preserve_recent_tokens` values are user overrides. If the latest turn itself exceeds budget, the code already falls back to summarizing the full history instead of keeping a partial turn.

packages/opencode/src/session/compaction.ts
Astro-Han: Fixed. Re-compaction history now excludes the parent compaction message by id instead of assuming it is the final array element.

packages/opencode/src/session/compaction.ts
Astro-Han: Not changing this in this PR. The update is idempotent and there is no compare-and-swap API on session parts here. Adding a new concurrency primitive would be a larger storage-layer change than this retained-tail fix.

packages/opencode/src/session/message-v2.ts
Astro-Han: Not changing this. `tail_start_id` is a Zod schema field in `message-v2.ts`, where `.describe()` is the local style. `.annotate({ description })` is used on Effect Schema config fields, so switching here would mix schema APIs rather than standardize them.

packages/opencode/src/session/message-v2.ts
Astro-Han: No code change in this round. Existing tests cover the important cases here: pending versus completed compactions, newest pending marker winning over older completed markers, and retained assistant follow-ups. Tool parts are attached to messages, so slicing by message id preserves them with the retained turn.

packages/opencode/src/session/message-v2.ts
Astro-Han: Not changing this. The second scan only runs if the newest-first pass did not capture a marker, and keeping it as a fallback makes the pending-compaction path explicit without changing behavior.

packages/opencode/src/session/session.ts
Astro-Han: Fixed. Added a comment explaining that `z.lazy` avoids eager access to `MessageV2.Assistant` while `session.ts` is still initializing.

packages/opencode/src/v2/session-entry.ts
Astro-Han: Fixed. Verified all tool event schemas require `callID`, then tightened `lastTool` to require a string call id and removed the undefined fallback that could match the wrong tool.

packages/opencode/src/v2/session-entry.ts
Astro-Han: Fixed with documentation. Legacy retry string events do not persist retryability metadata, so the normalized fallback keeps the conservative default and now says why.

packages/opencode/src/v2/session-event.ts
Astro-Han: Not changing this in this PR. These fields mirror provider retry error payload details and are not part of the retained-tail or pruning behavior being changed here.

packages/opencode/src/config/config.ts
Astro-Han: Already covered. The PR body has a Compatibility Notes section saying existing configs should set `compaction.prune: true` to keep automatic pruning, and the config/SDK descriptions now state that false or omitted disables pruning.

packages/opencode/test/session/messages-pagination.test.ts
Astro-Han: Keeping `it.live`. These tests exercise Effect-backed session fixtures via `provideTmpdirInstance`, and this was changed to the Effect harness after earlier review feedback. Switching back to raw `test` would reintroduce ad hoc runtime plumbing.

packages/opencode/test/session/compaction.test.ts
Astro-Han: Not changing this. The surrounding session tests use descriptive sentence titles broadly, and this title is grepable for both `prune` and `explicitly enabled`.

packages/opencode/src/session/compaction.ts
```diff
@@ -67,6 +73,20 @@ export interface Interface {
 
 export class Service extends Context.Service<Service, Interface>()("@opencode/SessionCompaction") {}
 
+function usable(input: { cfg: Config.Info; model: Provider.Model }) {
+  const reserved = input.cfg.compaction?.reserved ?? Math.min(20_000, ProviderTransform.maxOutputTokens(input.model))
+  return input.model.limit.input
+    ? Math.max(0, input.model.limit.input - reserved)
+    : Math.max(0, input.model.limit.context - ProviderTransform.maxOutputTokens(input.model))
```
Astro-Han: **P2 — `reserved` override silently dropped on the fallback branch.** When `model.limit.input` is falsy we recompute `context - maxOutputTokens(model)` and ignore `cfg.compaction?.reserved` entirely. A user who sets `compaction.reserved` to tune overflow headroom will see it respected for providers that expose `limit.input` and silently no-op for providers that only expose `limit.context`. Hoist the `reserved` binding so both branches subtract it:

```ts
const reserved = input.cfg.compaction?.reserved ?? Math.min(20_000, ProviderTransform.maxOutputTokens(input.model))
const base = input.model.limit.input ?? input.model.limit.context
return Math.max(0, base - reserved)
```

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +115,64 @@ export const layer: Layer.Layer<
...
+        tail_start_id: input.messages[start]?.info.id,
+      }
+    })
+
     // goes backwards through parts until there are PRUNE_PROTECT tokens worth of tool
     // calls, then erases output of older tool calls to free context space
     const prune = Effect.fn("SessionCompaction.prune")(function* (input: { sessionID: SessionID }) {
       const cfg = yield* config.get()
-      if (cfg.compaction?.prune === false) return
+      if (cfg.compaction?.prune !== true) return
```
Astro-Han, line 176: **P2 — default semantics flipped from opt-out to opt-in.** Pre-PR: `prune === false` disabled; everything else enabled. Post-PR: only `prune === true` enables. Users on `compaction: {}` (or any config that omits `prune`) silently lose automatic pruning on the next upgrade. The PR body flags this, but it would be good to (a) land a release note entry and (b) mirror the convention `auto` uses — `Schema.optional(Schema.Boolean).annotate({ description: "... (default: false)" })` — so the schema stays self-documenting.

packages/opencode/src/config/config.ts
Astro-Han: **P3 — copy nit.** The new description reads "Set true to prune old tool outputs. False or omitted disables automatic pruning." Two issues:
1. Neighboring `auto` uses the `(default: true)` suffix style; this one buries the default in prose. Prefer `Prune old tool outputs (default: false). Set true to enable.`
2. Use backticks around the literal values (`` `true` ``, `` `false` ``) so the generated OpenAPI/SDK docs render them as code.

Astro-Han (commented):
Nitpick review. Logic works, but a few places look worth tightening. P2 items could bite on upgrade; P3 items are clarity/dead-code.

packages/opencode/src/v2/session-event.ts
```diff
@@ -382,14 +393,16 @@ export namespace SessionEvent {
   export class Retried extends Schema.Class<Retried>("Session.Event.Retried")({
     ...Base,
     type: Schema.Literal("retried"),
-    error: Schema.String,
+    attempt: Schema.Number,
```
Astro-Han: **P2 — required `attempt` is a decode-side break for historical events.** `attempt: Schema.Number` means any previously-emitted `retried` event without `attempt` fails `Schema.decodeUnknownSync(Retried)`. You already handle legacy *error* shape via `RetryEventError = Union([RetryError, Schema.String])` for forward compatibility; `attempt` needs the same treatment if `Retried` ever decodes persisted rows or replayed streams. If events are always transient/ephemeral in the current runtime this is moot — worth either a one-line comment confirming that invariant, or `Schema.optional(...)` with a `?? 0` in `fromEvent`.

packages/opencode/src/v2/session-entry.ts
```diff
@@ -96,13 +96,37 @@ export namespace SessionEntry {
...
+      created: Schema.DateTimeUtc,
+    }),
+  }) {
+    static fromEvent(event: SessionEvent.Retried) {
+      const error =
+        typeof event.error === "string"
+          // Legacy retry events stored only the message, so retryability cannot be recovered.
+          ? new SessionEvent.RetryError({ message: event.error, isRetryable: false })
```
Astro-Han: **P3 — legacy-string retry default flips historical intent.** The event is named `retried`: the mere fact it was emitted means the retry logic decided the error was retryable at that moment. Defaulting the recovered `isRetryable` to `false` tells downstream UI "this was a non-retryable error that we retried anyway," which is contradictory. `true` is the safer reconstruction and matches the event's emission semantics; the comment can stay to explain that no finer-grained data is recoverable.

packages/opencode/src/session/message-v2.ts
```diff
@@ -926,14 +929,32 @@ export function get(input: { sessionID: SessionID; messageID: MessageID }): With
...
+    .find((part): part is CompactionPart => part.type === "compaction" && part.tail_start_id !== undefined)
+    ?.tail_start_id
```
Astro-Han: **P3 — this fallback is unreachable.** Inside the loop at line 937, `tailStartID ??= compactionPart?.tail_start_id` runs for every msg that gets pushed into `result`. So any compaction part with a `tail_start_id` present in the iterated stream is already captured. `result` is a strict subset of what the loop iterated over, so scanning `result` again after the loop can never find a `tail_start_id` the loop missed.

If the intent was to cover a case where iteration terminates early (e.g., break on completed-compaction-without-tail), that also can't set `tailStartID` because such a marker by definition has no `tail_start_id`. Either drop the block, or update the comment to describe the actual case you're guarding against.

packages/opencode/src/session/compaction.ts
```diff
@@ -187,6 +261,15 @@ export const layer: Layer.Layer<
       const model = agent.model
         ? yield* provider.getModel(agent.model.providerID, agent.model.modelID)
         : yield* provider.getModel(userMessage.model.providerID, userMessage.model.modelID)
+      const cfg = yield* config.get()
+      // Re-compaction passes the compaction request as parentID. Exclude it so the retained tail
```
Astro-Han: **P3 — comment is misleading.** "Re-compaction passes the compaction request as parentID" — every `process()` call does, not only re-compaction (grep for callers: initial compaction flows through the same path with `compactionPart` set on `parent`). Also, `select()` already filters compaction-bearing user messages out of `turns` at line 135-137, so excluding the parent from `history` is primarily about keeping the sliced token estimate clean, not about changing which turns are eligible. Suggest:

```ts
// Drop the new compaction marker itself; otherwise it shows up inside the last turn's
// slice below and inflates the token estimate.
```

packages/opencode/src/session/compaction.ts
```diff
@@ -95,11 +115,64 @@ export const layer: Layer.Layer<
...
+      const budget = tailBudget({ cfg: input.cfg, model: input.model })
+      const turns = input.messages.flatMap((msg, idx) =>
+        msg.info.role === "user" && !msg.parts.some((part) => part.type === "compaction") ? [idx] : [],
+      )
+      if (!turns.length) return { head: input.messages, tail_start_id: undefined }
+
+      let total = 0
+      let start = input.messages.length
+      let kept = 0
+
```
Astro-Han, line 143: **P3 — per-turn slice can straddle an older compaction boundary.** `const end = i + 1 < turns.length ? turns[i + 1] : input.messages.length` walks from one user turn to the next *eligible* user turn. If there's an intermediate compaction marker + summary assistant in between (previous compaction still lives in history), both get lumped into the earlier turn's slice for estimation. That biases `estimate()` upward for turns that sit just before a historical compaction boundary, making the latest-turn fallback more likely to fire. Not a correctness bug today but worth a comment, or switch `end` to `turns[i + 1] ?? input.messages.length` after first skipping over compaction-bearing indices.

packages/opencode/test/session/session.test.ts
```diff
@@ -12,6 +12,10 @@ const projectRoot = path.join(__dirname, "../..")
 void Log.init({ print: false })
 
 describe("PawWork runtime namespace", () => {
+  test("session error event schema can be imported without eager assistant shape access", () => {
+    expect(SessionNs.Event.Error.properties.shape.error).toBeDefined()
+  })
```
Astro-Han, line 21: **P3 — test is thinner than the regression it guards.** `expect(...shape.error).toBeDefined()` proves `z.lazy(...)` didn't crash at import. It doesn't prove the resulting schema still decodes a real error payload, which is the actual surface if `MessageV2.Assistant` shape drifts later. Add one `Event.Error.safeParse({ sessionID: ..., error: { name: ..., message: ... } })` call to exercise the lazy resolution end-to-end.

packages/opencode/src/session/compaction.ts
Astro-Han: Fixed. `compaction.reserved` now subtracts from `model.limit.input ?? model.limit.context` in retained-tail budgeting, and I aligned the actual overflow helper in `session/overflow.ts` so context-only providers honor the same override. Added a regression test for the context-only path.

packages/opencode/src/session/compaction.ts
Astro-Han: Partially fixed. The schema, OpenAPI, and generated SDK text now use explicit default copy: `Prune old tool outputs (default: `false`). Set `true` to enable.` I did not add a separate release-note file because this repo does not appear to have a source release-note or changeset mechanism for package changes; the PR Compatibility Notes already carry the migration note.

packages/opencode/src/config/config.ts
Astro-Han: Fixed. Updated config, OpenAPI, and generated SDK copy to the default-suffix style with literal `true` and `false` rendered as code.

packages/opencode/src/v2/session-event.ts
Astro-Han: Fixed. `Retried.attempt` now decodes as optional for legacy events, and `SessionEntry.AssistantRetry.fromEvent` normalizes missing attempts to `0`. Added a decode/reducer regression test for historical retry events without `attempt`.

packages/opencode/src/v2/session-entry.ts
Astro-Han: Fixed. Legacy string retry errors now reconstruct `isRetryable: true`, matching the fact that a `retried` event was emitted. The comment remains limited to the missing fine-grained metadata.

packages/opencode/src/session/message-v2.ts
Astro-Han: Fixed. Removed the unreachable fallback scan; the newest-first loop already sees every message added to `result`, so any `tail_start_id` in that subset is captured there.

packages/opencode/src/session/compaction.ts
Astro-Han: Fixed. Replaced the misleading re-compaction wording with a narrower comment explaining that the compaction marker is dropped to keep retained-tail token estimation clean.

packages/opencode/src/session/compaction.ts
Astro-Han: Fixed. The slice end now uses the next user message boundary, not just the next eligible retained-turn user, so historical compaction marker turns do not get folded into the previous eligible turn estimate.

packages/opencode/test/session/session.test.ts
Astro-Han: Fixed. The regression test now runs `SessionNs.Event.Error.properties.safeParse(...)` with a real `ContextOverflowError` payload, so the lazy schema resolution is exercised end to end.

packages/opencode/src/session/overflow.ts
```diff
@@ -15,8 +15,6 @@ export function isOverflow(input: { cfg: Config.Info; tokens: MessageV2.Assistan
 
   const reserved =
     input.cfg.compaction?.reserved ?? Math.min(COMPACTION_BUFFER, ProviderTransform.maxOutputTokens(input.model))
-  const usable = input.model.limit.input
-    ? input.model.limit.input - reserved
-    : context - ProviderTransform.maxOutputTokens(input.model)
+  const usable = (input.model.limit.input ?? context) - reserved
   return count >= usable
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `overflow.ts` now clamps `usable` with `Math.max(0, ...)`, matching the defensive behavior in retained-tail budgeting. I did not add a black-box test because nonnegative token counts make the negative-to-zero clamp externally equivalent for the current API surface.

packages/opencode/src/v2/session-event.ts
```diff
@@ -382,14 +393,16 @@ export namespace SessionEvent {
...
   }) {
-    static create(input: BaseInput & { error: string }) {
+    static create(input: BaseInput & { attempt: number; error: RetryError | string }) {
       return new Retried({
         id: input.id ?? ID.create(),
         type: "retried",
         timestamp: input.timestamp ?? DateTime.makeUnsafe(Date.now()),
         metadata: input.metadata,
+        attempt: input.attempt,
         error: input.error,
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `Retried.create` now accepts omitted `attempt` and only writes the field when present, matching the optional decode schema. The existing legacy no-attempt reducer test now constructs that shape through the factory.

packages/opencode/test/session/compaction.test.ts
```diff
@@ -580,12 +676,56 @@ describe("session.compaction.prune", () => {
...
+        }
+
+        const messages = await svc.messages({ sessionID: session.id })
+        const updated = messages.flatMap((message) => message.parts).find((candidate) => candidate.id === part.id)
+        expect(updated?.type).toBe("tool")
+        if (updated?.type !== "tool" || updated.state.status !== "completed") return
+        expect(updated.state.time.compacted).toBeNumber()
+      },
+    })
+  })
```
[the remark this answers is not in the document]
Astro-Han: Partially fixed. The prune-enabled test is now on `it.live(...)` with `provideTmpdirInstance(...)`, and it uses the harness-provided `SessionCompaction.Service` plus tmpdir config instead of a manual runtime. I kept the two retained-tail process tests on their explicit runtime because they need per-test LLM/provider/config replacement; the fixed `env` for this file uses the fake processor layer and does not exercise that path.

packages/opencode/test/session/messages-pagination.test.ts
```diff
@@ -711,6 +715,132 @@ describe("MessageV2.filterCompacted", () => {
...
+        expect(ids).toContain(newestTailAssistant)
+        expect(ids).not.toContain(previousTail)
+        expect(ids).not.toContain(previousTailAssistant)
+        expect(ids).not.toContain(old)
+        expect(ids).not.toContain(oldAssistant)
+
+        await svc.remove(session.id)
+      }),
+    ),
+  )
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The three new `it.live` filterCompacted tests now use `Effect.gen`, yield `SessionNs.Service` from the harness, and use effect helpers for user, assistant, and compaction part setup instead of calling the external `svc` runtime from `Effect.promise`.

Astro-Han:
@coderabbitai review

PR #146, opened 2026-04-22
ci: split Windows opencode unit tests

Astro-Han:
## Summary
Split the advisory Windows opencode unit signal into three stable directory shards: `opencode-session`, `opencode-config-project`, and `opencode-server-tools`.

## Why
The current `unit-windows-opencode` job is the long pole for advisory Windows CI. Recent successful runs showed the opencode Windows unit step commonly around 9 to 10 minutes and the whole check around 13 to 14 minutes. Sharding only the opencode Windows signal should reduce PR wall time while keeping Windows checks advisory.

## Related Issue
No issue. This is a small CI wall-time optimization with a narrow workflow-only scope.

## How To Verify
TDD RED: after adding the coverage contract, temporarily omitted `test/auth` from the server-tools shard and confirmed the test failed with missing `test/auth/auth.test.ts`.

```bash
bun --cwd packages/opencode test test/github/ci-workflow.test.ts -t "covers each opencode test file exactly once"
bun --cwd packages/opencode test test/github
git diff --check
```

Fresh-eyes review: first review found the missing coverage contract, fixed; second review returned no findings.

## Screenshots or Recordings
Not applicable, no visible UI changes.

## Checklist

## Release Notes

* **Chores**
  * Optimized Windows continuous integration infrastructure with improved test execution efficiency through better organization and parallel processing capabilities.
  * Enhanced test workflow configuration with refined resource allocation and improved reporting to provide clearer visibility into test suite performance.
  * Improved test infrastructure to support faster execution and more granular tracking across different test categories.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -347,9 +347,9 @@ jobs:
     needs: changes
     if: needs.changes.outputs.docs_only != 'true'
     runs-on: windows-latest
-    # Windows unit jobs are advisory and package-scoped. The 20-minute budget
-    # gives each matrix child room for Windows runner setup and install overhead
-    # while still surfacing the package that timed out.
+    # Windows unit jobs are advisory, package-scoped or shard-scoped. The
+    # 20-minute budget applies per matrix child; opencode has three parallel
+    # shards, so a full stall can consume up to three Windows runner slots.
     timeout-minutes: 20
     continue-on-error: true
     strategy:
@@ -357,12 +357,88 @@ jobs:
       matrix:
         include:
           - package: app
+            uses_turbo: true
             command: bun turbo test:ci --filter=@opencode-ai/app
             report_path: packages/app/.artifacts/unit/junit.xml
-          - package: opencode
-            command: bun turbo test:ci --filter=opencode
-            report_path: packages/opencode/.artifacts/unit/junit.xml
+          # The opencode-* package values below are synthetic shard names for
+          # readable advisory checks and matrix keys, not package.json names.
+          # Direct bun test keeps opencode's test:ci timeout while allowing
+          # shard-local paths and reporter output. Shards are balanced by
+          # observed runtime, not directory count.
+          # Session also carries plugin, permission, util, skill, and root
+          # tests to keep the Windows shard runtimes close.
+          - package: opencode-session
+            uses_turbo: false
+            command: >-
+              cd packages/opencode && bun test
+              --timeout 30000
+              --reporter=junit
+              --reporter-outfile=.artifacts/unit/junit-windows-session.xml
+              test/session
+              test/plugin
+              test/permission
+              test/util
+              test/skill
+              test/index-runtime-namespace.test.ts
+              test/npm.test.ts
+              test/permission-task.test.ts
+            report_path: packages/opencode/.artifacts/unit/junit-windows-session.xml
+          # Config and project cover configuration, project discovery, file,
+          # and GitHub workflow tests.
+          - package: opencode-config-project
+            uses_turbo: false
+            command: >-
+              cd packages/opencode && bun test
+              --timeout 30000
+              --reporter=junit
+              --reporter-outfile=.artifacts/unit/junit-windows-config-project.xml
+              test/config
+              test/project
+              test/file
+              test/github
+            report_path: packages/opencode/.artifacts/unit/junit-windows-config-project.xml
+          # Server tools carries many smaller and faster directories, which is
+          # why it has more entries than the other two opencode shards.
+          - package: opencode-server-tools
+            uses_turbo: false
+            command: >-
+              cd packages/opencode && bun test
+              --timeout 30000
+              --reporter=junit
+              --reporter-outfile=.artifacts/unit/junit-windows-server-tools.xml
+              test/server
+              test/snapshot
+              test/tool
+              test/mcp
+              test/question
+              test/effect
+              test/agent
+              test/git
+              test/storage
+              test/provider
+              test/pty
+              test/share
+              test/script
+              test/memory
+              test/lsp
+              test/fixture
+              test/acp
+              test/bus
+              test/cli
+              test/global
+              test/format
+              test/account
+              test/sync
+              test/filesystem
+              test/patch
+              test/shell
+              test/control-plane
+              test/ide
+              test/installation
+              test/auth
+            report_path: packages/opencode/.artifacts/unit/junit-windows-server-tools.xml
           - package: desktop
+            uses_turbo: true
             command: bun turbo test:ci --filter=@opencode-ai/desktop-electron
             report_path: packages/desktop-electron/.artifacts/unit/junit.xml
     permissions:
@@ -396,16 +472,20 @@ jobs:
             bun-${{ runner.os }}-
 
       - uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # actions/cache@v5
+        if: matrix.uses_turbo
         with:
           path: .turbo/cache
-          # Keep Turbo caches package-scoped because each matrix child runs a
-          # different test filter; restore keys reuse only that package's cache.
+          # Only app and desktop invoke Turbo in this Windows matrix.
           key: turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-${{ hashFiles('turbo.json', '**/package.json', 'bun.lock') }}-${{ github.sha }}
           restore-keys: |
             turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-${{ hashFiles('turbo.json', '**/package.json', 'bun.lock') }}-
             turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-
 
       - run: bun install --frozen-lockfile
+
+      - name: Prepare unit artifact directory
+        if: matrix.uses_turbo == false
+        run: mkdir -p "$(dirname "${{ matrix.report_path }}")"
 
       - name: unit
         id: unit
--- a/packages/opencode/test/github/ci-workflow.test.ts
+++ b/packages/opencode/test/github/ci-workflow.test.ts
@@ -19,7 +19,7 @@ const windowsUnitJobName = "unit-windows"
 
 // Suffixes drive readable job and artifact names; commands use package.json names verbatim.
 // `opencode` is intentionally unscoped because that is its actual package name.
-const unitPackages = [
+const linuxUnitPackages = [
   {
     suffix: "app",
     command: "bun turbo test:ci --filter=@opencode-ai/app",
@@ -37,14 +37,57 @@ const unitPackages = [
   },
 ] as const
 
-const linuxUnitJobs = unitPackages.map((pkg) => ({
+const windowsOpencodeShards = [
+  // Intentional dual source with ci.yml: this pins the exact shard command
+  // contract while the coverage test expands these paths to catch workflow
+  // drift and missing opencode tests. Update ci.yml and this list together.
+  {
+    suffix: "opencode-session",
+    usesTurbo: false,
+    command:
+      "cd packages/opencode && bun test --timeout 30000 --reporter=junit --reporter-outfile=.artifacts/unit/junit-windows-session.xml test/session test/plugin test/permission test/util test/skill test/index-runtime-namespace.test.ts test/npm.test.ts test/permission-task.test.ts",
+    reportPath: "packages/opencode/.artifacts/unit/junit-windows-session.xml",
+  },
+  {
+    suffix: "opencode-config-project",
+    usesTurbo: false,
+    command:
+      "cd packages/opencode && bun test --timeout 30000 --reporter=junit --reporter-outfile=.artifacts/unit/junit-windows-config-project.xml test/config test/project test/file test/github",
+    reportPath: "packages/opencode/.artifacts/unit/junit-windows-config-project.xml",
+  },
+  {
+    suffix: "opencode-server-tools",
+    usesTurbo: false,
+    command:
+      "cd packages/opencode && bun test --timeout 30000 --reporter=junit --reporter-outfile=.artifacts/unit/junit-windows-server-tools.xml test/server test/snapshot test/tool test/mcp test/question test/effect test/agent test/git test/storage test/provider test/pty test/share test/script test/memory test/lsp test/fixture test/acp test/bus test/cli test/global test/format test/account test/sync test/filesystem test/patch test/shell test/control-plane test/ide test/installation test/auth",
+    reportPath: "packages/opencode/.artifacts/unit/junit-windows-server-tools.xml",
+  },
+] as const
+
+const windowsUnitPackages = [
+  {
+    suffix: "app",
+    usesTurbo: true,
+    command: "bun turbo test:ci --filter=@opencode-ai/app",
+    reportPath: "packages/app/.artifacts/unit/junit.xml",
+  },
+  ...windowsOpencodeShards,
+  {
+    suffix: "desktop",
+    usesTurbo: true,
+    command: "bun turbo test:ci --filter=@opencode-ai/desktop-electron",
+    reportPath: "packages/desktop-electron/.artifacts/unit/junit.xml",
+  },
+] as const
+
+const linuxUnitJobs = linuxUnitPackages.map((pkg) => ({
   ...pkg,
   jobName: `unit-${pkg.suffix}`,
   checkName: `unit results (${pkg.suffix})`,
   artifactName: `unit-${pkg.suffix}-${runAttempt}`,
 }))
 
-const windowsUnitJobs = unitPackages.map((pkg) => ({
+const windowsUnitJobs = windowsUnitPackages.map((pkg) => ({
   ...pkg,
   jobName: `unit-windows-${pkg.suffix}`,
   artifactName: `unit-windows-${pkg.suffix}-${runAttempt}`,
@@ -148,10 +191,9 @@ describe("ci workflow", () => {
     }
   })
 
-  test("splits Windows unit signals by package without publishing advisory check runs", () => {
+  test("keeps Windows unit package and shard signals advisory", () => {
     const parsed = parseWorkflow(workflowPath)
     const job = parsed.jobs?.[windowsUnitJobName]
-    const matrixIncludes = job?.strategy?.matrix?.include ?? []
 
     expect(job?.name).toBe("unit-windows-${{ matrix.package }}")
     expect(job?.needs).toBe("changes")
@@ -162,6 +204,10 @@ describe("ci workflow", () => {
     expect(job?.strategy?.["fail-fast"]).toBe(false)
     expect(job?.permissions).toEqual({ contents: "read" })
     expect(job?.defaults?.run?.shell).toBe("bash")
+    expect(stepByName(windowsUnitJobName, "Prepare unit artifact directory")?.run).toBe(
+      'mkdir -p "$(dirname "${{ matrix.report_path }}")"',
+    )
+    expect(stepByName(windowsUnitJobName, "Prepare unit artifact directory")?.if).toBe("matrix.uses_turbo == false")
     expect(stepByName(windowsUnitJobName, "unit")?.id).toBe("unit")
     expect(stepByName(windowsUnitJobName, "unit")?.["continue-on-error"]).toBe(true)
     expect(stepByName(windowsUnitJobName, "unit")?.run).toContain("${{ matrix.command }}")
@@ -175,7 +221,8 @@ describe("ci workflow", () => {
     )
     expect(stepByName(windowsUnitJobName, "Upload unit artifacts")?.with?.path).toBe("${{ matrix.report_path }}")
 
-    const turboCacheStep = steps(windowsUnitJobName).filter((step) => step.uses?.startsWith("actions/cache@"))[1]
+    const turboCacheStep = steps(windowsUnitJobName).find((step) => step.with?.path === ".turbo/cache")
+    expect(turboCacheStep?.if).toBe("matrix.uses_turbo")
     expect(turboCacheStep?.with?.key).toBe(
       "turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-${{ hashFiles('turbo.json', '**/package.json', 'bun.lock') }}-${{ github.sha }}",
     )
@@ -183,16 +230,23 @@ describe("ci workflow", () => {
       "turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-${{ hashFiles('turbo.json', '**/package.json', 'bun.lock') }}-\n" +
         "turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-\n",
     )
+  })
+
+  test("defines Windows unit packages and opencode shards", () => {
+    const parsed = parseWorkflow(workflowPath)
+    const job = parsed.jobs?.[windowsUnitJobName]
+    const matrixIncludes = job?.strategy?.matrix?.include ?? []
 
     expect(matrixIncludes).toEqual(
-      windowsUnitJobs.map(({ jobName, command, reportPath }) => ({
+      windowsUnitJobs.map(({ jobName, usesTurbo, command, reportPath }) => ({
         package: jobName.replace("unit-windows-", ""),
+        uses_turbo: usesTurbo,
         command,
         report_path: reportPath,
       })),
     )
 
-    for (const { jobName, command, reportPath, artifactName } of windowsUnitJobs) {
+    for (const { jobName, artifactName } of windowsUnitJobs) {
       expect(parsed.jobs?.[jobName]).toBeUndefined()
       expect(artifactName).toBe(`unit-windows-${jobName.replace("unit-windows-", "")}-${runAttempt}`)
     }
@@ -210,6 +264,9 @@ describe("ci workflow", () => {
     expect(needs).not.toContain("unit-windows-app")
     expect(needs).not.toContain("unit-windows-desktop")
     expect(needs).not.toContain("unit-windows-opencode")
+    expect(needs).not.toContain("unit-windows-opencode-session")
+    expect(needs).not.toContain("unit-windows-opencode-config-project")
+    expect(needs).not.toContain("unit-windows-opencode-server-tools")
     expect(validate?.env?.DOCS_ONLY).toBe("${{ needs.changes.outputs.docs_only }}")
     expect(validate?.env?.TYPECHECK_RESULT).toBe("${{ needs.typecheck.result }}")
     expect(validate?.env?.UNIT_APP_RESULT).toBe("${{ needs['unit-app'].result }}")
--- a/packages/opencode/test/github/ci-workflow.test.ts
+++ b/packages/opencode/test/github/ci-workflow.test.ts
@@ -1,9 +1,11 @@
 import { describe, expect, test } from "bun:test"
+import { existsSync, readdirSync, statSync } from "node:fs"
 import path from "node:path"
 import { parseWorkflow, readWorkflow } from "./workflow-parser"
 
 const repoRoot = path.join(import.meta.dir, "../../../..")
 const workflowPath = path.join(repoRoot, ".github", "workflows", "ci.yml")
+const opencodeTestRoot = path.join(repoRoot, "packages", "opencode", "test")
 
 const pinned = {
   checkout: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd",
@@ -104,6 +106,47 @@ function checkoutStep(job: string) {
 
 function stepByName(job: string, name: string) {
   return steps(job).find((step) => step.name === name)
+}
+
+function toPosix(relativePath: string) {
+  return relativePath.split(path.sep).join("/")
+}
+
+function isTestFile(filePath: string) {
+  return /\.(test|spec)\.(ts|tsx|js|mjs|cjs)$/.test(filePath)
+}
+
+function listOpencodeTestFiles(dir = opencodeTestRoot): string[] {
+  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+    const fullPath = path.join(dir, entry.name)
+    if (entry.isDirectory()) return listOpencodeTestFiles(fullPath)
+    if (!entry.isFile() || !isTestFile(entry.name)) return []
+
+    return [`test/${toPosix(path.relative(opencodeTestRoot, fullPath))}`]
+  })
+}
+
+function expandOpencodeTestPath(testPath: string): string[] {
+  const fullPath = path.join(repoRoot, "packages", "opencode", testPath)
+  if (!existsSync(fullPath)) {
+    throw new Error(`Windows opencode shard path does not exist: ${testPath}`)
+  }
+  if (statSync(fullPath).isDirectory()) {
+    return listOpencodeTestFiles(fullPath)
+  }
+  return [testPath]
+}
+
+function testPathArgs(command: string) {
+  const testArgs = command.split(" bun test ")[1]
+  if (!testArgs) {
+    throw new Error(`Windows opencode shard command does not invoke bun test: ${command}`)
+  }
+  return testArgs.split(/\s+/).filter((arg) => arg.startsWith("test/"))
+}
+
+function isWindowsOpencodeShard(item: Record<string, unknown>): item is { command: string; package: string } {
+  return item.uses_turbo === false && typeof item.package === "string" && typeof item.command === "string"
 }
 
 describe("ci workflow", () => {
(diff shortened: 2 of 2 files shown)


Astro-Han:
@coderabbitai review

.github/workflows/ci.yml
```diff
@@ -347,9 +347,9 @@ jobs:
     needs: changes
     if: needs.changes.outputs.docs_only != 'true'
     runs-on: windows-latest
-    # Windows unit jobs are advisory and package-scoped. The 20-minute budget
-    # gives each matrix child room for Windows runner setup and install overhead
-    # while still surfacing the package that timed out.
+    # Windows unit jobs are advisory. The 20-minute budget gives each matrix
```
Astro-Han: The comment update removed 'package-scoped' but didn't clarify that shards are now used alongside packages. Consider: 'Windows unit jobs are advisory, package-scoped or shard-scoped.'

.github/workflows/ci.yml
```diff
@@ -359,9 +359,69 @@ jobs:
           - package: app
             command: bun turbo test:ci --filter=@opencode-ai/app
             report_path: packages/app/.artifacts/unit/junit.xml
-          - package: opencode
-            command: bun turbo test:ci --filter=opencode
-            report_path: packages/opencode/.artifacts/unit/junit.xml
+          - package: opencode-session
+            command: >-
+              cd packages/opencode && mkdir -p .artifacts/unit && bun test
+              --timeout 30000
```
Astro-Han, line 374: The shard commands duplicate `mkdir -p .artifacts/unit` inline. Consider moving this to a separate 'Prepare artifacts directory' step before the matrix runs, or relying on turbo's built-in artifact handling like the app/desktop jobs do.

.github/workflows/ci.yml
```diff
@@ -359,9 +359,69 @@ jobs:
...
+              --reporter=junit
+              --reporter-outfile=.artifacts/unit/junit-windows-session.xml
```
Astro-Han, line 376: Hardcoded `--timeout 30000` (30s per test) differs from the turbo-based jobs which inherit timeout from turbo.json. Is this intentional? If so, document why; otherwise consider matching turbo's defaults.

.github/workflows/ci.yml
```diff
@@ -359,9 +359,69 @@ jobs:
...
+            command: >-
+              cd packages/opencode && mkdir -p .artifacts/unit && bun test
+              --timeout 30000
+              --reporter=junit
+              --reporter-outfile=.artifacts/unit/junit-windows-config-project.xml
+              test/config
+              test/project
+              test/file
+              test/github
+            report_path: packages/opencode/.artifacts/unit/junit-windows-config-project.xml
```
Astro-Han, line 399: The test directory list is hardcoded and must be manually updated when new tests are added. While the coverage contract test catches mismatches, consider adding a comment explaining how shard allocation is maintained, or documenting the shard responsibilities (e.g., 'session: session lifecycle and plugin system').

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -37,14 +39,49 @@ const unitPackages = [
   },
 ] as const
 
-const linuxUnitJobs = unitPackages.map((pkg) => ({
+const windowsOpencodeShards = [
+  {
+    suffix: "opencode-session",
+    command:
```
Astro-Han, line 49: The test file list in `windowsOpencodeShards` must stay synchronized with ci.yml's matrix.include. Currently this duplication is required for the coverage contract test, but any mismatch will cause test failures. Consider a single source of truth pattern (e.g., a script that generates both) or at least add a comment noting the sync requirement.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -63,6 +100,37 @@ function stepByName(job: string, name: string) {
...
+function isTestFile(filePath: string) {
+  return /\.(test|spec)\.(ts|tsx|js)$/.test(filePath)
+}
+
+function listOpencodeTestFiles(dir = opencodeTestRoot): string[] {
+  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+    const fullPath = path.join(dir, entry.name)
+    if (entry.isDirectory()) return listOpencodeTestFiles(fullPath)
+    if (!entry.isFile() || !isTestFile(entry.name)) return []
+
```
Astro-Han, line 124: The regex `/\.(test|spec)\.(ts|tsx|js)$/` misses `.mjs` and `.cjs` test files. While unlikely in this TypeScript codebase, the pattern should be complete: `/\.(test|spec)\.(ts|tsx|js|mjs|cjs)$/`

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -63,6 +100,37 @@ function stepByName(job: string, name: string) {
...
+  if (!existsSync(fullPath)) return [testPath]
+  if (statSync(fullPath).isDirectory()) {
+    return listOpencodeTestFiles(fullPath)
+  }
+  return [testPath]
+}
+
+function isWindowsOpencodeShard(item: Record<string, unknown>): item is { command: string; package: string } {
+  return typeof item.package === "string" && item.package.startsWith("opencode-") && typeof item.command === "string"
+}
```
Astro-Han, line 150: When `testPath` doesn't exist, the function returns `[testPath]` which will be added to `extra`. However, this could mask errors if someone typos a test directory in ci.yml. Consider throwing or logging a warning instead of silently returning the nonexistent path.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -198,6 +266,47 @@ describe("ci workflow", () => {
...
+    const coverage = new Map<string, string[]>()
+    const extra: string[] = []
+
+    expect(opencodeShards.map((item) => item.package)).toEqual([
+      "opencode-session",
+      "opencode-config-project",
+      "opencode-server-tools",
+    ])
+
+    for (const item of opencodeShards) {
```
Astro-Han, line 316: The regex `/\btest\/\S+/g` might match unintended strings if the command contains non-test references (e.g., paths in other CLI args). While unlikely given current command structure, consider a more specific pattern like `/--reporter-outfile=.*? test\/\S+/g` after the outfile arg, or simply split on spaces and filter for `test/` prefixed args.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -198,6 +266,47 @@ describe("ci workflow", () => {
...
+  test("covers each opencode test file exactly once across Windows opencode shards", () => {
+    const parsed = parseWorkflow(workflowPath)
+    const matrixIncludes = parsed.jobs?.[windowsUnitJobName]?.strategy?.matrix?.include ?? []
+    const opencodeShards = matrixIncludes.filter(isWindowsOpencodeShard)
+    const allTestFiles = listOpencodeTestFiles().sort()
+    const allTestFilesSet = new Set(allTestFiles)
+    const coverage = new Map<string, string[]>()
+    const extra: string[] = []
+
+    expect(opencodeShards.map((item) => item.package)).toEqual([
```
Astro-Han, line 310: The test assumes shards will always be named 'opencode-session', 'opencode-config-project', 'opencode-server-tools'. If shard names change, this assertion will fail. Consider a softer check: `expect(opencodeShards.length).toBe(3)` and verify no overlap/duplicates without hardcoding names.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han, line 316: The regex might match unintended strings if the command contains non-test references. While unlikely given current command structure, consider a more specific approach: split the command by spaces and filter for arguments starting with 'test/'.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -148,7 +216,7 @@ describe("ci workflow", () => {
     }
```
Astro-Han, line 234: The test 'splits Windows unit signals and shards opencode without publishing advisory check runs' has a long title that's hard to read. Consider splitting into two tests: one for the advisory/check-run behavior, one for the shard structure verification.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -198,6 +266,47 @@ describe("ci workflow", () => {
...
   })
```
Astro-Han, line 297: The new test 'covers each opencode test file exactly once across Windows opencode shards' runs filesystem operations on every CI run. This is acceptable for coverage verification, but consider whether this test should be skipped in non-CI environments or cached somehow.

.github/workflows/ci.yml
```diff
@@ -359,9 +359,69 @@ jobs:
           - package: app
             command: bun turbo test:ci --filter=@opencode-ai/app
             report_path: packages/app/.artifacts/unit/junit.xml
-          - package: opencode
-            command: bun turbo test:ci --filter=opencode
-            report_path: packages/opencode/.artifacts/unit/junit.xml
+          - package: opencode-session
```
Astro-Han, line 370: The package name 'opencode-session' suggests this shard focuses on session tests, but it also includes plugin, permission, util, skill, npm, and index-runtime-namespace. Consider a more descriptive name like 'opencode-session-plugin' or document why these tests are grouped together.

.github/workflows/ci.yml
Astro-Han, line 399: The server-tools shard contains the majority of test directories (28 directories). This might still be the long pole. Consider checking actual test counts per directory to balance shard runtime more evenly, or document that server-tools is intentionally larger due to faster individual tests.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Keeping these names hardcoded because they are the stable advisory check names this PR is meant to expose. The coverage test still verifies no overlap, no extras, and no missing opencode test files.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Leaving this unskipped. The test only scans local test paths, and that filesystem check is the guard that catches shard drift when new opencode tests are added.

.github/workflows/ci.yml
```diff
@@ -359,9 +360,78 @@ jobs:
...
+              --reporter=junit
+              --reporter-outfile=.artifacts/unit/junit-windows-config-project.xml
+              test/config
+              test/project
+              test/file
+              test/github
+            report_path: packages/opencode/.artifacts/unit/junit-windows-config-project.xml
+          # Server tools carries many smaller and faster directories, which is
+          # why it has more entries than the other two opencode shards.
+          - package: opencode-server-tools
```
Astro-Han, line 402: **P1**: The turbo cache does not benefit opencode shards since they use `bun test` directly, not `turbo`. The shards do not invoke turbo, so caching `.turbo/cache` for them is unnecessary.

Consider either:
1. Removing the turbo cache step for opencode shards (conditional `if:` on matrix.package)
2. Updating the comment to clarify turbo cache only applies to app/desktop entries

The current comment implies the cache helps all matrix children, which is misleading.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -1,9 +1,11 @@
 import { describe, expect, test } from "bun:test"
+import { existsSync, readdirSync, statSync } from "node:fs"
 import path from "node:path"
```
Astro-Han, line 3: **P2**: Import ordering inconsistency. The `node:fs` import is placed between `bun:test` and `node:path`, disrupting the import order.

Consider ordering by source type: external packages (`bun:test`), then Node builtins (`node:fs`, `node:path`), then local imports.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -37,14 +39,51 @@ const unitPackages = [
...
+
+const windowsUnitPackages = [
+  {
+    suffix: "app",
+    command: "bun turbo test:ci --filter=@opencode-ai/app",
+    reportPath: "packages/app/.artifacts/unit/junit.xml",
+  },
+  ...windowsOpencodeShards,
+  {
+    suffix: "desktop",
```
Astro-Han, line 78: **P2**: The `windowsOpencodeShards` constant duplicates the ci.yml matrix entries. While the coverage test catches drift, this still requires manual sync when adding/removing shards.

Consider adding a comment noting the intentional duplication for coverage verification, or extracting shard paths from the parsed workflow instead of hardcoding.

.github/workflows/ci.yml
```diff
@@ -359,9 +360,78 @@ jobs:
...
             command: bun turbo test:ci --filter=@opencode-ai/app
             report_path: packages/app/.artifacts/unit/junit.xml
-          - package: opencode
-            command: bun turbo test:ci --filter=opencode
-            report_path: packages/opencode/.artifacts/unit/junit.xml
+          # Direct bun test keeps opencode's test:ci timeout while allowing
+          # shard-local paths and reporter output. Shards are balanced by
+          # observed runtime, not directory count.
+          # Session also carries plugin, permission, util, skill, and root
+          # tests to keep the Windows shard runtimes close.
```
Astro-Han, line 369: **P1**: The `package` values `opencode-session`, `opencode-config-project`, `opencode-server-tools` are shard identifiers, not actual package names. This works for cache key uniqueness, but the naming could be confusing.

Consider adding a comment clarifying that these are synthetic shard names for matrix differentiation, not package.json package names.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -63,6 +102,43 @@ function stepByName(job: string, name: string) {
   return steps(job).find((step) => step.name === name)
 }
 
+function toPosix(relativePath: string) {
```
Astro-Han, line 111: **P2**: The `isWindowsOpencodeShard` type guard uses `package.startsWith("opencode-")` to identify shards. If a future package is named `opencode-*` (e.g., `opencode-cli`), this would incorrectly classify it as a shard.

Consider using an explicit marker like a `shard: true` property in the matrix, or checking against the known shard names explicitly.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -148,10 +224,9 @@ describe("ci workflow", () => {
     }
   })
```
Astro-Han, line 236: **P2**: Test name change loses specificity. The original name "splits Windows unit signals by package without publishing advisory check runs" explicitly mentioned the package-splitting aspect, but the new name "keeps Windows unit signals advisory" omits it.

The matrix assertion was moved to a separate test, so this refactor is reasonable, but consider keeping the splitting context in the test name or description.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -63,6 +102,43 @@ function stepByName(job: string, name: string) {
...
+  })
+}
+
+function expandOpencodeTestPath(testPath: string): string[] {
+  const fullPath = path.join(repoRoot, "packages", "opencode", testPath)
+  if (!existsSync(fullPath)) {
+    throw new Error(`Windows opencode shard path does not exist: ${testPath}`)
+  }
+  if (statSync(fullPath).isDirectory()) {
+    return listOpencodeTestFiles(fullPath)
```
Astro-Han, line 135: **P2**: The `testPathArgs` function splits the command by whitespace and filters for arguments starting with `test/`. If a future command has flags like `--test-foo` (unlikely but possible), this would incorrectly match.

Consider using a more robust pattern like filtering for paths that also contain `/` after `test/`, or explicitly extracting paths after detecting the `bun test` command boundary.

.github/workflows/ci.yml
Astro-Han: Fixed. The Turbo cache step is now gated by matrix.uses_turbo, and the comment states that only app and desktop invoke Turbo in this Windows matrix.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. Imports are grouped by source type: external package, Node builtins, then local imports.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. The comment now calls out the intentional duplication and why it exists: the test pins the exact ci.yml shard contract while coverage expansion catches drift.

.github/workflows/ci.yml
Astro-Han: Fixed. Added a workflow comment clarifying that the opencode-* matrix package values are synthetic shard names, not package.json names.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. Added an explicit opencode_shard matrix marker and changed the type guard to use that marker instead of inferring from the package prefix.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. The test name now keeps the package and shard context: keeps Windows unit package and shard signals advisory.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. testPathArgs now finds the bun test boundary first, then extracts only test path arguments with a stricter test/.+ pattern.

.github/workflows/ci.yml
```diff
@@ -347,22 +347,104 @@ jobs:
...
           - package: app
+            uses_turbo: true
+            opencode_shard: false
             command: bun turbo test:ci --filter=@opencode-ai/app
             report_path: packages/app/.artifacts/unit/junit.xml
-          - package: opencode
-            command: bun turbo test:ci --filter=opencode
-            report_path: packages/opencode/.artifacts/unit/junit.xml
+          # The opencode-* package values below are synthetic shard names for
+          # readable advisory checks and matrix keys, not package.json names.
```
Astro-Han, line 364: P2: The shard definitions are hardcoded both in `ci.yml` (matrix entries) and `ci-workflow.test.ts` (`windowsOpencodeShards` array). While the coverage contract catches workflow drift, modifying the test file without updating the workflow would pass locally but fail in CI. Consider adding a comment in the test file noting the dual-source synchronization requirement.

The ordering of test directories within each shard's `command` field also differs in presentation (test file uses single-line strings, workflow uses YAML folded block scalars with `>-`), though `testPathArgs` correctly handles both due to space-split parsing.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -37,14 +41,62 @@ const unitPackages = [
...
+const windowsUnitPackages = [
+  {
+    suffix: "app",
+    usesTurbo: true,
+    opencodeShard: false,
+    command: "bun turbo test:ci --filter=@opencode-ai/app",
+    reportPath: "packages/app/.artifacts/unit/junit.xml",
+  },
+  ...windowsOpencodeShards,
+  {
```
Astro-Han, line 77: P3: The hardcoded shard list at lines 83-85 duplicates the contract defined in `windowsOpencodeShards`. While this provides explicit coverage verification, consider whether the duplication adds value versus deriving shard names from `windowsOpencodeShards.map(s => s.suffix)` directly. The current approach is fine if you want to pin the shard names explicitly as part of the contract, but the comment could clarify this intention.

Note: The hardcoded expectation ensures the shard names cannot drift silently, but any reordering or addition would require updating two locations (here and `windowsOpencodeShards`).

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -63,6 +115,48 @@ function stepByName(job: string, name: string) {
...
+function toPosix(relativePath: string) {
+  return relativePath.split(path.sep).join("/")
+}
+
+function isTestFile(filePath: string) {
+  return /\.(test|spec)\.(ts|tsx|js|mjs|cjs)$/.test(filePath)
+}
+
+function listOpencodeTestFiles(dir = opencodeTestRoot): string[] {
+  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
```
Astro-Han, line 120: P3: The `testPathArgs` regex `/^test\/.+/` correctly filters to paths starting with `test/`, but consider whether there's any edge case where the command might contain a non-test argument that starts with `test/` (unlikely in this context). The function assumes the test paths always appear after `bun test` and before any other flags, which matches the actual command structure. The implementation is solid for the current use case.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -176,28 +272,78 @@ describe("ci workflow", () => {
...
+          coverage.set(file, [...(coverage.get(file) ?? []), item.package])
+        }
+      }
+    }
+
+    const missing = allTestFiles.filter((file) => !coverage.has(file))
+    const duplicates = [...coverage.entries()]
+      .filter(([, shardNames]) => shardNames.length > 1)
+      .map(([file, shardNames]) => ({ file, shards: shardNames }))
+
```
Astro-Han, line 337: P3: The coverage test only validates matrix entries where `opencode_shard === true`. Consider adding an explicit assertion that `app` and `desktop` matrix entries have `opencode_shard === false` and `uses_turbo === true`. While the matrix equality check at lines 274-281 provides implicit coverage, explicit assertions would make the contract clearer and catch misconfiguration of the non-shard entries.

The current test structure relies on:
1. `isWindowsOpencodeShard` filtering to only check `opencode_shard === true` items
2. The matrix equality check covering all entries including `app` and `desktop`

If you want explicit per-field validation for all matrix entry types, split into separate test cases.

.github/workflows/ci.yml
Astro-Han: Fixed in the test file comment. It now explicitly calls this an intentional dual source with ci.yml and says to update ci.yml and the shard list together.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed with a local comment at the hardcoded assertion. The duplicate name list is intentional because this coverage test pins the public advisory check names explicitly.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: No code change here. Keeping the current parser because it already finds the bun test boundary first, filters only test/.+ path args, and expandOpencodeTestPath fails if a matching argument is not a real opencode test path.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. Added an explicit metadata assertion for every Windows matrix entry, including app and desktop having uses_turbo=true and opencode_shard=false.

.github/workflows/ci.yml
```diff
@@ -347,22 +347,104 @@ jobs:
...
+    # timed out.
     timeout-minutes: 20
     continue-on-error: true
     strategy:
       fail-fast: false
       matrix:
         include:
           - package: app
+            uses_turbo: true
+            opencode_shard: false
```
Astro-Han: **P2**: `uses_turbo` and `opencode_shard` are redundant — across all five entries `opencode_shard === !uses_turbo`, and only `uses_turbo` is consumed by the workflow (the cache `if:`). `opencode_shard` exists solely so the test file can filter the matrix, which it could do just as well from `!uses_turbo` or `package.startsWith('opencode-')`. Carrying two booleans that always flip together is an extra hand-maintained field and an extra line per matrix entry. Suggest dropping `opencode_shard` and filtering in the test via the negation.

.github/workflows/ci.yml
```diff
@@ -347,22 +347,104 @@ jobs:
     needs: changes
     if: needs.changes.outputs.docs_only != 'true'
     runs-on: windows-latest
-    # Windows unit jobs are advisory and package-scoped. The 20-minute budget
-    # gives each matrix child room for Windows runner setup and install overhead
-    # while still surfacing the package that timed out.
+    # Windows unit jobs are advisory, package-scoped or shard-scoped. The
+    # 20-minute budget gives each matrix child room for Windows runner setup
+    # and install overhead while still surfacing the package or shard that
+    # timed out.
```
Astro-Han: **P3**: With opencode now sharded into three matrix children, the worst-case opencode Windows wall-time on a stall is effectively `3 × 20 = 60` min (parallel, so not serialized, but still 3× the minutes consumed). Worth a single sentence in this comment so a future reader doesn't trip over the assumption that the 20-min cap still bounds opencode signal cost the way it used to.

.github/workflows/ci.yml
```diff
@@ -396,17 +478,20 @@ jobs:
...
+          # Only app and desktop invoke Turbo in this Windows matrix.
           key: turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-${{ hashFiles('turbo.json', '**/package.json', 'bun.lock') }}-${{ github.sha }}
           restore-keys: |
             turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-${{ hashFiles('turbo.json', '**/package.json', 'bun.lock') }}-
             turbo-${{ runner.os }}-unit-windows-${{ matrix.package }}-
 
       - run: bun install --frozen-lockfile
 
+      - name: Prepare unit artifact directory
+        run: mkdir -p "$(dirname "${{ matrix.report_path }}")"
```
Astro-Han, line 488: **P3**: This step runs for every matrix child, including `app` and `desktop` whose `test:ci` already does its own `mkdir -p .artifacts/unit`. It's a cheap no-op in those cases but the step name ("Prepare unit artifact directory") reads like it's load-bearing for all shards when it's really only needed for the opencode shards — those don't go through `test:ci`. Either gate it (`if: '!matrix.uses_turbo'`) or leave a one-line comment that it's a safety net for the shards.

Astro-Han (commented):
Nits only, no blockers. The shard plan and coverage contract are sound and the TDD-RED evidence for the missing `test/auth` entry is convincing. Leaving a handful of rough edges below, mostly P2/P3.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -1,9 +1,13 @@
 import { describe, expect, test } from "bun:test"
+
```
Astro-Han: **P3**: The rest of this file uses flat (ungrouped) imports. Introducing blank-line grouping here is inconsistent with the existing style; Biome formatter is disabled project-wide so there's no auto-fix. Either group all three blocks or drop the blanks to match.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -63,6 +115,48 @@ function stepByName(job: string, name: string) {
...
+}
+
+function testPathArgs(command: string) {
+  const args = command.split(/\s+/)
+  const bunTestIndex = args.findIndex((arg, index) => arg === "test" && args[index - 1] === "bun")
+  if (bunTestIndex === -1) {
+    throw new Error(`Windows opencode shard command does not invoke bun test: ${command}`)
+  }
+  return args.slice(bunTestIndex + 1).filter((arg) => /^test\/.+/.test(arg))
+}
```
Astro-Han, line 146: **P3**: The `args[i] === 'test' && args[i-1] === 'bun'` heuristic is more clever than it needs to be given all shard commands are hand-authored in `windowsOpencodeShards` a few lines above. Since each command already uses the exact form `cd packages/opencode && bun test …`, a direct `command.split(' bun test ')[1]?.split(/\s+/).filter(a => a.startsWith('test/'))` or even a regex is equally safe. Not a bug — the throw on mismatch is a fine guardrail — but the current code invites thought about edge cases that don't exist.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -176,28 +272,92 @@ describe("ci workflow", () => {
     expect(stepByName(windowsUnitJobName, "Upload unit artifacts")?.with?.path).toBe("${{ matrix.report_path }}")
 
     const turboCacheStep = steps(windowsUnitJobName).filter((step) => step.uses?.startsWith("actions/cache@"))[1]
```
Astro-Han: **P2**: The new `expect(turboCacheStep?.if).toBe('matrix.uses_turbo')` assertion depends on `…[1]` picking the turbo cache specifically. If a third `actions/cache@` step is ever added (or reordered) before the turbo one, this picks the wrong step and the assertion will either false-pass or fail with a confusing message. Consider selecting by a stable property, e.g. `find((s) => s.with?.path === '.turbo/cache')`, so the test stays pinned to the step by identity rather than position. Pre-existing, but the PR adds a new claim on top of it.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -176,28 +272,92 @@ describe("ci workflow", () => {
...
+      package: packageName,
+      uses_turbo,
+      opencode_shard,
+    }))).toEqual([
+      { package: "app", uses_turbo: true, opencode_shard: false },
+      { package: "opencode-session", uses_turbo: false, opencode_shard: true },
+      { package: "opencode-config-project", uses_turbo: false, opencode_shard: true },
+      { package: "opencode-server-tools", uses_turbo: false, opencode_shard: true },
+      { package: "desktop", uses_turbo: true, opencode_shard: false },
+    ])
```
Astro-Han: **P3**: This block is strictly redundant with the preceding `expect(matrixIncludes).toEqual(windowsUnitJobs.map(...))` — both sides derive from the same `windowsUnitPackages` source, so if the first passes, this one cannot fail. If the intent is to pin the public shard names as a human-readable checklist, a one-line comment saying so helps; otherwise this can be deleted. (The similar inline-name pin in the coverage test at L329 has a different purpose — it asserts ordering of the *filtered* shards — and is worth keeping.)

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -176,28 +272,92 @@ describe("ci workflow", () => {
...
+    const missing = allTestFiles.filter((file) => !coverage.has(file))
+    const duplicates = [...coverage.entries()]
+      .filter(([, shardNames]) => shardNames.length > 1)
+      .map(([file, shardNames]) => ({ file, shards: shardNames }))
+
+    expect({ duplicates, extra: extra.sort(), missing }).toEqual({
+      duplicates: [],
+      extra: [],
+      missing: [],
+    })
```
Astro-Han, line 357: **P2**: When this contract is broken (new test file, moved directory, shard drift), the failure arrives as a single `toEqual` diff on `{ duplicates, extra, missing }` with no hint about which shard should own the new file. Suggest: on failure, pick the shard with the smallest cumulative file count and name it in the error message, e.g. `Uncovered files: test/newdir/foo.test.ts. Add them to one of: opencode-session | opencode-config-project | opencode-server-tools`. Cheap ergonomic win — the whole point of the contract is to catch drift at CI-time, and future-you will be the one reading the error.

packages/opencode/test/github/ci-workflow.test.ts
```diff
@@ -63,6 +115,48 @@ function stepByName(job: string, name: string) {
...
+  return /\.(test|spec)\.(ts|tsx|js|mjs|cjs)$/.test(filePath)
+}
+
+function listOpencodeTestFiles(dir = opencodeTestRoot): string[] {
+  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+    const fullPath = path.join(dir, entry.name)
+    if (entry.isDirectory()) return listOpencodeTestFiles(fullPath)
+    if (!entry.isFile() || !isTestFile(entry.name)) return []
+
+    return [`test/${toPosix(path.relative(opencodeTestRoot, fullPath))}`]
```
Astro-Han, line 125: **P3**: `listOpencodeTestFiles` walks every entry under `test/` on every call, and the new coverage test also invokes `expandOpencodeTestPath(testPath) → listOpencodeTestFiles(subdir)` for each shard path, re-walking subtrees already visible to the single root walk. For this tree size it's trivially fast, but a single walk up front that groups files by top-level directory would be both faster and simpler (no recursion inside `expandOpencodeTestPath`). Take-it-or-leave-it.

.github/workflows/ci.yml
Astro-Han: Fixed. Dropped opencode_shard from the workflow matrix and test fixtures; tests now identify opencode shards as the non-Turbo Windows entries.

.github/workflows/ci.yml
Astro-Han: Fixed. The Windows advisory comment now states that the 20-minute timeout applies per matrix child and that opencode has three parallel shards.

.github/workflows/ci.yml
Astro-Han: Fixed. The artifact directory preparation step now runs only for non-Turbo entries, which are the direct opencode shard commands.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. Dropped the blank-line import grouping to match the file style.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. testPathArgs now uses the explicit hand-authored command shape and splits on the bun test boundary directly.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. The Turbo cache assertion now finds the step by with.path == .turbo/cache instead of relying on cache step position.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. Removed the redundant metadata assertion block; the full matrix equality assertion already covers uses_turbo for all entries.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: Fixed. The coverage test now throws a targeted drift message with uncovered files, unknown paths, duplicates, shard choices, and the smallest current shard as a starting suggestion.

packages/opencode/test/github/ci-workflow.test.ts
Astro-Han: No code change. Keeping the current recursive walk because this tree is small, the test is still fast, and the existing expand-by-path structure keeps the coverage check direct. Pre-indexing would add complexity without meaningful wall-time benefit.

PR #147, opened 2026-04-22
test: guard fresh startup release verification

### LINKED ISSUE #144: [Bug] v0.2.5 first launch hangs on loading screen
[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.5', packaged: true }
[2026-04-22 21:26:16.300] [info]  auto updater configured { channel: 'latest', allowPrerelease: false, allowDowngrade: true, currentVersion: '0.2.5' }
[2026-04-22 21:26:16.300] [info]  spawning sidecar { url: 'http://127.0.0.1:59635' }
[2026-04-22 21:26:16.767] [info]  sidecar connection started { url: 'http://127.0.0.1:59635' }
[2026-04-22 21:26:18.129] [info]  awaiting server ready
[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }

### LINKED ISSUE #145: [Bug] v0.2.5 cannot restart after first launch hang
### 环境信息
- **版本**: v0.2.5
- **操作系统**: macOS 25.4.0 (Darwin arm64)
- **安装方式**: 从浏览器下载安装

### 问题描述
PawWork 在使用过程中突然卡住，窗口完全无响应。强制退出后无法重新启动应用。

### 复现步骤
1. 正常使用 PawWork 一段时间
2. 应用突然卡住（界面无响应）
3. Cmd+Q 退出或 killall PawWork
4. 重新打开应用，无法启动

### 预期行为
退出后应该能正常重新启动

### 实际行为
通过 `open -a PawWork` 或直接执行 `/Applications/PawWork.app/Contents/MacOS/PawWork` 都无法启动，进程创建后立即退出。

### 尝试的解决方法
- ✅ `killall PawWork` 强制退出
- ✅ 清理缓存（Cache、Code Cache、GPUCache、Session Storage）
- ✅ 移除 `com.apple.quarantine` 属性
- ❌ 重启应用 - 无效
- ⏳ 重启电脑 - 未尝试

### 日志信息
应用日志路径: `~/Library/Logs/PawWork/main.log`

最后启动记录:
```
[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.5', packaged: true }
[2026-04-22 21:26:16.300] [info]  auto updater configured { channel: 'latest', allowPrerelease: false, allowDowngrade: true, currentVersion: '0.2.5' }
[2026-04-22 21:26:16.300] [info]  spawning sidecar { url: 'http://127.0.0.1:59635' }
[2026-04-22 21:26:16.767] [info]  sidecar connection started { url: 'http://127.0.0.1:59635' }
[2026-04-22 21:26:18.129] [info]  awaiting server ready
[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
```

### 其他信息
- 应用代码签名状态：正常 (`spctl -a -v` 显示 `accepted source=Notarized Developer ID`)
- 二进制文件类型：`Mach-O 64-bit executable arm64`

Astro-Han:
## Summary

Adds release verification coverage for fresh packaged startup logs so a build that reaches `server ready` but never finishes desktop initialization is caught before closing startup-blocking issues.

## Why

PawWork v0.2.5 exposed a first-launch loading hang tracked by #144 and confirmed again by #145. The runtime fix is already on `dev`; this PR hardens the release verification path so the same startup failure mode is checked before the next release is marked done.

## Related Issue

Fixes #144
Fixes #145

## How To Verify

```bash
bun install --frozen-lockfile
bun test scripts/verify-release.test.ts
bun test scripts/verify-release.test.ts src/main/index-sidecar-source.test.ts src/main/server.test.ts scripts/ci-smoke.test.ts
bun run typecheck
```

I also ran `git diff --check`.

## Screenshots or Recordings

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Added a post-release smoke verification step with platform guidance for fresh-user-data startup checks (macOS and Windows) and explicit startup-log assertions.

* **Tests**
  * Expanded startup-log parsing/validation tests covering version mismatches, missing markers, legacy formats, and file-read error handling.

* **Chores**
  * Added startup-log verification utilities and updated the app to emit an explicit "init done" startup marker and to isolate logs for smoke runs.

### DIFF
--- a/.github/RELEASE_CHECKLIST.md
+++ b/.github/RELEASE_CHECKLIST.md
@@ -136,6 +136,80 @@ The helper verifies:
 - `latest.yml` points to `pawwork-win-x64.exe`.
 - `latest-mac.yml` includes both `pawwork-mac-arm64.zip` and `pawwork-mac-x64.zip`.
 
+Also verify a fresh packaged startup before closing startup-blocking issues. The command below is for macOS; override `PAWWORK_RELEASE_APP_PATH` and `PAWWORK_RELEASE_STARTUP_LOG` if the app or log is in a custom location.
+
+```bash
+set -euo pipefail
+smoke_home=/tmp/pawwork-release-smoke/user-data
+smoke_user_data="$smoke_home/ai.pawwork.desktop"
+ready_file="$smoke_user_data/ci-smoke-ready.json"
+app_path=${PAWWORK_RELEASE_APP_PATH:-/Applications/PawWork.app/Contents/MacOS/PawWork}
+startup_log=${PAWWORK_RELEASE_STARTUP_LOG:-$smoke_user_data/logs/main.log}
+app_pid=""
+cleanup() {
+  if [ -n "$app_pid" ]; then
+    kill "$app_pid" 2>/dev/null || true
+  fi
+  rm -rf "$smoke_home"
+}
+trap cleanup EXIT
+rm -rf "$smoke_home"
+PAWWORK_CI_SMOKE=true PAWWORK_CI_SMOKE_HOME="$smoke_home" "$app_path" &
+app_pid=$!
+i=0
+while [ "$i" -lt 60 ]; do
+  test -f "$ready_file" && break
+  sleep 1
+  i=$((i + 1))
+done
+if [ ! -f "$ready_file" ]; then
+  echo "Timed out waiting for $ready_file"
+  exit 1
+fi
+sleep 1
+PAWWORK_RELEASE_STARTUP_LOG="$startup_log" bun packages/desktop-electron/scripts/verify-release.ts vX.Y.Z
+```
+
+The startup log check reads the latest `app starting` block and verifies it reaches `server ready`, `loading task finished`, and `init step done`. This catches first-launch hangs where the sidecar becomes reachable but the desktop shell never opens the main window.
+
+For Windows releases, run the same fresh-user-data check from PowerShell:
+
+```powershell
+$ErrorActionPreference = "Stop"
+$smokeHome = "$env:TEMP\pawwork-release-smoke\user-data"
+$smokeUserData = "$smokeHome\ai.pawwork.desktop"
+$readyFile = "$smokeUserData\ci-smoke-ready.json"
+$appPath = if ($env:PAWWORK_RELEASE_APP_PATH) { $env:PAWWORK_RELEASE_APP_PATH } else { "$env:LOCALAPPDATA\Programs\PawWork\PawWork.exe" }
+$startupLog = if ($env:PAWWORK_RELEASE_STARTUP_LOG) { $env:PAWWORK_RELEASE_STARTUP_LOG } else { "$smokeUserData\logs\main.log" }
+Remove-Item -Recurse -Force $smokeHome -ErrorAction SilentlyContinue
+$previousCiSmoke = $env:PAWWORK_CI_SMOKE
+$previousCiSmokeHome = $env:PAWWORK_CI_SMOKE_HOME
+$previousStartupLog = $env:PAWWORK_RELEASE_STARTUP_LOG
+$env:PAWWORK_CI_SMOKE = "true"
+$env:PAWWORK_CI_SMOKE_HOME = $smokeHome
+$app = Start-Process -FilePath $appPath -PassThru
+try {
+  $ready = $false
+  for ($i = 0; $i -lt 60; $i++) {
+    if (Test-Path $readyFile) {
+      $ready = $true
+      break
+    }
+    Start-Sleep -Seconds 1
+  }
+  if (-not $ready) { throw "Timed out waiting for $readyFile" }
+  Start-Sleep -Seconds 1
+  $env:PAWWORK_RELEASE_STARTUP_LOG = $startupLog
+  bun packages/desktop-electron/scripts/verify-release.ts vX.Y.Z
+} finally {
+  if ($app -and -not $app.HasExited) { Stop-Process -Id $app.Id -Force }
+  if ($null -eq $previousCiSmoke) { Remove-Item Env:PAWWORK_CI_SMOKE -ErrorAction SilentlyContinue } else { $env:PAWWORK_CI_SMOKE = $previousCiSmoke }
+  if ($null -eq $previousCiSmokeHome) { Remove-Item Env:PAWWORK_CI_SMOKE_HOME -ErrorAction SilentlyContinue } else { $env:PAWWORK_CI_SMOKE_HOME = $previousCiSmokeHome }
+  if ($null -eq $previousStartupLog) { Remove-Item Env:PAWWORK_RELEASE_STARTUP_LOG -ErrorAction SilentlyContinue } else { $env:PAWWORK_RELEASE_STARTUP_LOG = $previousStartupLog }
+  Remove-Item -Recurse -Force $smokeHome -ErrorAction SilentlyContinue
+}
+```
+
 Keep `.zip`, `.blockmap`, and `latest*.yml` assets unless updater requirements are proven safe without them.
 
 If verification fails, check the reported missing or malformed asset first, rerun only the affected build phase, and publish the release only after the verification helper passes.
--- a/packages/desktop-electron/scripts/verify-release.test.ts
+++ b/packages/desktop-electron/scripts/verify-release.test.ts
@@ -1,11 +1,17 @@
 import { afterEach, describe, expect, test } from "bun:test"
+import { mkdtemp, rm, writeFile } from "node:fs/promises"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
 
 import {
+  escapeRegExp,
   fetchJson,
   fetchText,
   normalizeTag,
   parseUpdaterFileUrls,
+  readStartupLogFile,
   verifyReleasePayload,
+  verifyStartupLog,
   type GithubRelease,
 } from "./verify-release"
 
@@ -68,6 +74,8 @@ describe("verify-release", () => {
     expect(normalizeTag("0.2.6")).toBe("v0.2.6")
     expect(normalizeTag("v0.2.6")).toBe("v0.2.6")
     expect(() => normalizeTag("vv0.2.6")).toThrow("Invalid release tag")
+    expect(() => normalizeTag("")).toThrow("Invalid release tag")
+    expect(() => normalizeTag("v")).toThrow("Invalid release tag")
     expect(() => normalizeTag("abc")).toThrow("Invalid release tag")
   })
 
@@ -212,6 +220,198 @@ path: pawwork-win-x64.exe
     expect(failures).toContain("latest.yml does not include pawwork-win-x64.exe")
     expect(failures).toContain("latest-mac.yml does not include pawwork-mac-arm64.zip")
     expect(failures).toContain("latest-mac.yml does not include pawwork-mac-x64.zip")
+  })
+
+  test("accepts a complete startup log for the release version", () => {
+    expect(
+      verifyReleasePayload({
+        release: baseRelease,
+        latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+        latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+        startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init done
+`,
+      }),
+    ).toEqual([])
+  })
+
+  test("reports an empty startup log", () => {
+    expect(
+      verifyReleasePayload({
+        release: baseRelease,
+        latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+        latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+        startupLog: "",
+      }),
+    ).toEqual(["Latest startup log does not include any app starting entry"])
+  })
+
+  test("reports a fresh startup log stuck after sidecar readiness", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:16.300] [info]  spawning sidecar { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:16.767] [info]  sidecar connection started { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.129] [info]  awaiting server ready
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include loading task finished")
+    expect(failures).toContain("Latest startup log does not include init step done")
+    expect(failures).toHaveLength(2)
+  })
+
+  test("does not accept awaiting server ready as server ready", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:18.129] [info]  awaiting server ready
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init done
+`,
+    })
+
+    expect(failures).toEqual(["Latest startup log does not include server ready"])
+  })
+
+  test("checks the latest startup attempt instead of an older successful launch", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 20:00:00.000] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 20:00:01.000] [info]  loading task finished
+[2026-04-22 20:00:01.001] [info]  init done
+[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:16.767] [info]  sidecar connection started { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include loading task finished")
+    expect(failures).toContain("Latest startup log does not include init step done")
+    expect(failures).toHaveLength(2)
+  })
+
+  test("reports release version mismatches in the startup log", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.5', packaged: true }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init done
+`,
+    })
+
+    expect(failures).toContain("Latest startup log version does not match expected 0.2.6")
+  })
+
+  test("reports startup logs from unpackaged desktop runs", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: false }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init done
+`,
+    })
+
+    expect(failures).toEqual(["Latest startup log does not include packaged true"])
+  })
+
+  test("reports invalid release tags during startup log verification", () => {
+    expect(
+      verifyStartupLog(
+        `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init done
+`,
+        "v",
+      ),
+    ).toEqual(["Invalid release tag: v. Expected vX.Y.Z or X.Y.Z."])
+  })
+
+  test("reports invalid release tags with other startup failures", () => {
+    expect(
+      verifyStartupLog(
+        `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+`,
+        "v",
+      ),
+    ).toEqual([
+      "Invalid release tag: v. Expected vX.Y.Z or X.Y.Z.",
+      "Latest startup log does not include server ready",
+      "Latest startup log does not include loading task finished",
+      "Latest startup log does not include init step done",
+    ])
+  })
+
+  test("does not accept 'phase: done' in a non-init-step log line", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init step { step: { phase: 'loading' } }
+[2026-04-22 21:26:18.132] [info]  unrelated message containing phase: 'done'
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include init step done")
+  })
+
+  test("does not accept legacy init step done without the dedicated marker", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init step { step: { phase: 'done' } }
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include init step done")
+  })
+
+  test("reads startup log files", async () => {
+    const dir = await mkdtemp(join(tmpdir(), "pawwork-release-log-"))
+    const logPath = join(dir, "main.log")
+
+    try {
+      await writeFile(logPath, "startup log contents", "utf8")
+      await expect(readStartupLogFile(logPath)).resolves.toBe("startup log contents")
+    } finally {
+      await rm(dir, { recursive: true, force: true })
+    }
+  })
+
+  test("reports unreadable startup log files with the path", async () => {
+    const dir = await mkdtemp(join(tmpdir(), "pawwork-release-log-"))
+    const missingPath = join(dir, "missing-main.log")
+
+    try {
+      await expect(readStartupLogFile(missingPath)).rejects.toThrow(
+        new RegExp(`^Failed to read startup log file ${escapeRegExp(missingPath)}: .+`),
+      )
+    } finally {
+      await rm(dir, { recursive: true, force: true })
+    }
   })
 
   test("fetchText reports GitHub rate limit headers on HTTP errors", async () => {
--- a/packages/desktop-electron/scripts/verify-release.ts
+++ b/packages/desktop-electron/scripts/verify-release.ts
@@ -15,6 +15,7 @@ type VerificationInput = {
   release: GithubRelease
   latestYml?: string
   latestMacYml?: string
+  startupLog?: string
 }
 
 const DEFAULT_REPO = "Astro-Han/pawwork"
@@ -103,6 +104,69 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames: 
   }
 }
 
+function latestStartupAttempt(source: string) {
+  const marker = "app starting"
+  const index = source.lastIndexOf(marker)
+  if (index === -1) return undefined
+  return source.slice(index)
+}
+
+function firstLine(source: string) {
+  return source.split(/\r?\n/, 1)[0] ?? ""
+}
+
+function hasInitDone(source: string) {
+  return source.split(/\r?\n/).some((line) => line.trim().endsWith("init done"))
+}
+
+function hasServerReady(source: string) {
+  return source.split(/\r?\n/).some((line) => /\bserver ready\b/.test(line) && /\{\s*url:\s*['"]/.test(line))
+}
+
+function releaseVersion(tag: string) {
+  return normalizeTag(tag).slice(1)
+}
+
+export function escapeRegExp(value: string) {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")
+}
+
+function hasStartupVersion(startupLine: string, expectedVersion: string) {
+  return new RegExp(`version:\\s*['"]${escapeRegExp(expectedVersion)}['"]`).test(startupLine)
+}
+
+function hasPackagedStartup(startupLine: string) {
+  return /packaged:\s*true/.test(startupLine)
+}
+
+export function verifyStartupLog(source: string, expectedTag: string) {
+  const failures: string[] = []
+  const latest = latestStartupAttempt(source)
+
+  if (!latest) {
+    failures.push("Latest startup log does not include any app starting entry")
+    return failures
+  }
+
+  const startupLine = firstLine(latest)
+  let expectedVersion: string | undefined
+  try {
+    expectedVersion = releaseVersion(expectedTag)
+  } catch (error) {
+    failures.push(error instanceof Error ? error.message : String(error))
+  }
+
+  if (expectedVersion && !hasStartupVersion(startupLine, expectedVersion)) {
+    failures.push(`Latest startup log version does not match expected ${expectedVersion}`)
+  }
+  if (!hasPackagedStartup(startupLine)) failures.push("Latest startup log does not include packaged true")
+  if (!hasServerReady(latest)) failures.push("Latest startup log does not include server ready")
+  if (!latest.includes("loading task finished")) failures.push("Latest startup log does not include loading task finished")
+  if (!hasInitDone(latest)) failures.push("Latest startup log does not include init step done")
+
+  return failures
+}
+
 export function verifyReleasePayload(input: VerificationInput) {
   const failures: string[] = []
   const assetNames = new Set(input.release.assets.map((asset) => asset.name))
@@ -125,6 +189,8 @@ export function verifyReleasePayload(input: VerificationInput) {
   for (const asset of ["pawwork-mac-arm64.zip", "pawwork-mac-x64.zip"]) {
     if (!hasUpdaterEntry(latestMacUrls, asset)) failures.push(`latest-mac.yml does not include ${asset}`)
   }
+
+  if (input.startupLog !== undefined) failures.push(...verifyStartupLog(input.startupLog, input.release.tag_name))
 
   return failures
 }
@@ -164,6 +230,15 @@ export async function fetchJson<T>(url: string) {
   }
 }
 
+export async function readStartupLogFile(path: string) {
+  try {
+    return await Bun.file(path).text()
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error)
+    throw new Error(`Failed to read startup log file ${path}: ${message}`)
+  }
+}
+
 async function fetchWithTimeout(url: string) {
   try {
     return await fetch(url, {
@@ -198,7 +273,9 @@ async function main() {
   try {
     const tag = process.argv[2]
     if (!tag) {
-      console.error("Usage: bun packages/desktop-electron/scripts/verify-release.ts <tag> [owner/repo]")
+      console.error(
+        "Usage: bun packages/desktop-electron/scripts/verify-release.ts <tag> [owner/repo] [env: PAWWORK_RELEASE_STARTUP_LOG=/path/to/main.log]",
+      )
       process.exit(2)
     }
 
@@ -212,7 +289,9 @@ async function main() {
     )
     const latestYml = await fetchAssetText(release, "latest.yml")
     const latestMacYml = await fetchAssetText(release, "latest-mac.yml")
-    const failures = verifyReleasePayload({ release, latestYml, latestMacYml })
+    const startupLogPath = process.env.PAWWORK_RELEASE_STARTUP_LOG
+    const startupLog = startupLogPath ? await readStartupLogFile(startupLogPath) : undefined
+    const failures = verifyReleasePayload({ release, latestYml, latestMacYml, startupLog })
 
     if (failures.length) {
       console.error(`Release verification failed for ${repo} ${normalizedTag}:`)
--- a/packages/desktop-electron/src/main/index-sidecar-source.test.ts
+++ b/packages/desktop-electron/src/main/index-sidecar-source.test.ts
@@ -6,6 +6,9 @@ describe("desktop sidecar source guard", () => {
 
     expect(source).toContain("username: PAWWORK_RUNTIME.serverUsername")
     expect(source).toContain("const needsMigration = false")
+    expect(source).toContain('app.setPath("logs", join(app.getPath("userData"), "logs"))')
+    expect(source).toContain('logger.log("server ready", { url: res.url })')
+    expect(source).toContain('logger.log("init done")')
     expect(source).not.toContain("sqliteFileExists")
     expect(source).not.toContain('username: "opencode"')
   })
--- a/packages/desktop-electron/src/main/index.ts
+++ b/packages/desktop-electron/src/main/index.ts
@@ -39,6 +39,10 @@ if (CI_SMOKE_HOME) {
   app.setPath("appData", CI_SMOKE_HOME)
 }
 app.setPath("userData", join(userDataRoot, app.isPackaged ? APP_IDS[CHANNEL] : "ai.pawwork.desktop.dev"))
+if (CI_SMOKE_HOME) {
+  // Keep smoke logs inside the isolated profile so release checks cannot read stale user logs.
+  app.setPath("logs", join(app.getPath("userData"), "logs"))
+}
 const CI_SMOKE_READY_FILE = join(app.getPath("userData"), "ci-smoke-ready.json")
 const { autoUpdater } = pkg
 
@@ -327,6 +331,7 @@ function openMainWindow() {
 function setInitStep(step: InitStep) {
   initStep = step
   logger.log("init step", { step })
+  if (step.phase === "done") logger.log("init done")
   initEmitter.emit("step", step)
 }
 


packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -15,6 +15,7 @@ type VerificationInput = {
   release: GithubRelease
   latestYml?: string
   latestMacYml?: string
+  startupLog?: string
 }
 
 const DEFAULT_REPO = "Astro-Han/pawwork"
```
Astro-Han, line 21: The regex `/init step[\s\S]*phase:\s*['"]done['"]/` uses `[\s\S]*` which matches any characters including newlines. This could cause false positives if `phase: 'done'` appears in unrelated log entries after an `init step` marker.

For example, this log would incorrectly pass validation:
```
[info] init step { step: { phase: 'loading' } }
[info] unrelated message containing phase: 'done'
```

A more precise regex would constrain the match to the `init step` context, such as `/init step\s*\{[^}]*phase:\s*['"]done['"]/` for single-line logs.

Astro-Han (commented):
Overall the implementation is solid and addresses the startup verification gap. A few minor suggestions:

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +214,41 @@ path: pawwork-win-x64.exe
     expect(failures).toContain("latest-mac.yml does not include pawwork-mac-x64.zip")
```
Astro-Han, line 222: Both test cases cover failure scenarios, but there's no test verifying that a complete startup log passes validation. Adding a positive test case would ensure the `verifyStartupLog` function correctly accepts valid logs:

```ts
test("accepts a complete startup log", () => {
  const failures = verifyReleasePayload({
    release: baseRelease,
    latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
    latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
    startupLog: `[info] app starting\n[info] server ready\n[info] loading task finished\n[info] init step { step: { phase: 'done' } }`
  })
  expect(failures).toEqual([])
})
```

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,29 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
+
+  if (!latest.includes("app starting")) failures.push("Startup log does not include app starting")
+  if (!latest.includes("server ready")) failures.push("Latest startup log does not include server ready")
+  if (!latest.includes("loading task finished")) failures.push("Latest startup log does not include loading task finished")
+  if (!hasInitDone(latest)) failures.push("Latest startup log does not include init step done")
+
+  return failures
+}
+
 export function verifyReleasePayload(input: VerificationInput) {
```
Astro-Han, line 170: If `PAWWORK_RELEASE_STARTUP_LOG` points to a non-existent file, `Bun.file(path).text()` throws an unhandled error. Consider adding an existence check or try-catch:

```ts
const startupLogPath = process.env.PAWWORK_RELEASE_STARTUP_LOG
let startupLog: string | undefined
if (startupLogPath) {
  try {
    startupLog = await Bun.file(startupLogPath).text()
  } catch {
    console.error(`Startup log file not found: ${startupLogPath}`)
  }
}
```

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,32 @@ The helper verifies:
 - `latest.yml` points to `pawwork-win-x64.exe`.
 - `latest-mac.yml` includes both `pawwork-mac-arm64.zip` and `pawwork-mac-x64.zip`.
 
+Also verify a fresh packaged startup before closing startup-blocking issues:
+
+```bash
+set -euo pipefail
+smoke_home=/tmp/pawwork-release-smoke/user-data
```
Astro-Han, line 143: The script hardcodes `/Applications/PawWork.app`. If a tester installed the app in a custom location (e.g., portable testing or non-standard install), this path fails. Consider documenting this assumption or making the app path configurable via an environment variable.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,29 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
+  const index = source.lastIndexOf(marker)
+  if (index === -1) return source
+  return source.slice(index)
+}
+
+function hasInitDone(source: string) {
+  return /init step[\s\S]*phase:\s*['"]done['"]/.test(source)
+}
+
+export function verifyStartupLog(source: string) {
```
Astro-Han: This check is redundant. `latestStartupAttempt` returns the slice starting from the last `"app starting"` marker. If the original source doesn't contain the marker, `latest` equals `source` and the same check fails anyway. If it does contain the marker, `latest` definitely includes it.

Consider removing this line or changing the error message to reflect that we're checking for *any* startup attempt in the log (not just the latest one).

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,93 @@ path: pawwork-win-x64.exe
...
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:16.300] [info]  spawning sidecar { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:16.767] [info]  sidecar connection started { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.129] [info]  awaiting server ready
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include loading task finished")
+    expect(failures).toContain("Latest startup log does not include init step done")
```
Astro-Han, line 265: **P1**: Test checks for two specific failure messages but doesn't verify `failures.length`. If the code incorrectly adds extra failures beyond `loading task finished` and `init step done`, this test would still pass. Consider adding `expect(failures.length).toBe(2)` for stricter verification.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,93 @@ path: pawwork-win-x64.exe
...
+[2026-04-22 20:00:01.000] [info]  loading task finished
+[2026-04-22 20:00:01.001] [info]  init step { step: { phase: 'done' } }
+[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:16.767] [info]  sidecar connection started { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include loading task finished")
+    expect(failures).toContain("Latest startup log does not include init step done")
```
Astro-Han, line 299: **P1**: Same issue as above. Test only checks two `toContain` assertions without verifying the complete failure list. The startup log contains `server ready` and `version: '0.2.6'`, so version check passes. But the test doesn't explicitly verify that no other failures are present. Add `expect(failures.length).toBe(2)` for stricter verification.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,46 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
 
+function latestStartupAttempt(source: string) {
+  const marker = "app starting"
+  const index = source.lastIndexOf(marker)
+  if (index === -1) return undefined
+  return source.slice(index)
+}
+
+function hasInitDone(source: string) {
+  return source.split(/\r?\n/).some((line) => line.includes("init step") && /phase:\s*['"]done['"]/.test(line))
```
Astro-Han: **P2**: The regex `/phase:\s*['"]done['"]` requires matching quote types (the character class `['"]` matches a single character, so both opening and closing quotes must be the same type). This is correct behavior, just noting for awareness. If there's ever a malformed log with mismatched quotes like `phase: 'done"`, it won't match, which is probably the right failure mode.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,93 @@ path: pawwork-win-x64.exe
...
+[2026-04-22 21:26:18.132] [info]  unrelated message containing phase: 'done'
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include init step done")
+  })
+
+  test("reports unreadable startup log files with the path", async () => {
+    await expect(readStartupLogFile("/tmp/pawwork-missing-startup-log")).rejects.toThrow(
+      "Failed to read startup log file /tmp/pawwork-missing-startup-log",
```
Astro-Han: **P2**: `toThrow` only checks that the error message contains the substring. The actual error message format is `Failed to read startup log file ${path}: ${originalError}`. The test passes because the substring matches, but if the error format changes, the test might still pass unexpectedly. Consider asserting the full expected pattern or using a more specific regex match.

Astro-Han (commented):
## Summary

Good addition for catching startup hangs. Found some test strictness issues worth addressing.

### P1 Issues
- Test assertions check only specific failure messages but don't verify the complete failure set, which could miss unexpected failures

### P2 Issues
- Regex pattern edge case awareness
- `readStartupLogFile` test only checks error message prefix

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by making this failure test exact: it now asserts the failure list has exactly the two expected startup-completion errors.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed. I verified the previous regex actually allowed mismatched quotes, so `hasInitDone` now uses a captured quote and backreference, with a malformed quote regression test.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by adding the same exact failure-count assertion for the latest-startup-attempt case.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by tightening the unreadable-log assertion to require the path plus trailing underlying error detail.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,36 @@ The helper verifies:
 - `latest.yml` points to `pawwork-win-x64.exe`.
 - `latest-mac.yml` includes both `pawwork-mac-arm64.zip` and `pawwork-mac-x64.zip`.
 
+Also verify a fresh packaged startup before closing startup-blocking issues. The command below is for macOS; override `PAWWORK_RELEASE_APP_PATH` and `PAWWORK_RELEASE_STARTUP_LOG` if the app or log is in a custom location.
+
+```bash
+set -euo pipefail
+smoke_home=/tmp/pawwork-release-smoke/user-data
```
Astro-Han, line 143: **P1**: The cleanup trap only kills the process but doesn't remove `$smoke_home`. This directory persists after script exit. Should either add cleanup for the directory, or document that cleanup is intentionally left for debugging.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,36 @@ The helper verifies:
...
+ready_file="$smoke_home/ai.pawwork.desktop/ci-smoke-ready.json"
+app_path=${PAWWORK_RELEASE_APP_PATH:-/Applications/PawWork.app/Contents/MacOS/PawWork}
+startup_log=${PAWWORK_RELEASE_STARTUP_LOG:-$HOME/Library/Logs/PawWork/main.log}
+app_pid=""
+cleanup() {
```
Astro-Han, line 149: **P2**: The brace expansion `{1..60}` is bash-specific. This works in bash but fails in zsh/ksh/etc. Since the script lacks a shebang and uses `set -euo pipefail` (which is bash/zsh compatible), consider documenting that bash is required, or use a POSIX-compatible loop: `while [ $i -le 60 ]; do ... i=$((i+1)); done`.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,36 @@ The helper verifies:
...
+    kill "$app_pid" 2>/dev/null || true
+  fi
+}
+trap cleanup EXIT
+rm -rf "$smoke_home"
+PAWWORK_CI_SMOKE=true PAWWORK_CI_SMOKE_HOME="$smoke_home" "$app_path" &
+app_pid=$!
+for _ in {1..60}; do
+  test -f "$ready_file" && break
+  sleep 1
```
Astro-Han, line 162: **P2**: After the wait loop exits (line 20-23), the script immediately runs the verifier. If `ci-smoke-ready.json` is written but `main.log` hasn't flushed the final log entries yet, verification could fail. Consider adding a small `sleep 1` after detecting the ready file, or note this potential race.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,46 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
   }
 }
 
+function latestStartupAttempt(source: string) {
```
Astro-Han, line 107: **P1**: Error message inconsistency. This message says "Startup log does not include any app starting entry" but other messages below use "Latest startup log" prefix. Should be consistent: "Latest startup log does not include any app starting entry".

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -15,6 +15,7 @@ type VerificationInput = {
   release: GithubRelease
   latestYml?: string
   latestMacYml?: string
+  startupLog?: string
```
Astro-Han, line 18: **P2**: The `releaseVersion` function doesn't validate edge cases: empty tag, just "v", or non-semver format. Consider adding validation or at least handling `tag === "v"` which would return empty string.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,46 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
+}
+
+function hasInitDone(source: string) {
+  return source.split(/\r?\n/).some((line) => line.includes("init step") && /phase:\s*(['"])done\1/.test(line))
+}
+
+function releaseVersion(tag: string) {
+  return tag.startsWith("v") ? tag.slice(1) : tag
+}
+
```
Astro-Han, line 129: **P1**: Missing documentation in usage message. The script's usage output (line ~201 in main) doesn't mention `PAWWORK_RELEASE_STARTUP_LOG` env var. Users running the script standalone won't know about this optional feature. Consider updating usage to: `Usage: bun verify-release.ts <tag> [owner/repo] [env: PAWWORK_RELEASE_STARTUP_LOG]`.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -5,6 +5,7 @@ import {
   fetchText,
   normalizeTag,
   parseUpdaterFileUrls,
+  readStartupLogFile,
```
Astro-Han, line 12: **P2**: Missing test for empty startup log string. What happens when `startupLog: ""` is passed to `verifyReleasePayload`? Should verify this edge case returns the "no app starting" failure.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,110 @@ path: pawwork-win-x64.exe
```
Astro-Han, line 213: **P2**: Missing test for startup log with `packaged: false`. The release checklist emphasizes "fresh packaged startup" but the code doesn't verify `packaged: true`. Should document whether this is intentional, or add a check.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,110 @@ path: pawwork-win-x64.exe
...
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 20:00:00.000] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 20:00:01.000] [info]  loading task finished
+[2026-04-22 20:00:01.001] [info]  init step { step: { phase: 'done' } }
+[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:16.767] [info]  sidecar connection started { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
```
Astro-Han, line 294: **P3**: Test description "does not accept phase done outside the init step line" is fine but could be more precise: "does not accept 'phase: done' in a non-init-step log line" to clarify the exact rejection criteria.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,110 @@ path: pawwork-win-x64.exe
...
+    expect(failures).toContain("Latest startup log does not include loading task finished")
+    expect(failures).toContain("Latest startup log does not include init step done")
+    expect(failures).toHaveLength(2)
+  })
+
+  test("reports release version mismatches in the startup log", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
```
Astro-Han, line 307: **P3**: The `readStartupLogFile` test only covers the error case (file not found). Consider adding a happy-path test with a temporary file containing valid log content to verify `Bun.file().text()` integration.

Astro-Han (commented):
## Review Summary

Overall the PR is well-structured with good test coverage. Found several issues.

**P1 (suggest fixing):** 3 issues
**P2 (record only):** 7 issues
**P3 (nitpick):** 2 issues

See inline comments for details.

Astro-Han (commented):
## Review Summary

Overall the PR is well-structured with good test coverage. Found several issues.

**P1 (suggest fixing):** 3 issues
**P2 (record only):** 7 issues
**P3 (nitpick):** 2 issues

See inline comments for details.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,36 @@ The helper verifies:
 - `latest.yml` points to `pawwork-win-x64.exe`.
 - `latest-mac.yml` includes both `pawwork-mac-arm64.zip` and `pawwork-mac-x64.zip`.
 
+Also verify a fresh packaged startup before closing startup-blocking issues. The command below is for macOS; override `PAWWORK_RELEASE_APP_PATH` and `PAWWORK_RELEASE_STARTUP_LOG` if the app or log is in a custom location.
+
+```bash
+set -euo pipefail
+smoke_home=/tmp/pawwork-release-smoke/user-data
+ready_file="$smoke_home/ai.pawwork.desktop/ci-smoke-ready.json"
+app_path=${PAWWORK_RELEASE_APP_PATH:-/Applications/PawWork.app/Contents/MacOS/PawWork}
```
Astro-Han, line 146: **P1**: The cleanup trap only kills the process but doesn't remove `$smoke_home`. This directory persists after script exit. Should either add cleanup for the directory, or document that cleanup is intentionally left for debugging.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,36 @@ The helper verifies:
...
+startup_log=${PAWWORK_RELEASE_STARTUP_LOG:-$HOME/Library/Logs/PawWork/main.log}
+app_pid=""
+cleanup() {
+  if [ -n "$app_pid" ]; then
+    kill "$app_pid" 2>/dev/null || true
+  fi
+}
```
Astro-Han, line 154: **P2**: The brace expansion `{1..60}` is bash-specific. This works in bash but fails in zsh/ksh/etc. Since the script lacks a shebang and uses `set -euo pipefail`, consider documenting that bash is required, or use a POSIX-compatible loop.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,36 @@ The helper verifies:
...
+trap cleanup EXIT
+rm -rf "$smoke_home"
+PAWWORK_CI_SMOKE=true PAWWORK_CI_SMOKE_HOME="$smoke_home" "$app_path" &
+app_pid=$!
+for _ in {1..60}; do
+  test -f "$ready_file" && break
+  sleep 1
+done
```
Astro-Han, line 164: **P2**: After the wait loop exits, the script immediately runs the verifier. If `ci-smoke-ready.json` is written but `main.log` hasn't flushed the final log entries yet, verification could fail. Consider adding a small delay or note this potential race.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,46 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
+function latestStartupAttempt(source: string) {
+  const marker = "app starting"
+  const index = source.lastIndexOf(marker)
+  if (index === -1) return undefined
+  return source.slice(index)
+}
+
+function hasInitDone(source: string) {
+  return source.split(/\r?\n/).some((line) => line.includes("init step") && /phase:\s*(['"])done\1/.test(line))
+}
```
Astro-Han, line 124: **P1**: Error message inconsistency. This message says "Startup log does not include any app starting entry" but other messages use "Latest startup log" prefix. Should be consistent: "Latest startup log does not include any app starting entry".

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,46 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
   }
 }
 
+function latestStartupAttempt(source: string) {
+  const marker = "app starting"
```
Astro-Han, line 108: **P2**: The `releaseVersion` function doesn't validate edge cases: empty tag, just "v", or non-semver format. Consider adding validation or handling edge cases.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -212,7 +264,9 @@ async function main() {
     )
     const latestYml = await fetchAssetText(release, "latest.yml")
     const latestMacYml = await fetchAssetText(release, "latest-mac.yml")
-    const failures = verifyReleasePayload({ release, latestYml, latestMacYml })
+    const startupLogPath = process.env.PAWWORK_RELEASE_STARTUP_LOG
```
Astro-Han, line 292: **P1**: Missing documentation in usage message. The script's usage output doesn't mention `PAWWORK_RELEASE_STARTUP_LOG` env var. Users running the script standalone won't know about this optional feature.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,110 @@ path: pawwork-win-x64.exe
     expect(failures).toContain("latest-mac.yml does not include pawwork-mac-x64.zip")
   })
 
+  test("accepts a complete startup log for the release version", () => {
```
Astro-Han, line 225: **P2**: Missing test for empty startup log string. What happens when `startupLog: ""` is passed? Should verify this edge case returns the "no app starting" failure.

Astro-Han, line 225: **P2**: Missing test for startup log with `packaged: false`. The release checklist emphasizes "fresh packaged startup" but the code doesn't verify `packaged: true`. Should document whether this is intentional.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,110 @@ path: pawwork-win-x64.exe
...
+
+  test("does not accept phase done outside the init step line", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
+[2026-04-22 21:26:18.130] [info]  loading task finished
+[2026-04-22 21:26:18.131] [info]  init step { step: { phase: 'loading' } }
```
Astro-Han, line 369: **P3**: Test description "does not accept phase done outside the init step line" could be more precise: "does not accept 'phase: done' in a non-init-step log line".

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,110 @@ path: pawwork-win-x64.exe
...
+    expect(failures).toContain("Latest startup log does not include init step done")
+  })
+
+  test("does not accept malformed init step done quotes", () => {
+    const failures = verifyReleasePayload({
+      release: baseRelease,
+      latestYml: "files:\n  - url: pawwork-win-x64.exe\n",
+      latestMacYml: "files:\n  - url: pawwork-mac-arm64.zip\n  - url: pawwork-mac-x64.zip\n",
+      startupLog: `[2026-04-22 21:26:16.088] [info]  app starting { version: '0.2.6', packaged: true }
+[2026-04-22 21:26:18.129] [info]  server ready { url: 'http://127.0.0.1:59635' }
```
Astro-Han, line 383: **P3**: The `readStartupLogFile` test only covers the error case. Consider adding a happy-path test with a temporary file.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by cleaning up the isolated smoke user-data directory in the trap after terminating the app process.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by cleaning up the isolated smoke user-data directory in the trap after terminating the app process.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by replacing the brace expansion with a portable counter loop.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by replacing the brace expansion with a portable counter loop.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by adding a short delay after the ready marker before running the verifier, so the final log entries can flush.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by adding a short delay after the ready marker before running the verifier, so the final log entries can flush.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by changing the missing-startup message to use the same Latest startup log prefix.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by changing the missing-startup message to use the same Latest startup log prefix.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by routing releaseVersion through normalizeTag, with coverage for empty tags and v-only tags.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by routing releaseVersion through normalizeTag, with coverage for empty tags and v-only tags.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by documenting PAWWORK_RELEASE_STARTUP_LOG in the script usage output.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by documenting PAWWORK_RELEASE_STARTUP_LOG in the script usage output.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed with an empty startup log regression test that expects the no app starting failure.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed with an empty startup log regression test that expects the no app starting failure.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by making packaged: true part of startup verification, with a packaged: false regression test.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by making packaged: true part of startup verification, with a packaged: false regression test.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by making the test name explicit about rejecting phase: done in a non-init-step log line.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by making the test name explicit about rejecting phase: done in a non-init-step log line.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by adding a temp-file happy-path test for readStartupLogFile.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed by adding a temp-file happy-path test for readStartupLogFile.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,36 @@ The helper verifies:
...
 - `latest-mac.yml` includes both `pawwork-mac-arm64.zip` and `pawwork-mac-x64.zip`.
 
+Also verify a fresh packaged startup before closing startup-blocking issues. The command below is for macOS; override `PAWWORK_RELEASE_APP_PATH` and `PAWWORK_RELEASE_STARTUP_LOG` if the app or log is in a custom location.
+
+```bash
+set -euo pipefail
+smoke_home=/tmp/pawwork-release-smoke/user-data
+ready_file="$smoke_home/ai.pawwork.desktop/ci-smoke-ready.json"
+app_path=${PAWWORK_RELEASE_APP_PATH:-/Applications/PawWork.app/Contents/MacOS/PawWork}
+startup_log=${PAWWORK_RELEASE_STARTUP_LOG:-$HOME/Library/Logs/PawWork/main.log}
```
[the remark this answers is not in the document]
Astro-Han: Fixed by making the default startup_log point at the isolated smoke user data logs directory instead of the normal user profile log.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -214,6 +215,110 @@ path: pawwork-win-x64.exe
...
+`,
+    })
+
+    expect(failures).toContain("Latest startup log does not include init step done")
+  })
+
+  test("reports unreadable startup log files with the path", async () => {
+    await expect(readStartupLogFile("/tmp/pawwork-missing-startup-log")).rejects.toThrow(
+      /^Failed to read startup log file \/tmp\/pawwork-missing-startup-log: .+/,
+    )
```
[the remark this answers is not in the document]
Astro-Han: Fixed by generating a unique missing file path under a temporary directory and escaping it in the expected error regex.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,41 @@ The helper verifies:
...
 
+Also verify a fresh packaged startup before closing startup-blocking issues. The command below is for macOS; override `PAWWORK_RELEASE_APP_PATH` and `PAWWORK_RELEASE_STARTUP_LOG` if the app or log is in a custom location.
+
+```bash
+set -euo pipefail
+smoke_home=/tmp/pawwork-release-smoke/user-data
+smoke_user_data="$smoke_home/ai.pawwork.desktop"
+ready_file="$smoke_user_data/ci-smoke-ready.json"
+app_path=${PAWWORK_RELEASE_APP_PATH:-/Applications/PawWork.app/Contents/MacOS/PawWork}
+startup_log=${PAWWORK_RELEASE_STARTUP_LOG:-$smoke_user_data/logs/main.log}
```
Astro-Han, line 147: **P1** — The default `$smoke_user_data/logs/main.log` does not exist on macOS. electron-log v5 resolves the log file via `app.getPath('logs')`, which on macOS defaults to `~/Library/Logs/<AppName>/main.log` and is not derived from `appData`. `PAWWORK_CI_SMOKE_HOME` overrides `appData`/`userData` but not `logs`, and `packages/desktop-electron/src/main/index.ts` never calls `app.setPath('logs', ...)`. Confirmed on a developer machine the packaged app writes to `~/Library/Logs/PawWork/main.log` (not under Application Support). As written, the documented macOS smoke will fail with `Failed to read startup log file` unless the operator also sets `PAWWORK_RELEASE_STARTUP_LOG`. Either (a) change the fallback to `$HOME/Library/Logs/PawWork/main.log`, or (b) call `app.setPath('logs', join(CI_SMOKE_HOME, 'logs'))` under smoke mode so the documented fallback is actually correct.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,58 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
+  } catch (error) {
+    failures.push(error instanceof Error ? error.message : String(error))
+    return failures
+  }
+
+  if (!hasStartupVersion(latest, expectedVersion)) {
+    failures.push(`Latest startup log version does not match expected ${expectedVersion}`)
+  }
+  if (!hasPackagedStartup(latest)) failures.push("Latest startup log does not include packaged true")
+  if (!latest.includes("server ready")) failures.push("Latest startup log does not include server ready")
```
Astro-Han: **P2** — Substring alias: `packages/desktop-electron/src/main/index.ts:443` emits `logger.log("awaiting server ready")` just before line 445's real `logger.log("server ready", { url: res.url })`. A fresh startup that reaches the `await server-ready` barrier but never resolves will still have `awaiting server ready` in the log, and `latest.includes("server ready")` matches that substring. The check therefore passes even in the exact hang mode this PR is meant to catch (sidecar reachable-ish but `serverReady.resolve(...)` / `server ready` event never fires). Tighten to something like `latest.includes("] server ready")` or `/\bserver ready\b[^\n]*\{\s*url:/` — and add a test for a log that contains `awaiting server ready` only. (The existing "stuck after sidecar readiness" test includes the real `server ready` line, so it does not exercise this path.)

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,58 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
 
+function latestStartupAttempt(source: string) {
+  const marker = "app starting"
+  const index = source.lastIndexOf(marker)
+  if (index === -1) return undefined
+  return source.slice(index)
+}
+
+function hasInitDone(source: string) {
+  return source.split(/\r?\n/).some((line) => line.includes("init step") && /phase:\s*(['"])done\1/.test(line))
```
Astro-Han: **P2** — `phase:\s*(['"])done\1` is coupled to electron-log's current object-inspect serialization of `{ step: { phase: "done" } }` (lowercase unquoted keys, single-quoted strings). If electron-log is ever upgraded to one that emits JSON, or someone turns on `log.transports.file.format = '...json...'`, the regex silently stops matching and the final init-step check becomes a no-op without any runtime change — the exact class of silent regression this PR exists to prevent. Consider either (a) adding a unit test at the emit site (`setInitStep({ phase: "done" })`) that asserts the rendered log line matches this regex, or (b) emitting a dedicated marker line (e.g. `logger.log("init done")`) owned by this check so the release verifier is not coupled to inspect-format details.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,58 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
+}
+
+function releaseVersion(tag: string) {
+  return normalizeTag(tag).slice(1)
+}
+
+function hasStartupVersion(source: string, expectedVersion: string) {
+  const escaped = expectedVersion.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")
+  return new RegExp(`app starting[^\\n]*version:\\s*['"]${escaped}['"]`).test(source)
```
Astro-Han: **P3** — `latest` is already sliced by `latestStartupAttempt` to start at the last `app starting` marker, so re-anchoring with `app starting[^\n]*` here is redundant. A single-line check like `new RegExp(\`^[^\\n]*version:\\s*['\"]${escaped}['\"]\`).test(latest)` (or just `.split("\n", 1)[0]` then match) expresses the same intent more directly, and removes the need for `hasPackagedStartup` to re-anchor too.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,41 @@ The helper verifies:
...
+  i=$((i + 1))
+done
+test -f "$ready_file"
+sleep 1
+PAWWORK_RELEASE_STARTUP_LOG="$startup_log" bun packages/desktop-electron/scripts/verify-release.ts vX.Y.Z
+```
+
+The startup log check reads the latest `app starting` block and verifies it reaches `server ready`, `loading task finished`, and `init step done`. This catches first-launch hangs where the sidecar becomes reachable but the desktop shell never opens the main window.
+
+For Windows releases, run an equivalent fresh-user-data launch of the installed `.exe`, then run the same verifier with `PAWWORK_RELEASE_STARTUP_LOG` pointing at that install's `main.log`.
```
Astro-Han: **P3** — "Run an equivalent fresh-user-data launch of the installed `.exe`" hand-waves the whole Windows path. AGENTS.md treats macOS and Windows both as first-class, and Windows has its own first-launch quirks (roaming `%APPDATA%` layout, `latest.yml` vs `latest-mac.yml` asymmetry, AV scan latency before the installed `.exe` can even spawn). Worth adding a mirrored PowerShell snippet (delete `%APPDATA%\PawWork`, set `PAWWORK_CI_SMOKE=true` + `PAWWORK_CI_SMOKE_HOME`, launch the installed `.exe`, poll for `ci-smoke-ready.json`, run the verifier pointed at `%APPDATA%\PawWork\logs\main.log`) so Windows release gates are not operator-specific tribal knowledge. Otherwise #144/#145-class regressions on Windows-only can still ship.

Astro-Han (commented):
Release verification hardening is a worthwhile goal, and the test matrix is thorough. Surfacing a few issues — one **P1** where the documented macOS smoke script will not find the log file as written, one **P2** alias bug in the `server ready` check that defeats the stated hang detection goal, plus smaller polish items. See inline.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,58 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
...
+    failures.push("Latest startup log does not include any app starting entry")
+    return failures
+  }
+
+  let expectedVersion: string
+  try {
+    expectedVersion = releaseVersion(expectedTag)
+  } catch (error) {
+    failures.push(error instanceof Error ? error.message : String(error))
+    return failures
```
Astro-Han: **P3** — When `expectedTag` fails `normalizeTag`, this early-returns with only the tag error. If the same log is also missing `server ready` / `init step done`, the operator gets a single misleading line and has to re-run to see the rest. Prefer recording the tag error and continuing with the remaining checks (or skipping only the version match) so one run surfaces the full failure set — consistent with how `verifyReleasePayload` accumulates failures rather than short-circuiting.

Astro-Han (commented):
## Review Summary

**P0 (Blocking)**: None - code logic is sound

**P1 (Should Fix)**: 1 issue
- `escapeRegExp` helper defined only in test file, but `hasStartupVersion` uses the same logic inline. Consider consolidating.

**P2 (Nice to Fix)**: 2 issues  
- RELEASE_CHECKLIST.md: script fails silently after 60s timeout without explicit `Timed out waiting for ready file` message
- cleanup() does not verify if `app_pid` process still exists before `kill`

**P3 (Nit)**: 3 issues
- `[^\n]` in regex literals - minor readability concern  
- No test for version strings containing regex special chars
- Error message prefix inconsistency (`Invalid release tag` vs `Latest startup log does not include`)

Overall: good coverage, well-tested. P1 is a minor code organization issue, P2s are documentation/script robustness nits.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -1,16 +1,25 @@
 import { afterEach, describe, expect, test } from "bun:test"
+import { mkdtemp, rm, writeFile } from "node:fs/promises"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
```
Astro-Han, line 5: **P1**: `escapeRegExp` is defined here but `hasStartupVersion` (verify-release.ts:125) implements the same regex escaping inline. Consider exporting this helper or consolidating to avoid duplication.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -103,6 +104,58 @@ function verifyReferencedAssets(sourceName: string, urls: string[], assetNames:
   }
 }
 
+function latestStartupAttempt(source: string) {
+  const marker = "app starting"
+  const index = source.lastIndexOf(marker)
+  if (index === -1) return undefined
```
Astro-Han, line 110: **P2**: `[^\n]` in regex literal - while valid, this pattern relies on character class escaping. Consider using `[^\r\n]` or documenting the convention for consistency with `hasInitDone` which uses `/\r?\n/`.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,41 @@ The helper verifies:
...
+}
+trap cleanup EXIT
+rm -rf "$smoke_home"
+PAWWORK_CI_SMOKE=true PAWWORK_CI_SMOKE_HOME="$smoke_home" "$app_path" &
+app_pid=$!
+i=0
+while [ "$i" -lt 60 ]; do
+  test -f "$ready_file" && break
+  sleep 1
+  i=$((i + 1))
```
Astro-Han, line 163: **P2**: After 60-second timeout, if `test -f "$ready_file"` fails, the script exits with non-zero but no explicit timeout message. Consider adding: `if [ "$i" -ge 60 ]; then echo "Timed out waiting for ci-smoke-ready.json"; exit 1; fi` before the final `test`.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by setting the Electron logs path under the isolated smoke userData directory when PAWWORK_CI_SMOKE_HOME is set. The documented default startup_log now points at the path the packaged smoke app actually writes.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by requiring a real server ready log line with the url object instead of a raw substring match. Added a regression test where the log only contains awaiting server ready.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by emitting a dedicated init done marker when initialization reaches done, and by making the verifier require that marker instead of parsing object-inspect formatting.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by matching version and packaged on the latest startup line directly, instead of re-anchoring each regex on app starting.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by adding a mirrored PowerShell fresh-user-data smoke snippet for Windows releases, using the isolated smoke profile and startup log path.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by recording invalid tag errors without returning early, so the verifier still reports the remaining startup-log failures from the same run.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Consolidated by exporting the production escapeRegExp helper and importing it in the test. I classify this as P3 rather than P1 because it was duplication, not a release-verification behavior break.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed by removing the re-anchored multi-line regexes that used [^\\n]; startup metadata is now checked on the first startup line.

.github/RELEASE_CHECKLIST.md
Astro-Han: Fixed by replacing the final bare test with an explicit timeout message before exiting.

.github/RELEASE_CHECKLIST.md
```diff
@@ -136,6 +136,74 @@ The helper verifies:
...
+    }
+    Start-Sleep -Seconds 1
+  }
+  if (-not $ready) { throw "Timed out waiting for $readyFile" }
+  Start-Sleep -Seconds 1
+  $env:PAWWORK_RELEASE_STARTUP_LOG = $startupLog
+  bun packages/desktop-electron/scripts/verify-release.ts vX.Y.Z
+} finally {
+  if ($app -and -not $app.HasExited) { Stop-Process -Id $app.Id -Force }
+  Remove-Item -Recurse -Force $smokeHome -ErrorAction SilentlyContinue
```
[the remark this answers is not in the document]
Astro-Han: Fixed by saving the previous PowerShell env var values before setting smoke-mode variables and restoring or removing them in finally before deleting the smoke profile.

packages/desktop-electron/src/main/index-sidecar-source.test.ts
```diff
@@ -6,6 +6,8 @@ describe("desktop sidecar source guard", () => {
 
     expect(source).toContain("username: PAWWORK_RUNTIME.serverUsername")
     expect(source).toContain("const needsMigration = false")
+    expect(source).toContain('app.setPath("logs", join(app.getPath("userData"), "logs"))')
+    expect(source).toContain('logger.log("init done")')
```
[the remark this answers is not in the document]
Astro-Han: Fixed by adding the raw-source guard for logger.log("server ready", { url: res.url }) next to the existing smoke log and init-done contract checks. I classify this as P3 test hardening rather than Major because it guards a future refactor, not a current behavior break.

Issue #141, opened 2026-04-22
Improve Windows compatibility and restore Windows unit signal


Astro-Han:
## Problem

PR #139 split Windows unit coverage by package and kept the Windows jobs as diagnostic signal instead of required gates. The PR checks are green now, but the diagnostics still show underlying Windows-specific compatibility work remains. `unit-windows-app` can pass, while `unit-windows-desktop` and `unit-windows-opencode` surface path, shell, dynamic import, runtime, and test fixture issues.

This matters for more than CI cleanliness. Some failures touch desktop runtime behavior, including user data paths, server runtime setup, and shell environment loading. Those can affect packaged Windows usage even if the release build itself completes.

## Evidence

From the Windows diagnostics in PR #139:

- `unit-windows-app` passed.
- `unit-windows-desktop` reported path normalization and desktop server runtime failures, plus a Windows import error around `shell-env.ts`.
- `unit-windows-opencode` reported broad Windows incompatibilities around `/tmp`, backslash paths, permissions, shell behavior, plugins, worktrees, providers, and tooling.

## Upstream reference

Upstream `anomalyco/opencode` currently runs full Windows unit coverage in its `test.yml` matrix, and recent upstream `unit (windows)` runs are green. Upstream has fixed similar issues through targeted Windows compatibility patches, including:

- path separator and glob normalization
- `os.tmpdir()` instead of hardcoded `/tmp`
- CRLF handling
- NTFS mtime tolerance
- Windows-safe local dynamic imports
- Windows dependency install behavior
- file watcher and CI runner adjustments

## Goal

Bring the relevant upstream Windows fixes into PawWork where they apply, then fix the PawWork-specific desktop Windows failures. The desired outcome is not just a green CI page, but Windows diagnostics that are reliable enough to promote back toward required CI signal.

## Scope

- Cherry-pick or manually port small upstream Windows compatibility fixes that apply cleanly to PawWork.
- Preserve PawWork-specific runtime and desktop behavior.
- Fix PawWork desktop Windows runtime issues exposed by `unit-windows-desktop`.
- Keep CI changes minimal until the underlying Windows failures are understood.

## Non-goals

- Do not do a broad upstream merge.
- Do not hide failures by skipping large groups of Windows tests.
- Do not redesign the release workflow as part of this issue.

## Follow-up

[...]

PR #148, opened 2026-04-22
fix(opencode): improve Windows unit compatibility

Astro-Han:
## Summary

- Normalize runtime namespace tests for Windows path separators.
- Port the narrow Bun Windows tar workaround for npm and Arborist installs.
- Update managed ripgrep lookup and version for Windows, including arm64 and ia32 mappings.
- Stabilize plugin workspace retry, worktree cleanup, shell queue timing, and cross-spawn stream tests on Windows.

## Why

Issue #141's PR direction 1 targets the current `unit-windows-opencode` failure families from run 24781051308: runtime namespace path assertions, config dependency installs, plugin workspace timing and cleanup, and shell queue timing. Electron desktop Windows hardening remains a separate PR direction.

## Related Issue

Part of #141.

## How To Verify

```bash
cd packages/opencode
bun test test/global/runtime-namespace.test.ts
bun test test/effect/cross-spawn-spawner.test.ts
bun test test/config/config.test.ts test/config/seed-e2e.test.ts test/tool/registry.test.ts
bun test test/plugin/workspace-adaptor.test.ts
bun test test/session/prompt-effect.test.ts -t "loop waits while shell runs and starts after shell exits"
bun test test/session/prompt-effect.test.ts -t "shell completion resumes queued loop callers"
bun test test/file/ripgrep.test.ts
bun run test:ci
```

Latest local `bun run test:ci` result: 2016 pass, 8 skip, 1 todo, 0 fail.

## Screenshots or Recordings

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added ia32-Windows platform support and updated ripgrep to v15.1.0

* **Bug Fixes**
  * Windows runtime compatibility workaround to avoid tar-related issues
  * More reliable directory cleanup on Windows via increased retry handling

* **Tests**
  * Improved test stability: concurrency handling, dynamic temp directories, polling helper for flaky timing, and increased timeouts for shell-related tests

### DIFF
--- a/packages/opencode/src/file/ripgrep.ts
+++ b/packages/opencode/src/file/ripgrep.ts
@@ -118,6 +118,7 @@ export namespace Ripgrep {
     "x64-darwin": { platform: "x86_64-apple-darwin", extension: "tar.gz" },
     "x64-linux": { platform: "x86_64-unknown-linux-musl", extension: "tar.gz" },
     "arm64-win32": { platform: "aarch64-pc-windows-msvc", extension: "zip" },
+    "ia32-win32": { platform: "i686-pc-windows-msvc", extension: "zip" },
     "x64-win32": { platform: "x86_64-pc-windows-msvc", extension: "zip" },
   } as const
 
@@ -158,7 +159,7 @@ export namespace Ripgrep {
       const config = PLATFORM[platformKey]
       if (!config) throw new UnsupportedPlatformError({ platform: platformKey })
 
-      const version = "14.1.1"
+      const version = "15.1.0"
       const filename = `ripgrep-${version}-${config.platform}.${config.extension}`
       const url = `https://github.com/BurntSushi/ripgrep/releases/download/${version}/${filename}`
 
--- a/packages/opencode/src/npm/index.ts
+++ b/packages/opencode/src/npm/index.ts
@@ -7,7 +7,6 @@ import path from "path"
 import { readdir, rm } from "fs/promises"
 import { Filesystem } from "@/util/filesystem"
 import { Flock } from "@/util/flock"
-import { Arborist } from "@npmcli/arborist"
 
 export namespace Npm {
   const log = Log.create({ service: "npm" })
@@ -27,6 +26,16 @@ export namespace Npm {
 
   function directory(pkg: string) {
     return path.join(Global.Path.cache, "packages", sanitize(pkg))
+  }
+
+  async function loadArborist() {
+    if (process.platform === "win32") {
+      // Bun on Windows does not support the UV_FS_O_FILEMAP flag used by tar
+      // for small files. tar snapshots this env var during module init, so set
+      // it immediately before Arborist can import tar.
+      process.env.__FAKE_PLATFORM__ = "linux"
+    }
+    return import("@npmcli/arborist")
   }
 
   function resolveEntryPoint(name: string, dir: string) {
@@ -68,6 +77,7 @@ export namespace Npm {
       pkg,
     })
 
+    const { Arborist } = await loadArborist()
     const arborist = new Arborist({
       path: dir,
       binLinks: true,
@@ -108,6 +118,7 @@ export namespace Npm {
     log.info("checking dependencies", { dir })
 
     const reify = async () => {
+      const { Arborist } = await loadArborist()
       const arb = new Arborist({
         path: dir,
         binLinks: true,
--- a/packages/opencode/src/worktree/index.ts
+++ b/packages/opencode/src/worktree/index.ts
@@ -358,14 +358,22 @@ export namespace Worktree {
       }
 
       function cleanDirectory(target: string) {
-        return Effect.promise(() =>
-          import("fs/promises")
-            .then((fsp) => fsp.rm(target, { recursive: true, force: true, maxRetries: 5, retryDelay: 100 }))
-            .catch((error) => {
+        // Match test preload cleanup: Windows can keep git and sqlite handles
+        // alive briefly after process teardown, so EBUSY needs a longer budget.
+        const maxRetries = process.platform === "win32" ? 30 : 5
+        const remove = (remaining: number): Effect.Effect<void> =>
+          fs.remove(target, { recursive: true, force: true }).pipe(
+            Effect.catch((error) => {
+              if (remaining > 0) {
+                return Effect.sleep(100).pipe(Effect.flatMap(() => remove(remaining - 1)))
+              }
               const message = errorMessage(error)
-              throw new RemoveFailedError({ message: message || "Failed to remove git worktree directory" })
+              return Effect.sync(() => {
+                throw new RemoveFailedError({ message: message || "Failed to remove git worktree directory" })
+              })
             }),
-        )
+          )
+        return remove(maxRetries)
       }
 
       const remove = Effect.fn("Worktree.remove")(function* (input: RemoveInput) {
--- a/packages/opencode/test/effect/cross-spawn-spawner.test.ts
+++ b/packages/opencode/test/effect/cross-spawn-spawner.test.ts
@@ -169,7 +169,12 @@ describe("cross-spawn spawner", () => {
             'process.stderr.write("stderr\\n", done)',
           ].join("\n"),
         )
-        const [stdout, stderr] = yield* Effect.all([decodeByteStream(handle.stdout), decodeByteStream(handle.stderr)])
+        // Drain both pipes in parallel so Windows CI does not stall when stderr
+        // waits behind stdout decoding.
+        const [stdout, stderr] = yield* Effect.all(
+          [decodeByteStream(handle.stdout), decodeByteStream(handle.stderr)],
+          { concurrency: 2 },
+        )
         expect(stdout).toBe("stdout")
         expect(stderr).toBe("stderr")
       }),
--- a/packages/opencode/test/global/runtime-namespace.test.ts
+++ b/packages/opencode/test/global/runtime-namespace.test.ts
@@ -1,14 +1,33 @@
 import { describe, expect, test } from "bun:test"
 import path from "path"
+import { tmpdir } from "../fixture/fixture"
 
-function readGlobalPath(namespace?: string) {
+type Roots = {
+  data: string
+  cache: string
+  config: string
+  state: string
+}
+
+function rootsFor(dir: string): Roots {
+  return {
+    data: path.join(dir, "share"),
+    cache: path.join(dir, "cache"),
+    config: path.join(dir, "config"),
+    state: path.join(dir, "state"),
+  }
+}
+
+function readGlobalPath(roots: Roots, namespace?: string) {
   const script = `
-    process.env.XDG_DATA_HOME = "/tmp/pawwork-runtime-test/share"
-    process.env.XDG_CACHE_HOME = "/tmp/pawwork-runtime-test/cache"
-    process.env.XDG_CONFIG_HOME = "/tmp/pawwork-runtime-test/config"
-    process.env.XDG_STATE_HOME = "/tmp/pawwork-runtime-test/state"
+    process.env.XDG_DATA_HOME = ${JSON.stringify(roots.data)}
+    process.env.XDG_CACHE_HOME = ${JSON.stringify(roots.cache)}
+    process.env.XDG_CONFIG_HOME = ${JSON.stringify(roots.config)}
+    process.env.XDG_STATE_HOME = ${JSON.stringify(roots.state)}
     if (${JSON.stringify(namespace)} !== undefined) {
       process.env.PAWWORK_RUNTIME_NAMESPACE = ${JSON.stringify(namespace)}
+    } else {
+      delete process.env.PAWWORK_RUNTIME_NAMESPACE
     }
     const { Global } = await import("./src/global/index.ts")
     console.log(JSON.stringify(Global.Path))
@@ -25,24 +44,29 @@ function readGlobalPath(namespace?: string) {
   return JSON.parse(Buffer.from(result.stdout).toString()) as Record<string, string>
 }
 
-describe("Global runtime namespace", () => {
-  test("defaults to OpenCode namespace outside PawWork desktop", () => {
-    const paths = readGlobalPath()
+function expectCoreRoots(paths: Record<string, string>, roots: Roots, namespace: string) {
+  expect(paths.data).toBe(path.join(roots.data, namespace))
+  expect(paths.cache).toBe(path.join(roots.cache, namespace))
+  expect(paths.config).toBe(path.join(roots.config, namespace))
+  expect(paths.state).toBe(path.join(roots.state, namespace))
+}
 
-    expect(paths.data).toBe("/tmp/pawwork-runtime-test/share/opencode")
-    expect(paths.cache).toBe("/tmp/pawwork-runtime-test/cache/opencode")
-    expect(paths.config).toBe("/tmp/pawwork-runtime-test/config/opencode")
-    expect(paths.state).toBe("/tmp/pawwork-runtime-test/state/opencode")
+describe("Global runtime namespace", () => {
+  test("defaults to OpenCode namespace outside PawWork desktop", async () => {
+    await using tmp = await tmpdir()
+    const roots = rootsFor(tmp.path)
+    const paths = readGlobalPath(roots)
+
+    expectCoreRoots(paths, roots, "opencode")
   })
 
-  test("uses PawWork namespace when enabled", () => {
-    const paths = readGlobalPath("pawwork")
+  test("uses PawWork namespace when enabled", async () => {
+    await using tmp = await tmpdir()
+    const roots = rootsFor(tmp.path)
+    const paths = readGlobalPath(roots, "pawwork")
 
-    expect(paths.data).toBe("/tmp/pawwork-runtime-test/share/pawwork")
-    expect(paths.cache).toBe("/tmp/pawwork-runtime-test/cache/pawwork")
-    expect(paths.config).toBe("/tmp/pawwork-runtime-test/config/pawwork")
-    expect(paths.state).toBe("/tmp/pawwork-runtime-test/state/pawwork")
-    expect(paths.bin).toBe("/tmp/pawwork-runtime-test/cache/pawwork/bin")
-    expect(paths.log).toBe("/tmp/pawwork-runtime-test/share/pawwork/log")
+    expectCoreRoots(paths, roots, "pawwork")
+    expect(paths.bin).toBe(path.join(roots.cache, "pawwork", "bin"))
+    expect(paths.log).toBe(path.join(roots.data, "pawwork", "log"))
   })
 })
--- a/packages/opencode/test/plugin/workspace-adaptor.test.ts
+++ b/packages/opencode/test/plugin/workspace-adaptor.test.ts
@@ -45,6 +45,28 @@ async function waitFor(fn: () => boolean, timeout = 5_000) {
   throw new Error("timed out waiting for workspace status")
 }
 
+async function waitForCounter(file: string, min: number) {
+  const read = async () => {
+    const text = await Bun.file(file)
+      .text()
+      .catch((error) => {
+        if (typeof error === "object" && error !== null && "code" in error && error.code === "ENOENT") return "0"
+        throw error
+      })
+    const value = Number(text)
+    if (!Number.isFinite(value)) throw new Error(`invalid retry counter value: ${text}`)
+    return value
+  }
+  const end = Date.now() + 5_000
+  let value = 0
+  while (Date.now() < end) {
+    value = await read()
+    if (value > min) return value
+    await wait(50)
+  }
+  throw new Error(`timed out waiting for counter ${file} to exceed ${min}; last value=${value}`)
+}
+
 async function pluginProject() {
   return tmpdir({
     git: true,
@@ -362,13 +384,11 @@ describe("plugin.workspace", () => {
     await Instance.disposeAll()
 
     await Workspace.get(workspace.id)
-    await wait(100)
-    const first = Number(await Bun.file(tmp.extra.counter).text())
+    const first = await waitForCounter(tmp.extra.counter, 0)
     expect(first).toBeGreaterThan(0)
 
     await Workspace.get(workspace.id)
-    await wait(100)
-    const second = Number(await Bun.file(tmp.extra.counter).text())
+    const second = await waitForCounter(tmp.extra.counter, first)
     expect(second).toBeGreaterThan(first)
 
     const status = Workspace.status().find((item) => item.workspaceID === workspace.id)
--- a/packages/opencode/test/session/prompt-effect.test.ts
+++ b/packages/opencode/test/session/prompt-effect.test.ts
@@ -1236,6 +1236,8 @@ unix(
   30_000,
 )
 
+const shellQueueTimeout = process.platform === "win32" ? 10_000 : 3_000
+
 it.live(
   "loop waits while shell runs and starts after shell exits",
   () =>
@@ -1271,7 +1273,7 @@ it.live(
       }),
       { git: true, config: providerCfg },
     ),
-  3_000,
+  shellQueueTimeout,
 )
 
 it.live(
@@ -1311,7 +1313,7 @@ it.live(
       }),
       { git: true, config: providerCfg },
     ),
-  3_000,
+  shellQueueTimeout,
 )
 
 unix(


packages/opencode/src/file/ripgrep.ts
```diff
@@ -145,7 +146,7 @@ export namespace Ripgrep {
   )
 
   const state = lazy(async () => {
-    const system = which("rg")
+    const system = which(process.platform === "win32" ? "rg.exe" : "rg")
```
Astro-Han: P1: This change adds `.exe` suffix when calling `which()` for system ripgrep lookup on Windows. However, the fallback path construction elsewhere in this file uses a similar pattern. Please verify these are consistent.

**Questions:**
1. Does Bun's `which()` automatically resolve `.exe` suffixes on Windows?
2. If no, was the previous code buggy and is this fix correct?

packages/opencode/src/npm/index.ts
```diff
@@ -1,3 +1,10 @@
+// Bun on Windows does not support the UV_FS_O_FILEMAP flag used by tar for
```
Astro-Han: P1: Setting `process.env.__FAKE_PLATFORM__` at module load time is a global side effect that affects the entire process.

**Concerns:**
1. **Scope**: This affects all code running after this module loads. Other code depending on `process.platform` could break.
2. **Cleanup**: No mechanism restores the environment after Arborist operations.
3. **Alternative**: Consider wrapping this in a function that temporarily sets the env var during Arborist operations, then restores it. Or document this as a known compromise with a TODO.

packages/opencode/src/worktree/index.ts
```diff
@@ -360,7 +360,14 @@ export namespace Worktree {
       function cleanDirectory(target: string) {
         return Effect.promise(() =>
           import("fs/promises")
-            .then((fsp) => fsp.rm(target, { recursive: true, force: true, maxRetries: 5, retryDelay: 100 }))
+            .then((fsp) =>
+              fsp.rm(target, {
+                recursive: true,
+                force: true,
+                maxRetries: process.platform === "win32" ? 30 : 5,
```
Astro-Han: P2: The `maxRetries` increase from 5 to 30 on Windows is a 6x jump. Please add a comment explaining why 30 is the chosen value.

**Questions:**
- Was this empirically determined from testing?
- Is 30 sufficient for worst-case Windows file locking scenarios?
- Should this be configurable?

Without context, this magic number could be too high or too low.

packages/opencode/test/plugin/workspace-adaptor.test.ts
```diff
@@ -45,6 +45,17 @@ async function waitFor(fn: () => boolean, timeout = 5_000) {
   throw new Error("timed out waiting for workspace status")
 }
 
+async function waitForCounter(file: string, min: number) {
+  const end = Date.now() + 5_000
+  let value = 0
+  while (Date.now() < end) {
+    value = Number(await Bun.file(file).text().catch(() => "0"))
+    if (value > min) return value
+    await wait(50)
```
Astro-Han, line 65: P2: The catch handler returns "0" when reading fails, which silently masks:
- File not found errors
- Permission errors
- File corruption

Consider logging the error or throwing if the file exists but is unreadable. The current behavior makes debugging harder when something goes wrong.

packages/opencode/src/file/ripgrep.ts
```diff
@@ -158,7 +159,7 @@ export namespace Ripgrep {
       const config = PLATFORM[platformKey]
       if (!config) throw new UnsupportedPlatformError({ platform: platformKey })
 
-      const version = "14.1.1"
+      const version = "15.1.0"
```
Astro-Han, line 162: P3: Ripgrep 15.1.0 fixes a `--line-buffered` regression from 15.0.0. Adding a brief comment explaining the upgrade reason would help future maintenance:

```ts
// 15.1.0 fixes --line-buffered regression from 15.0.0
const version = "15.1.0"
```

packages/opencode/test/global/runtime-namespace.test.ts
```diff
@@ -25,24 +32,25 @@ function readGlobalPath(namespace?: string) {
   return JSON.parse(Buffer.from(result.stdout).toString()) as Record<string, string>
 }
 
+function expectRoot(paths: Record<string, string>, namespace: string) {
```
Astro-Han: P3: `expectRoot` only validates `data`, `cache`, `config`, `state` but the `pawwork` namespace also has `bin` and `log` fields.

**Suggestions:**
1. Add assertions for `bin` and `log` inside `expectRoot` for completeness
2. Or rename to `expectRootFields` / add a JSDoc clarifying it only validates the 4 common fields

Minor but would make the test more comprehensive.

packages/opencode/test/effect/cross-spawn-spawner.test.ts
```diff
@@ -169,7 +169,10 @@ describe("cross-spawn spawner", () => {
             'process.stderr.write("stderr\\n", done)',
           ].join("\n"),
         )
-        const [stdout, stderr] = yield* Effect.all([decodeByteStream(handle.stdout), decodeByteStream(handle.stderr)])
+        const [stdout, stderr] = yield* Effect.all(
```
Astro-Han, line 174: P3: Adding `concurrency: 2` makes stdout and stderr decode run in parallel. Please add a comment explaining:

- Was this added to fix a race condition where stdout/stderr streams weren't both being captured?
- Or is it purely a performance optimization?

A brief comment would clarify intent.

Astro-Han (commented):
## Review Summary

Overall this PR addresses Windows compatibility issues well. The changes are appropriate for fixing the unit test failures on Windows. However, I found several issues that should be addressed before merge.

---

### P0 (Blocking)
None found.

---

### P1 (Recommend fixing before merge)

1. **ripgrep.ts: inconsistent executable suffix handling** - The `which()` call now explicitly checks for `rg.exe`, but this should be verified against Bun's actual behavior on Windows.

2. **npm/index.ts: global side effect at module load time** - Setting `process.env.__FAKE_PLATFORM__` affects the entire process. Consider wrapping this or documenting as a known compromise.

---

### P2 (Nice to fix, optional)

3. **worktree/index.ts: magic retry number** - The 6x jump from 5 to 30 retries lacks justification.

4. **workspace-adaptor.test.ts: silent error masking** - The catch handler returns "0" which could hide real issues.

---

### P3 (Minor nitpicks)

5. **ripgrep.ts: version upgrade lacks changelog reference** - Adding a comment would help future maintenance.

6. **runtime-namespace.test.ts: expectRoot incomplete** - Only validates 4 of 6 fields.

7. **cross-spawn-spawner.test.ts: concurrency change lacks explanation** - No comment clarifying intent.

---

### Overall assessment

The PR is well-structured with clear intent. Most issues are minor (P2/P3). The P1 issues are worth addressing for code quality but may not block the Windows tests from passing. Recommend addressing P1 items or documenting why the current approach is acceptable.

packages/opencode/src/npm/index.ts
```diff
@@ -1,3 +1,10 @@
...
+// small files. tar swallows that error and can leave dependency folders empty.
+// This must be set before Arborist imports tar.
+if (process.platform === "win32") {
+  process.env.__FAKE_PLATFORM__ = "linux"
+}
```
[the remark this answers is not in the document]
Astro-Han: I kept the implementation as an env-based workaround because the affected code is the third-party `tar` package, which reads `__FAKE_PLATFORM__` during module initialization. A function parameter here would not reach that code path. I updated the comment to document the import-time constraint and why this must be set before dynamic Arborist imports.

packages/opencode/src/file/ripgrep.ts
Astro-Han: Verified against the local `which` wrapper: it delegates to the `which` package with PATHEXT support, so `which("rg")` already resolves `rg.exe` on Windows. I reverted the explicit `.exe` suffix for the system lookup. The managed fallback still constructs `rg.exe`, which is the correct installed filename.

packages/opencode/src/npm/index.ts
Astro-Han: Documented the compromise in code. This does not change `process.platform`; it only sets the env var that `tar` snapshots during module initialization. Temporary scoping around a single Arborist operation is riskier here because Arborist and tar loading happens at import time. Keeping the env set before dynamic Arborist imports matches the upstream workaround and avoids a lazy import missing it.

packages/opencode/src/worktree/index.ts
Astro-Han: Added a comment. The retry budget now points back to the existing test preload cleanup pattern: Windows can keep git and sqlite handles alive briefly after teardown, so EBUSY needs the same longer retry budget. I kept it non-configurable because this is local cleanup hardening, not user-facing behavior.

packages/opencode/test/plugin/workspace-adaptor.test.ts
Astro-Han: Fixed. The helper now treats only ENOENT as 0 while polling for the counter file to appear. Other read errors rethrow, and non-numeric counter contents now throw an explicit error.

packages/opencode/src/file/ripgrep.ts
Astro-Han: Not adding the line-buffered release-note comment. The verified local reason for this PR is Windows managed-install parity with the upstream 15.1.0 baseline, especially Windows ARM64 archive support. Adding an unrelated release-note detail here would make the maintenance context less precise.

packages/opencode/test/global/runtime-namespace.test.ts
Astro-Han: Fixed by renaming the helper to `expectCommonRoots`, so its scope is explicit. The PawWork namespace test continues to assert `bin` and `log` separately because those fields are namespace-specific derived paths.

packages/opencode/test/effect/cross-spawn-spawner.test.ts
Astro-Han: Added a comment. The explicit concurrency is to drain stdout and stderr together for the Windows CI race, not as a performance optimization.

packages/opencode/test/global/runtime-namespace.test.ts
```diff
@@ -1,12 +1,19 @@
 import { describe, expect, test } from "bun:test"
 import path from "path"
 
+const roots = {
+  data: path.join(path.sep, "tmp", "pawwork-runtime-test", "share"),
+  cache: path.join(path.sep, "tmp", "pawwork-runtime-test", "cache"),
+  config: path.join(path.sep, "tmp", "pawwork-runtime-test", "config"),
+  state: path.join(path.sep, "tmp", "pawwork-runtime-test", "state"),
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `Global.Path` import creates the configured data/config/state/log/bin directories and can touch cache state, so the hard-coded shared `/tmp` root was not just a string fixture. The test now uses a `tmpdir` lifecycle per test and derives XDG roots under it.

packages/opencode/test/global/runtime-namespace.test.ts
```diff
@@ -1,12 +1,19 @@
...
-    process.env.XDG_CACHE_HOME = "/tmp/pawwork-runtime-test/cache"
-    process.env.XDG_CONFIG_HOME = "/tmp/pawwork-runtime-test/config"
-    process.env.XDG_STATE_HOME = "/tmp/pawwork-runtime-test/state"
+    process.env.XDG_DATA_HOME = ${JSON.stringify(roots.data)}
+    process.env.XDG_CACHE_HOME = ${JSON.stringify(roots.cache)}
+    process.env.XDG_CONFIG_HOME = ${JSON.stringify(roots.config)}
+    process.env.XDG_STATE_HOME = ${JSON.stringify(roots.state)}
     if (${JSON.stringify(namespace)} !== undefined) {
       process.env.PAWWORK_RUNTIME_NAMESPACE = ${JSON.stringify(namespace)}
     }
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The child process now deletes `PAWWORK_RUNTIME_NAMESPACE` when the test does not pass a namespace, so the default OpenCode case is isolated from the parent environment.

packages/opencode/src/npm/index.ts
```diff
@@ -1,3 +1,11 @@
+// Bun on Windows does not support the UV_FS_O_FILEMAP flag used by tar for
+// small files. tar swallows that error and can leave dependency folders empty.
+// tar snapshots this env var during module initialization, so keep it set
+// before any dynamic Arborist import can load tar.
+if (process.platform === "win32") {
+  process.env.__FAKE_PLATFORM__ = "linux"
```
Astro-Han: **P2** — This mutates `process.env` at module-import time, so it fires for any module that transitively imports `@/npm`, not just callers of `install`/`reify`. It also persists for the rest of the process. Since both install paths now use `await import("@npmcli/arborist")`, the env assignment can live right before each dynamic import — tar will still snapshot on its own module init, and the hack stays colocated with its only consumer. If you're deliberately mirroring an upstream shape for graft/squash cleanliness, mention that in the comment so future readers don't "fix" it.

packages/opencode/test/plugin/workspace-adaptor.test.ts
```diff
@@ -45,6 +45,28 @@ async function waitFor(fn: () => boolean, timeout = 5_000) {
...
+    if (!Number.isFinite(value)) throw new Error(`invalid retry counter value: ${text}`)
+    return value
+  }
+  const end = Date.now() + 5_000
+  let value = 0
+  while (Date.now() < end) {
+    value = await read()
+    if (value > min) return value
+    await wait(50)
+  }
```
Astro-Han, line 66: **P2** — Inconsistent with `waitFor` (line 45) which `throw`s on timeout. Here you silently return the last-read value, so a real timeout surfaces downstream as a confusing `expect(first).toBeGreaterThan(0)` or `expect(second).toBeGreaterThan(first)` failure rather than `timed out waiting for counter > ${min}`. Recommend throwing in the fall-through branch to match `waitFor`'s contract.

packages/opencode/src/worktree/index.ts
```diff
@@ -358,9 +358,19 @@ export namespace Worktree {
       }
 
       function cleanDirectory(target: string) {
+        // Match test preload cleanup: Windows can keep git and sqlite handles
+        // alive briefly after process teardown, so EBUSY needs a longer budget.
+        const maxRetries = process.platform === "win32" ? 30 : 5
```
Astro-Han, line 363: **P3** — The comment accurately references `test/preload.ts` (30-retry rm loop). But `test/fixture/fixture.ts:27-31` does its own `clean(dir)` with `maxRetries: 5, retryDelay: 100` for the very same Windows-handle-lingering reason. If the 3-second budget is the right answer here, the per-fixture teardown is subject to the same EBUSY flake. Not this PR, but worth a follow-up so all three cleanup sites agree.

packages/opencode/src/file/ripgrep.ts
Astro-Han, line 162: **P3** — The `14.1.1 → 15.1.0` bump is not mentioned in the PR title or summary and isn't required by the `ia32-win32` addition — the 14.1.1 release also ships `ripgrep-14.1.1-i686-pc-windows-msvc.zip`. Per the repo's `commit small` / one-reversible-intent rule this reads as a separate concern from Windows test stabilization. Validation so far is a local `test:ci` pass, which likely never downloads on macOS/Linux CI because the managed binary is cached. Consider splitting into its own commit or PR so it can be reverted independently if a downstream regression shows up.

packages/opencode/test/session/prompt-effect.test.ts
```diff
@@ -1271,7 +1271,7 @@ it.live(
       }),
       { git: true, config: providerCfg },
     ),
-  3_000,
+  10_000,
```
Astro-Han: **P3** — 3_000 → 10_000 is a 3.3× bump applied unconditionally. If the real observation is Windows-only queue/shell timing, prefer `process.platform === "win32" ? 10_000 : 3_000` so macOS/Linux don't sit ~7 extra seconds on a real regression. If the 10s budget is genuinely needed on all platforms, a one-line comment explaining the expected lower bound would help future maintainers spot real slowdowns.

Astro-Han (commented):
Nitpick pass on the Windows stabilization PR. Nothing blocking — all P2/P3. Details inline.

- P2: `src/npm/index.ts` mutates a Bun-global env var at module import time.
- P2: `waitForCounter` silently returns last value on timeout; inconsistent with the `waitFor` helper right above it.
- P3: `test/fixture/fixture.ts` still uses `maxRetries: 5` for the same Windows-EBUSY reason.
- P3: The ripgrep `14.1.1 → 15.1.0` bump is outside the stated scope of this PR.
- P3: The `3_000 → 10_000` test timeout is applied to all platforms, not just Windows.

packages/opencode/src/npm/index.ts
```diff
@@ -1,3 +1,11 @@
+// Bun on Windows does not support the UV_FS_O_FILEMAP flag used by tar for
```
Astro-Han: P1: Global side effect at module load time

Setting `process.env.__FAKE_PLATFORM__` at the module top level affects the entire Node.js process. This env var is set permanently for the lifetime of the process.

**Concerns:**
1. The `__FAKE_PLATFORM__` variable name appears undocumented — is this a Bun-internal env var with stability guarantees?
2. Any other module that checks `process.env.__FAKE_PLATFORM__` (perhaps transitively via Bun internals) will now see `"linux"` on Windows.
3. The comment mentions "before any dynamic Arborist import can load tar" but this code runs regardless of whether `Npm.add` or `Npm.install` is ever called.

**Alternative approach:** Move the env var mutation inside `add` and `install` functions, setting it just before the dynamic `import("@npmcli/arborist")` and restoring it afterward. This limits the scope of the workaround to only when npm operations actually happen.

packages/opencode/test/plugin/workspace-adaptor.test.ts
```diff
@@ -45,6 +45,28 @@ async function waitFor(fn: () => boolean, timeout = 5_000) {
   throw new Error("timed out waiting for workspace status")
 }
 
+async function waitForCounter(file: string, min: number) {
```
Astro-Han, line 48: P2: Timeout returns value instead of throwing error

When the 5-second timeout expires, `waitForCounter` returns `value` (the last read) instead of throwing. This is inconsistent with `waitFor` (line 39-45) which throws `"timed out waiting for workspace status"` on timeout.

**Consequence:** A caller like `expect(second).toBeGreaterThan(first)` may produce confusing test failures. If `waitForCounter` times out returning `first`, the `expect` throws `Expected: > first, Received: first` — the error points to `expect` not the timeout logic, making debugging harder.

**Suggested fix:** Throw an error on timeout with contextual information about the file path, expected minimum, and last observed value.

packages/opencode/src/worktree/index.ts
```diff
@@ -358,9 +358,19 @@ export namespace Worktree {
       }
 
       function cleanDirectory(target: string) {
+        // Match test preload cleanup: Windows can keep git and sqlite handles
```
Astro-Han, line 361: P3: Retry count multiplier lacks justification

The jump from `maxRetries: 5` to `maxRetries: 30` (6x) on Windows is significant. The comment mentions "Match test preload cleanup" but `test/fixture/fixture.ts` still uses `maxRetries: 5` for the same Windows EBUSY scenario.

**Question:** Should `fixture.ts` also adopt `maxRetries: 30` on Windows for consistency, or is there a reason the test fixture cleanup needs fewer retries than the worktree cleanup?

packages/opencode/src/file/ripgrep.ts
```diff
@@ -158,7 +159,7 @@ export namespace Ripgrep {
       const config = PLATFORM[platformKey]
       if (!config) throw new UnsupportedPlatformError({ platform: platformKey })
```
Astro-Han, line 161: P3: Version bump without changelog reference

The `14.1.1 → 15.1.0` upgrade is a major version jump. While the PR scope mentions ripgrep lookup, it does not explain why 15.1.0 is necessary.

**Questions:**
1. Does 15.1.0 include a fix that 14.1.1 lacks?
2. Is this upgrade related to the ia32-win32 support addition?
3. Any breaking changes in 15.1.0?

Adding a one-line comment linking to the relevant ripgrep release notes or explaining the motivation would help future maintainers.

packages/opencode/src/file/ripgrep.ts
```diff
@@ -118,6 +118,7 @@ export namespace Ripgrep {
     "x64-darwin": { platform: "x86_64-apple-darwin", extension: "tar.gz" },
     "x64-linux": { platform: "x86_64-unknown-linux-musl", extension: "tar.gz" },
```
Astro-Han, line 119: P3: ia32-win32 support — is this needed?

The ia32 (32-bit x86) Windows platform is increasingly rare. Microsoft ended mainstream support for 32-bit Windows 10 in 2020.

**Questions:**
1. Is there a documented user demand for ia32-win32?
2. Does the CI matrix include ia32 Windows testing?
3. If ia32 binaries fail to download, will the fallback be graceful?

If ia32 support is speculative rather than requested, consider omitting it to reduce maintenance surface.

packages/opencode/test/session/prompt-effect.test.ts
```diff
@@ -1271,7 +1271,7 @@ it.live(
       }),
```
Astro-Han, line 1273: P3: Timeout applies to all platforms, not just Windows

The `3_000 → 10_000` timeout increase addresses Windows CI flake but applies to macOS and Linux CI as well. This does not match the pattern of other Windows-specific fixes in this PR (which use `process.platform === "win32"` guards).

**Alternative:** Consider platform-specific timeouts if the root cause is Windows-specific. This is a minor nit — the 10_000 value is acceptable for all platforms if it simplifies maintenance.

Astro-Han (commented):
Nitpick review for Windows stabilization PR. Issues graded P1-P3 per project convention.

packages/opencode/src/file/ripgrep.ts
```diff
@@ -118,6 +118,7 @@ export namespace Ripgrep {
...
     "arm64-win32": { platform: "aarch64-pc-windows-msvc", extension: "zip" },
+    "ia32-win32": { platform: "i686-pc-windows-msvc", extension: "zip" },
```
Astro-Han, line 121: P3: ia32-win32 support — is this needed?

The ia32 (32-bit x86) Windows platform is increasingly rare. Microsoft ended mainstream support for 32-bit Windows 10 in 2020.

**Questions:**
1. Is there documented user demand for ia32-win32?
2. Does the CI matrix include ia32 Windows testing?
3. If ia32 binaries fail to download, will fallback be graceful?

If ia32 support is speculative, consider omitting it.

packages/opencode/test/effect/cross-spawn-spawner.test.ts
```diff
@@ -169,7 +169,12 @@ describe("cross-spawn spawner", () => {
             'process.stderr.write("stderr\\n", done)',
           ].join("\n"),
         )
-        const [stdout, stderr] = yield* Effect.all([decodeByteStream(handle.stdout), decodeByteStream(handle.stderr)])
+        // Drain both pipes together; Windows CI can race when stderr waits
+        // behind stdout decoding.
+        const [stdout, stderr] = yield* Effect.all(
```
Astro-Han, line 174: P3: Concurrency change lacks inline explanation

The `{ concurrency: 2 }` addition to `Effect.all` is a behavioral change that fixes a Windows CI race condition, but the comment only mentions "Drain both pipes together". A more precise comment like "Windows CI races when stderr decodes behind stdout; parallel decode prevents deadlock" would help future readers understand the intent without needing to trace the error history.

packages/opencode/test/global/runtime-namespace.test.ts
```diff
@@ -25,24 +44,29 @@ function readGlobalPath(namespace?: string) {
   return JSON.parse(Buffer.from(result.stdout).toString()) as Record<string, string>
 }
 
+function expectCommonRoots(paths: Record<string, string>, roots: Roots, namespace: string) {
```
Astro-Han: P3: `expectCommonRoots` only validates 4 of 6 Global.Path fields

The `expectCommonRoots` helper checks `data`, `cache`, `config`, `state` but omits `bin` and `log`. The second test manually asserts those two, but the helper name implies comprehensive validation. Consider either expanding `expectCommonRoots` to all 6 fields or renaming to `expectCoreRoots`.

Astro-Han (commented):
Additional nitpick comments for PR #148.

packages/opencode/src/npm/index.ts
Astro-Han: Fixed in the npm commit. The Bun tar workaround now lives in loadArborist(), which is called directly before both dynamic Arborist imports, so importing @/npm alone no longer mutates process.env. I intentionally do not restore the env var afterward because Arborist and tar can still lazy-load transitive modules after the import boundary.

packages/opencode/src/npm/index.ts
Astro-Han: Fixed in the npm commit. The mutation is no longer module-load behavior and is colocated with the only consumer path. I kept the env var set after the dynamic import because restoring it would be unsafe if Arborist or tar lazy-loads after the import returns.

packages/opencode/test/plugin/workspace-adaptor.test.ts
Astro-Han: Fixed. waitForCounter now throws with the file path, expected minimum, and last observed value when the timeout expires.

packages/opencode/test/plugin/workspace-adaptor.test.ts
```diff
@@ -45,6 +45,28 @@ async function waitFor(fn: () => boolean, timeout = 5_000) {
...
+  }
+  const end = Date.now() + 5_000
+  let value = 0
+  while (Date.now() < end) {
+    value = await read()
+    if (value > min) return value
+    await wait(50)
+  }
+  return value
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed. waitForCounter now fails at the timeout helper with contextual details instead of returning the stale counter value.

packages/opencode/test/plugin/workspace-adaptor.test.ts
Astro-Han: Fixed. The timeout path now throws directly, so failures point to the polling timeout rather than a later expect().

packages/opencode/src/worktree/index.ts
```diff
@@ -358,9 +358,19 @@ export namespace Worktree {
...
           import("fs/promises")
-            .then((fsp) => fsp.rm(target, { recursive: true, force: true, maxRetries: 5, retryDelay: 100 }))
+            .then((fsp) =>
+              fsp.rm(target, {
+                recursive: true,
+                force: true,
+                maxRetries,
+                retryDelay: 100,
+              }),
+            )
```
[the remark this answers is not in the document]
Astro-Han: Fixed. cleanDirectory now uses the injected AppFileSystem.Service and an Effect retry loop with the same Windows retry budget.

packages/opencode/src/worktree/index.ts
Astro-Han: Leaving fixture teardown unchanged in this PR. This PR now routes product worktree cleanup through AppFileSystem with the longer Windows retry budget. Aligning test fixture teardown is a separate test-infra follow-up because it touches a broader cleanup helper outside this issue direction.

packages/opencode/src/worktree/index.ts
Astro-Han: Leaving fixture.ts unchanged here. The product worktree cleanup path is now fixed at its own layer; fixture teardown consistency is worth a follow-up but is outside this PR scope.

packages/opencode/src/file/ripgrep.ts
Astro-Han: Kept this change. I verified the release assets: ripgrep 14.1.1 has i686 and x64 Windows zips, but it does not have aarch64-pc-windows-msvc. 15.1.0 adds the Windows ARM64 asset, which is the upstream #23477 reason. The ripgrep update is already isolated in its own reversible commit.

packages/opencode/src/file/ripgrep.ts
Astro-Han: Kept this change. The practical reason for 15.1.0 is Windows ARM64 asset parity: 14.1.1 lacks aarch64-pc-windows-msvc while 15.1.0 includes it. The PR body already calls out the managed ripgrep version update and it remains a separate commit.

packages/opencode/src/file/ripgrep.ts
Astro-Han: Kept ia32 as upstream platform-map parity, but it is not the reason for the version bump. The non-speculative reason for the bump is Windows ARM64 support, since 14.1.1 lacks the aarch64 Windows asset.

packages/opencode/src/file/ripgrep.ts
Astro-Han: Kept ia32 as upstream parity. I verified 14.1.1 already has the i686 zip, so ia32 is not the version-bump driver; Windows ARM64 is.

packages/opencode/test/session/prompt-effect.test.ts
Astro-Han: Fixed. The two shell queue tests now use 10s only on Windows and keep the original 3s timeout elsewhere.

packages/opencode/test/session/prompt-effect.test.ts
Astro-Han: Fixed. The timeout is now process.platform === "win32" ? 10_000 : 3_000 for the shell queue tests.

packages/opencode/test/effect/cross-spawn-spawner.test.ts
Astro-Han: Fixed. The inline comment now states that the pipes are drained in parallel so Windows CI does not stall when stderr waits behind stdout decoding.

packages/opencode/test/global/runtime-namespace.test.ts
Astro-Han: Fixed by renaming the helper to expectCoreRoots, since it intentionally covers data/cache/config/state while bin and log remain explicit assertions in the PawWork namespace case.

PR #149, opened 2026-04-22
fix(app): stabilize session row left-side visuals (status slot + Pin/Filter icons)

### LINKED ISSUE #143: [Feature] Stabilize session row left-side visuals (status slot width + Pin/Filter icons)
Keep the session name at a stable horizontal baseline in the left sidebar as a session toggles between idle / running / waiting-on-permission / errored / has-unseen-messages. At the same time, make the Pin and Sort (filter) icons visually match the rest of the sidebar icon set instead of rendering noticeably bolder.

### What do you do today?

Two independent issues on the left side of each session row.

**1. Status slot width is conditional.** `packages/app/src/pages/layout/sidebar-items.tsx:116-136` wraps the entire status indicator (`size-6`) in `<Show when={props.isWorking() || props.hasPermissions() || props.hasError() || props.unseenCount() > 0}>`. When a session has none of those four states the whole `size-6` block is removed from layout, so the session title starts roughly one slot-width plus the `gap-2` further left (≈32px) than an adjacent row that does have one of those states. As states come and go (a run starts, a permission prompt clears, an error is acknowledged) the title shifts horizontally.

Note that the Pin button slot to the left of the status slot is *not* the source of the shift. Its base class at `packages/app/src/pages/layout/pawwork-sidebar.tsx:146` (`"inline-flex size-6 ...": true`) is unconditional; Pin toggles `opacity` and `pointer-events` but always reserves `size-6`.

**2. Pin and Filter SVGs are off-family.** `packages/app/src/pages/layout/pawwork-sidebar.tsx` ships two hand-drawn SVGs at `viewBox="0 0 12 12"`: `FilterIcon` (L20-27) uses `stroke-width="1.2"` + `stroke-linecap="round"`, and `PinIcon` (L29-41) uses `stroke-width="1.1"` + `stroke-linejoin="round"`. The majority of sidebar icons in `packages/ui/src/components/icon.tsx` use `viewBox="0 0 20 20"` with default stroke width and per-glyph linecap, so the Pin and Filter icons render noticeably bolder and less uniform than their neighbours.

1. Every session row reserves a stable `size-6` status slot whether or not any status is currently active; the session title starts at the same horizontal position on every row at the same nesting level. The existing priority switch inside the slot is unchanged, including its `tint` from `messageAgentColor(...)` and its behaviour on child sessions. (Effective priority is `permission → running → error → unseen` — `hasPermissions()` forces `isWorking()` to `false` at `sidebar-items.tsx:173`, so permission wins over running even though the `<Switch>` lists `isWorking` first.)
2. Pin and Filter icons sit in the same visual weight family as the rest of the sidebar: 20×20 grid, default stroke width, per-glyph linecap chosen to match neighbouring glyphs.

Both

### Extra context

**Scope** — ship as **one PR with two commits**, one per section below.

*Commit 1 — `fix(app): reserve sidebar status slot width to keep session title baseline stable`*

- `packages/app/src/pages/layout/sidebar-items.tsx:116-136` — change the outer `<Show when={...}>` into an unconditional `size-6` container. The inner `<Switch>` (running / permission / error / unseen) stays conditional, so when no state is active the container renders empty but still occupies `size-6` of width.
- No behaviour change for active states: spinner keeps its `tint`, dot colours unchanged, priority order unchanged.
- Child sessions are automatically covered for the **status slot** because `SessionRow` is the same component for every session regardless of `level`. The **Pin slot** remains top-level only by design — `SessionItem` gates the Pin `leadingSlot` on `!props.level` at `sidebar-items.tsx:231`, and this issue does not change that. Child rows therefore still have a narrower leading offset than top-level rows; that is intentional, not a regression.

*Commit 2 — `refactor(app): redraw sidebar Pin/Filter icons on the 20×20 grid`*

- `packages/app/src/pages/layout/pawwork-sidebar.tsx:20-27` (`FilterIcon`) and `:29-41` (`PinIcon`) — change `viewBox` to `"0 0 20 20"` and **rewrite the path coordinates to the 20-unit grid** (changing only the `viewBox` attribute would render the glyph at ~60% of its visual size). Drop the explicit `stroke-width`. Pick `stroke-linecap` per glyph to match the visual weight of the neighbouring icons it sits beside (`new-session`, `dot-grid`, `archive`); upstream `icon.tsx` uses both `square` and `round` on different glyphs, so there is no single "correct" value to paste.
- The icons stay file-local in `pawwork-sidebar.tsx`. Do **not** add them to `packages/ui/src/components/icon.tsx`, which follows upstream.

**Relationship to #77**

- Supersedes #77 P0-4 (PinIcon / FilterIcon redraw). That item moves into Commit 2 here.
- #77 P1-5 ("Reserve the brand accent for primary actions" — swap pinned / filter-active `text-accent-brand` to `text-text-strong` or `text-accent-brand/60`) stays in #77 for a later polish pass.

**Not in scope**

- Pin slot. Already `size-6` unconditional, no change needed.
- Pin button a11y contract (`tabIndex={isPinned ? 0 : -1}` + `aria-hidden={isPinned ? undefined : "true"}`, hover-to-discover on unpinned rows) — unchanged.
- Status slot priority switch internals — unchanged.
- Accent-brand colour on pinned / filter-active states — unchanged (see #77 P1-5).
- Top running progress bar — unchanged.
- Sort button interaction — redraw only.
- Promoting Pin / Filter into `packages/ui/src/components/icon.tsx`.

**Manual UI verification** (run `bun dev:desktop`)

After Commit 1:
1. Open a project with a mix of idle, running, pinned, and unpinned sessions. Scroll through and hover each row. Titles do not shift horizontally — every row at the same nesting level starts at the same x-position regardless of status slot activity.
2. Start a session. Spinner appears in the status slot (in agent `tint`); title does not move.
3. A session reporting a permission request, an error, or unseen messages shows the yellow / red / blue dot respectively in the status slot; title does not move.
4. Expand a parent session that has child sessions. Child rows obey the same stable baseline; their status indicators (when any) appear in the reserved slot.

After Commit 2:
5. Place the sidebar next to rows that use the `new-session` glyph and the `dot-grid` action menu. Pin and Filter icons read as the same visual weight — no bolder stroke, no tighter curves.
6. Pinned, no hover → filled pin in `text-accent-brand` on a 20×20 grid.
7. Unpinned, hover → outline pin in `text-text-weak` on the 20×20 grid.
8. Sort button active (sort=project) → filter icon in `text-accent-brand` on the 20×20 grid.
9. Keyboard-only: Tab through a pinned session row → focus lands on pin button, Enter toggles unpin. Tab through an unpinned row → pin button is not in the tab order (unchanged behaviour, just re-verify the redraw did not regress the contract).

Astro-Han:
Closes #143.

## Summary

- **Commit 1** (`fix`) removes the outer `<Show>` wrapper around the sidebar session-row status slot in `sidebar-items.tsx`, so the `size-6` container always reserves 24px of width. The inner `<Switch>` (running / permission / error / unseen) is unchanged. Session title baseline no longer shifts when status indicators toggle.
- **Commit 2** (`refactor`) redraws `FilterIcon` and `PinIcon` in `pawwork-sidebar.tsx` on the 20x20 grid. `FilterIcon` uses `stroke-linecap="square"` to match upstream straight-line glyphs (`close`, `arrow-right`, `menu`). `PinIcon` uses a hexagonal head polygon + diagonal needle, default stroke width, default size 16. Needle origin `(8, 11)` sits on the head's bottom-left edge `y = x + 3` for clean attachment.

## Follow-up (out of scope)

During live review we noticed that the Pin button (rendered in `SessionItem` `leadingSlot`) and the status slot (inside `SessionRow`) are two separate reserved slots. The original intent implied by the issue reporter was for running state to appear *in the pin button position* — i.e. one unified slot with priority `running > permission > error > unseen > pin > empty`. This PR ships the #143 scope as written (two stable slots, matching the issue body's "Not in scope: Pin slot" note), and defers the merge refactor to #150. Merge is medium-risk: changes pin button a11y contract (pin must remain keyboard-reachable when pinned + running), requires passing leadingSlot into `SessionRow`, and shifts the top-level title baseline left by 28px.

## Test plan

- [x] `bun run typecheck` passes
- [x] Manual live verification on `bun dev:desktop`:
  - idle / pinned / running rows maintain stable title baseline
  - Pin icon at size 16 reads at roughly the visual weight of neighbouring sidebar icons
  - Filter icon three bars descending cleanly on 20x20 grid
  - Pinned (accent brand) / hover-to-reveal (weak) pin states render correctly
- [x] crosscheck reviewed (zero P0/P1; one P2 geometry fix autosquashed)

### DIFF
--- a/packages/app/src/pages/layout/pawwork-sidebar.tsx
+++ b/packages/app/src/pages/layout/pawwork-sidebar.tsx
@@ -20,22 +20,23 @@ export type PawworkSidebarSession = {
 const FilterIcon = (props: { size?: number }) => {
   const size = props.size ?? 14
   return (
-    <svg width={size} height={size} viewBox="0 0 12 12" fill="none" aria-hidden="true">
-      <path d="M1.5 3h9M3 6h6M4.5 9h3" stroke="currentColor" stroke-width="1.2" stroke-linecap="round" />
+    <svg width={size} height={size} viewBox="0 0 20 20" fill="none" aria-hidden="true">
+      <path d="M2.5 5h15M5 10h10M7.5 15h5" stroke="currentColor" stroke-linecap="square" />
     </svg>
   )
 }
 
 const PinIcon = (props: { filled?: boolean; size?: number }) => {
-  const size = props.size ?? 12
+  const size = props.size ?? 16
   return (
-    <svg width={size} height={size} viewBox="0 0 12 12" fill={props.filled ? "currentColor" : "none"} aria-hidden="true">
+    <svg width={size} height={size} viewBox="0 0 20 20" fill="none" aria-hidden="true">
       <path
-        d="M7.5 1.5l3 3-2 .5-1.5 3-1.5-1.5-2.5 2.5L3.5 9 6 6.5 4.5 5l3-1.5.5-2z"
+        d="M11 2L18 9L14 11L11 14L6 9L9 6Z"
         stroke="currentColor"
-        stroke-width="1.1"
-        stroke-linejoin="round"
+        stroke-linejoin="miter"
+        fill={props.filled ? "currentColor" : "none"}
       />
+      <path d="M8 11L2 18" stroke="currentColor" stroke-linecap="square" />
     </svg>
   )
 }
--- a/packages/app/src/pages/layout/sidebar-items.tsx
+++ b/packages/app/src/pages/layout/sidebar-items.tsx
@@ -113,27 +113,25 @@ const SessionRow = (props: {
         props.clearHoverProjectSoon()
       }}
     >
-      <Show when={props.isWorking() || props.hasPermissions() || props.hasError() || props.unseenCount() > 0}>
-        <div
-          class="shrink-0 size-6 flex items-center justify-center"
-          style={{ color: props.tint() ?? "var(--icon-interactive-base)" }}
-        >
-          <Switch>
-            <Match when={props.isWorking()}>
-              <Spinner class="size-[15px]" />
-            </Match>
-            <Match when={props.hasPermissions()}>
-              <div class="size-1.5 rounded-full bg-surface-warning-strong" />
-            </Match>
-            <Match when={props.hasError()}>
-              <div class="size-1.5 rounded-full bg-text-diff-delete-base" />
-            </Match>
-            <Match when={props.unseenCount() > 0}>
-              <div class="size-1.5 rounded-full bg-text-interactive-base" />
-            </Match>
-          </Switch>
-        </div>
-      </Show>
+      <div
+        class="shrink-0 size-6 flex items-center justify-center"
+        style={{ color: props.tint() ?? "var(--icon-interactive-base)" }}
+      >
+        <Switch>
+          <Match when={props.isWorking()}>
+            <Spinner class="size-[15px]" />
+          </Match>
+          <Match when={props.hasPermissions()}>
+            <div class="size-1.5 rounded-full bg-surface-warning-strong" />
+          </Match>
+          <Match when={props.hasError()}>
+            <div class="size-1.5 rounded-full bg-text-diff-delete-base" />
+          </Match>
+          <Match when={props.unseenCount() > 0}>
+            <div class="size-1.5 rounded-full bg-text-interactive-base" />
+          </Match>
+        </Switch>
+      </div>
       <div class="min-w-0 flex-1 flex items-center gap-2">
         <Show when={props.titleContent} fallback={<span class="text-14-regular text-text-strong min-w-0 flex-1 truncate">{title()}</span>}>
           {props.titleContent}


packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -20,22 +20,23 @@ export type PawworkSidebarSession = {
...
-    <svg width={size} height={size} viewBox="0 0 12 12" fill={props.filled ? "currentColor" : "none"} aria-hidden="true">
+    <svg width={size} height={size} viewBox="0 0 20 20" fill="none" aria-hidden="true">
       <path
-        d="M7.5 1.5l3 3-2 .5-1.5 3-1.5-1.5-2.5 2.5L3.5 9 6 6.5 4.5 5l3-1.5.5-2z"
+        d="M11 2L18 9L14 11L11 14L6 9L9 6Z"
         stroke="currentColor"
-        stroke-width="1.1"
-        stroke-linejoin="round"
+        stroke-linejoin="miter"
+        fill={props.filled ? "currentColor" : "none"}
```
Astro-Han, line 37: [P3] In the filled state this now fills only the pin head, while the needle remains the separate stroked open path below. The old single-path icon filled the whole pushpin when `filled` was true, and issue #143's manual check calls for a filled pin in the pinned state, so the active pin can read like a mixed filled/outline glyph. Consider making the needle a closed filled shape in the active variant, or keeping a single filled path for `filled` pins.

Astro-Han (commented):
## Review Summary

Overall the changes align with the issue requirements, but there are some geometry and visual consistency concerns worth addressing.

### P0: None

### P1: 1 issue
- **PinIcon geometry mismatch**: The pin head is a pentagon (not hexagon as described), and needle origin is slightly misaligned with the bottom edge.

### P2: 2 issues
- **FilterIcon rendering size**: size=14 is 70% of 20x20 grid, making it smaller than adjacent icons (16px/80%)
- **Icon shape redesign scope**: PinIcon changed from traditional pushpin to geometric shape without explicit mention in issue

### P3: 1 note
- **Empty Switch fallback**: When no status matches, Switch renders empty - this is intentional for baseline stability

See inline comments for details.

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -20,22 +20,23 @@ export type PawworkSidebarSession = {
...
       />
+      <path d="M8 11L2 18" stroke="currentColor" stroke-linecap="square" />
```
Astro-Han, line 39: **P1: PinIcon geometry mismatch with PR description**

The PR description states the pin head is "hexagonal" and the needle origin sits on the bottom edge y = x + 3. However:

1. The path M11 2L18 9L14 11L11 14L6 9L9 6Z has only 5 distinct vertices, forming a pentagon/kite shape, not a hexagon.

2. The bottom edge runs from (6, 9) to (11, 14) satisfying y = x + 3. The needle starts at (8, 11), but 11 ≠ 8 + 3 — it is off the edge by approximately (0.5, -0.5) units.

This geometric imprecision may cause a visible gap or misalignment between the pin head and needle, especially with stroke-linecap="square" which adds square caps at endpoints.

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -20,22 +20,23 @@ export type PawworkSidebarSession = {
 const FilterIcon = (props: { size?: number }) => {
   const size = props.size ?? 14
   return (
-    <svg width={size} height={size} viewBox="0 0 12 12" fill="none" aria-hidden="true">
-      <path d="M1.5 3h9M3 6h6M4.5 9h3" stroke="currentColor" stroke-width="1.2" stroke-linecap="round" />
+    <svg width={size} height={size} viewBox="0 0 20 20" fill="none" aria-hidden="true">
```
Astro-Han, line 23: **P2: FilterIcon rendering size smaller than adjacent icons**

The size prop defaults to 14, and the call site also passes size=14. With viewBox="0 0 20 20", this scales the 20-unit grid to 14px (70% scale).

Adjacent sidebar icons from packages/ui/src/components/icon.tsx:
- Icon size="small" → 16px (80% scale of 20x20 grid)
- Icon size="normal" → 20px (100% scale)

At 14px, FilterIcon renders smaller than the typical "small" icon size (16px). This may make the sort button icon visually smaller than its neighbors (e.g., new-session at 16px, dot-grid action menu at 16px).

Consider aligning FilterIcon size to 16 for consistency with other sidebar icons, or verify that the 14px size intentionally matches the surrounding visual weight.

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -20,22 +20,23 @@ export type PawworkSidebarSession = {
...
+      <path d="M2.5 5h15M5 10h10M7.5 15h5" stroke="currentColor" stroke-linecap="square" />
     </svg>
   )
 }
 
 const PinIcon = (props: { filled?: boolean; size?: number }) => {
-  const size = props.size ?? 12
+  const size = props.size ?? 16
```
Astro-Han, line 30: **P2: PinIcon default size vs container spacing**

The PinIcon default size is set to 16, which aligns with Icon size="small" (16px). This is consistent.

However, note that the Pin button container uses size-6 (24px) class at line 146 in the render section, while the icon itself is 16px. The icon is centered in a 24px container, which provides 8px of padding (4px on each side). Verify this spacing works well visually with the new geometric shape design.

packages/app/src/pages/layout/sidebar-items.tsx
```diff
@@ -113,27 +113,25 @@ const SessionRow = (props: {
...
-            <Match when={props.hasError()}>
-              <div class="size-1.5 rounded-full bg-text-diff-delete-base" />
-            </Match>
-            <Match when={props.unseenCount() > 0}>
-              <div class="size-1.5 rounded-full bg-text-interactive-base" />
-            </Match>
-          </Switch>
-        </div>
-      </Show>
+      <div
```
Astro-Han, line 116: **P3: Empty Switch behavior for baseline stability**

When all four conditions (isWorking, hasPermissions, hasError, unseenCount > 0) are false, the Switch component renders no content (no Match is satisfied), and the outer div remains empty but still reserves size-6 width.

This is intentional to keep the session title baseline stable regardless of status changes. The empty container is a design trade-off for layout stability.

Minor note: the style={{ color: props.tint() ... }} is still computed and applied to the empty container, which is harmless but slightly redundant when no content is visible.

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -20,22 +20,23 @@ export type PawworkSidebarSession = {
 const FilterIcon = (props: { size?: number }) => {
   const size = props.size ?? 14
   return (
-    <svg width={size} height={size} viewBox="0 0 12 12" fill="none" aria-hidden="true">
-      <path d="M1.5 3h9M3 6h6M4.5 9h3" stroke="currentColor" stroke-width="1.2" stroke-linecap="round" />
+    <svg width={size} height={size} viewBox="0 0 20 20" fill="none" aria-hidden="true">
+      <path d="M2.5 5h15M5 10h10M7.5 15h5" stroke="currentColor" stroke-linecap="square" />
```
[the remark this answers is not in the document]
Astro-Han: Deliberate choice, not oversight. The redraw's stated goal is aligning with the upstream family in `packages/ui/src/components/icon.tsx`, which uses the SVG default stroke (`1`) on a 20x20 grid — not preserving the previous 1.2/12 effective weight. Matching the family is what makes the sidebar icons read as a set; halving the on-screen stroke is the intended outcome. Verified visually on `bun dev:desktop` alongside `new-session`, `close`, `menu`, `arrow-right`.

packages/app/src/pages/layout/pawwork-sidebar.tsx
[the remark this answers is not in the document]
Astro-Han: Same rationale as the FilterIcon thread: deliberate alignment with the upstream icon family (default stroke on 20x20). On `stroke-linejoin="miter"`: the polygon's six vertices form interior angles of 90°-108° (e.g. (11,14) is 90°, (18,9) and (11,2) land near 108°), all well inside SVG's default miter-limit of 4, so no spike clipping occurs. `round` would soften the angular silhouette deliberately chosen to match the upstream rectilinear + diagonal pattern (cf. `pencil-line`, `fork`).

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: Valid observation, deliberate trade-off. The new design follows the upstream `pencil-line`/`fork` pattern (closed rectilinear region + separate diagonal stroke), which is stronger family parity than a single-path pushpin silhouette. The cost is that filled state renders head-solid + needle-stroked rather than unified. Verified visually on `bun dev:desktop` for both pinned (accent brand) and hover-outline (weak) states — reads as a pin at size 16 on a 24px button. Flagging as potential polish if user feedback surfaces; not blocking.

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: Both claims check false.

**Vertex count.** Path `M11 2L18 9L14 11L11 14L6 9L9 6Z` has six distinct pre-close coordinates: (11,2), (18,9), (14,11), (11,14), (6,9), (9,6). Hexagon, not pentagon/kite.

**Needle origin on edge.** Edge (6,9)→(11,14) has slope 1 through (6,9), so its equation is `y = x + 3`. At x=8: y = 8 + 3 = **11**. The needle starts at (8, 11). 11 = 11 — the point is exactly on the edge. (This is in fact an autosquashed fix from an earlier crosscheck round that moved the origin from (9,11) to (8,11) specifically to land on this edge.)

Downgrading the concern.

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: Call site is the solo sort button in the "全部会话" label row (`pawwork-sidebar.tsx:310-325`), which is icon-only with `p-1` wrapper — no adjacent icons inside it to compare against. Verified visually on `bun dev:desktop` at size=14; the three bars read at appropriate weight for that context. The plan preserved the original 14 default rather than second-guess a call site that already opted into 14 explicitly. Bumping to 16 is a valid alternative; deferring to a follow-up if visual feedback surfaces.

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: Verified visually on `bun dev:desktop`. 4px inset reads well for both outline (hover) and filled (pinned) states on the size-6 button. The default jumped from 12 → 16 precisely after live review showed 12 was too small; 16 was the chosen landing point.

packages/app/src/pages/layout/sidebar-items.tsx
Astro-Han: Intentional. `style.color` is a CSS custom-property fallback that costs nothing to apply to an empty container (no paint when no content). Adding a conditional `style` would trade zero runtime cost for extra JSX complexity, which is the opposite of this refactor's direction (remove conditional wrapper to stabilize layout). Leaving as is.

PR #152, opened 2026-04-22
fix(desktop): harden Electron Windows runtime

### LINKED ISSUE #141: Improve Windows compatibility and restore Windows unit signal

Astro-Han:
## Summary
- Lazy-load `electron-store` so settings storage is not constructed during main-process module import.
- Move renderer startup data behind preload IPC and run renderer windows with sandbox, contextIsolation, and no renderer Node access.
- Load packaged renderer HTML through the privileged `pawwork-renderer://renderer` origin and narrow sidecar CORS to that origin.

## Why
This is direction 2 from #141: Electron/process-layer Windows hardening. It keeps the PR scoped to upstream candidates #23373, #23523, and #23516. It does not include #23633, #23396, Windows ARM64, or the opencode-layer work already covered by #148.

## Related Issue
Closes #141. PR #148 already covered the opencode-layer Windows compatibility track; this PR covers the remaining Electron/process-layer Windows hardening track.

## How To Verify
- `cd packages/desktop-electron && bun test`
- `cd packages/desktop-electron && bun run typecheck`
- `cd packages/desktop-electron && bun run build`

## Screenshots/Recordings
N/A, desktop runtime hardening only.

## Checklist
- [x] Tests added or updated
- [x] Focused and full desktop verification run locally
- [x] PR remains scoped to the Electron/process-layer slice of #141

## Summary by CodeRabbit

* **New Features**
  - Implemented custom renderer protocol for serving application assets securely
  - Exposed window configuration and deep link handling through new IPC-based APIs

* **Improvements**
  - Renderer process now operates with enhanced isolation through sandboxing and context isolation
  - CORS policy restricted to renderer protocol origin for improved security posture


packages/desktop-electron/src/main/renderer-protocol.ts
```diff
@@ -0,0 +1,31 @@
...
+  let url: URL
+  try {
+    url = new URL(requestUrl)
+  } catch {
+    return null
+  }
+
+  if (url.protocol !== `${rendererProtocol}:`) return null
+  if (url.hostname !== rendererHost) return null
+
```
Astro-Han, line 24: **P1:** `decodeURIComponent` throws `URIError` on malformed input (e.g. `pawwork-renderer://renderer/%zz`). That will escape this function as an unhandled rejection inside `protocol.handle`, which means the renderer gets a network error instead of the intentional 404. Wrap the decode in a try/catch and return `null` the same way the `new URL` guard above does, or just skip decoding and let `path.normalize` + the `relative(..).startsWith('..')` check do the work. A fuzz input from inside the renderer (an `` in packaged HTML) is enough to trip this.

packages/desktop-electron/src/main/renderer-protocol.ts
```diff
@@ -0,0 +1,31 @@
...
+  const path = decodeURIComponent(url.pathname)
```
Astro-Han: **P2:** `pawwork-renderer://renderer` with no path yields `url.pathname === ""` under the WHATWG URL parser for non-special schemes (before `registerSchemesAsPrivileged` kicks in, which is the exact path your test harness runs in). `path === "/"` is false, `path.slice(1)` is `""`, `resolve(root, "")` returns `root` itself, and `net.fetch` on a directory file URL errors out instead of returning the index. Collapse the check to `path === "" || path === "/"`. Worth adding a test case alongside the existing `%2F` traversal one.

packages/desktop-electron/src/main/renderer-protocol.ts
```diff
@@ -0,0 +1,31 @@
...
+  const file = resolve(root, path === "/" ? "index.html" : path.slice(1))
+  const rel = relative(root, file)
+  if (rel.startsWith("..") || isAbsolute(rel)) return null
+  return file
+}
+
+export function rendererRootFromMain(root: string) {
+  return join(root, "../renderer")
+}
```
Astro-Han: **P3:** `rendererRootFromMain` is a one-liner used in exactly one place (`windows.ts`). Exporting it, adding a test for it, and routing it through the module surface is more ceremony than the `join(root, "../renderer")` deserves. Inline it at the caller and drop the export.

packages/desktop-electron/src/main/windows.ts
```diff
@@ -1,16 +1,25 @@
...
 
-type Globals = {
-  updaterEnabled: boolean
-  deepLinks?: string[]
-}
-
 const root = dirname(fileURLToPath(import.meta.url))
+const rendererRoot = rendererRootFromMain(root)
+
+protocol.registerSchemesAsPrivileged([
```
Astro-Han: **P2:** this is a module-import side effect that depends on being evaluated before `app.ready`. It works today only because `index.ts` imports `./windows` at top level. A future refactor that lazy-loads or dynamically imports this module (or any test that imports it after `whenReady` fires) will silently no-op the scheme registration and break the renderer with a cryptic fetch failure. Either leave a one-line comment pinning the invariant, or split the registration out into an exported `registerRendererScheme()` that `index.ts` calls explicitly before `whenReady` — paired with `registerRendererProtocol()` which handles the post-ready half.

packages/desktop-electron/src/main/windows-protocol.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe("desktop packaged renderer protocol", () => {
+  test("registers a privileged renderer scheme and loads packaged HTML from it", () => {
+    expect(source).toContain("protocol.registerSchemesAsPrivileged")
+    expect(source).toContain("scheme: rendererProtocol")
+    expect(source).toContain("secure: true")
+    expect(source).toContain("standard: true")
+    expect(source).toContain("corsEnabled: true")
+    expect(source).toContain("supportFetchAPI: true")
+    expect(source).toContain("protocol.handle(rendererProtocol")
```
Astro-Han, line 16: **P2:** the three `windows-*.test.ts` and `ipc-window-config.test.ts` files added in this PR all grep source strings rather than exercising behavior. `expect(source).not.toContain("win.loadFile")`, `toContain("sandbox: true")`, etc. can be satisfied by commented-out code, and they won't catch a refactor that keeps the tokens but changes the semantics (e.g. `const sandbox: boolean = true` assigned to an unused variable). For the security-critical invariants (`sandbox`/`contextIsolation`/`nodeIntegration`), consider a harness that instantiates a hidden `BrowserWindow` with the production options and asserts via `win.webContents.getWebPreferences()` — ~10 lines and actually tests the contract. The string-match layer can stay as a cheap smoke.

packages/desktop-electron/src/main/server.test.ts
```diff
@@ -144,6 +148,9 @@ describe("desktop server runtime namespace", () => {
         OPENCODE_CLIENT: "desktop",
         OPENCODE_SERVER_USERNAME: "PawWork",
       })
+      expect(listenOptions).toMatchObject({
+        cors: ["pawwork-renderer://renderer"],
```
Astro-Han: **P2:** hardcodes `"pawwork-renderer://renderer"` instead of importing `rendererOrigin` from `./renderer-protocol`. If the constant is ever renamed, prod breaks silently while this test stays green. `import { rendererOrigin } from "./renderer-protocol"` and assert against it.

packages/desktop-electron/src/renderer/startup-state.ts
```diff
@@ -0,0 +1,55 @@
...
+  let initialDeepLinks: string[] = []
+
+  const ready = Promise.all([api.getWindowConfig(), api.consumeInitialDeepLinks()])
+    .then(([nextConfig, nextDeepLinks]) => {
+      config = nextConfig
+      initialDeepLinks = Array.isArray(nextDeepLinks) ? nextDeepLinks : []
+    })
+    .catch(() => {
+      config = closedConfig
+      initialDeepLinks = []
```
Astro-Han, line 36: **P2:** the `.catch(() => ...)` swallows IPC failures with no breadcrumb. If `getWindowConfig` or `consumeInitialDeepLinks` throws in prod (preload mis-wired, schema drift, renamed channel), the app boots with updater disabled, WSL disabled, and any queued deep link dropped — with nothing in the console. Add a `console.warn("[desktop] startup IPC failed", err)` so the next bug report is diagnosable. Low-cost, high-diagnostic-value.

packages/desktop-electron/src/renderer/startup-state.ts
```diff
@@ -0,0 +1,55 @@
...
+  wslEnabled: false,
+}
+
+const closedApi: Pick<ElectronAPI, "getWindowConfig" | "consumeInitialDeepLinks"> = {
+  getWindowConfig: async () => closedConfig,
+  consumeInitialDeepLinks: async () => [],
+}
+
+function defaultApi() {
+  return typeof window === "undefined" ? closedApi : window.api
```
Astro-Han: **P3:** `defaultApi()` guards `typeof window === "undefined"` but not the case where `window` exists and `window.api` is undefined (preload failed to contextBridge-expose `api`). That path reaches `api.getWindowConfig()` which throws `TypeError: Cannot read properties of undefined`, which is then caught by the `.catch` above — so it works, but by accident. Tighten to `window?.api ?? closedApi` and the intent matches the behavior.

packages/desktop-electron/src/renderer/startup-state.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function pushPendingDeepLinks(target: DeepLinkWindow, urls: string[]) {
+  if (urls.length === 0) return
+  target.__OPENCODE__ ??= {}
+  const pending = target.__OPENCODE__.deepLinks ?? []
+  target.__OPENCODE__.deepLinks = [...pending, ...urls]
+}
+
+export const startupState = createStartupState()
```
Astro-Han: **P3:** top-level `createStartupState()` executes at module import, which means every unit test that imports this module (directly or transitively) fires a real IPC invoke unless `window.api` is mocked first. The added test file creates its own instance via `createStartupState()`, which is the right pattern — but leaving the singleton in module scope makes it easy for a future test to accidentally await `startupState.ready` with the real `window` globals from another test leaking in. Consider exporting a lazy getter or at least documenting that consumers must not touch the singleton in tests.

packages/desktop-electron/src/renderer/index.tsx
```diff
@@ -314,119 +317,121 @@ let menuTrigger = null as null | ((id: string) => void)
...
-    const server: ServerConnection.Sidecar = {
-      displayName: "Local Server",
-      type: "sidecar",
-      variant: "base",
-      http: {
-        url: data.url,
-        username: data.username ?? undefined,
-        password: data.password ?? undefined,
-      },
+void startupState.ready.then(() => {
```
Astro-Han, line 321: **P3:** the entire render bootstrap now awaits `startupState.ready`, which is one preload IPC roundtrip. It's sub-millisecond in the happy path, but `get-window-config` transitively calls `getWslConfig()` which hits `electron-store` — and thanks to the lazy-init commit in this PR, the *first* call pays the `electron-store` constructor cost (file read on disk). Not a regression worth fixing, but worth knowing: the first paint is now gated on a synchronous disk read in main. If you ever see a loading-window flash regression on slow spinning-disk Windows boxes, this is where to look.

packages/desktop-electron/src/main/server.ts
```diff
@@ -38,6 +39,7 @@ export async function spawnLocalServer(hostname: string, port: number, password:
   const listener = await Server.listen({
     port,
     hostname,
```
Astro-Han, line 41: **P3:** `getStore()` is now called four times per settings touch (get/set url, get/set wsl), each going through the `cache.get(name)` lookup. Trivial cost, but the old singleton was also trivial, so the only reason to switch was the lazy-init requirement. Could hoist to a module-local `const store = () => getStore()` if you want a shorter name, or leave as-is. Purely aesthetic.

Astro-Han (commented):
Nits only; the three-commit slice is well scoped and the threat model is coherent. Findings below are mostly defensive-hardening gaps and test-strength nitpicks.

packages/desktop-electron/src/renderer/startup-state.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    consumeInitialDeepLinks: () => initialDeepLinks.splice(0),
+  }
+}
+
+export function pushPendingDeepLinks(target: DeepLinkWindow, urls: string[]) {
+  if (urls.length === 0) return
+  target.__OPENCODE__ ??= {}
+  const pending = target.__OPENCODE__.deepLinks ?? []
+  target.__OPENCODE__.deepLinks = [...pending, ...urls]
```
Astro-Han: **P3:** `pushPendingDeepLinks` allocates a fresh array on each call via `[...pending, ...urls]` rather than `pending.push(...urls)`. The whole purpose here is transient write-through to `__OPENCODE__.deepLinks` before the app layout drains it, so the copy buys nothing — it just creates GC pressure if deep links arrive in a burst (e.g. `second-instance` firing multiple URLs). `target.__OPENCODE__.deepLinks = pending` after `pending.push(...urls)` is equivalent and cheaper.

packages/desktop-electron/src/main/renderer-protocol.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    const root = "/Applications/PawWork.app/Contents/Resources/app.asar/out/renderer"
+
+    expect(resolveRendererFile(root, "pawwork-renderer://renderer/index.html")).toBe(join(root, "index.html"))
+    expect(resolveRendererFile(root, "pawwork-renderer://renderer/assets/app.js")).toBe(
+      join(root, "assets/app.js"),
+    )
+    expect(resolveRendererFile(root, "pawwork-renderer://renderer/..%2Fmain%2Findex.js")).toBeNull()
+    expect(resolveRendererFile(root, "pawwork-renderer://wrong/index.html")).toBeNull()
+    expect(resolveRendererFile(root, pathToFileURL(join(root, "index.html")).toString())).toBeNull()
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Added malformed URL coverage in `renderer-protocol.test.ts`; `resolveRendererFile` returns `null` for invalid URL input. Classified as P3.

packages/desktop-electron/src/main/renderer-protocol.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (url.protocol !== `${rendererProtocol}:`) return null
+  if (url.hostname !== rendererHost) return null
+
+  const path = decodeURIComponent(url.pathname)
+  const file = resolve(root, path === "/" ? "index.html" : path.slice(1))
+  const rel = relative(root, file)
+  if (rel.startsWith("..") || isAbsolute(rel)) return null
+  return file
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `resolveRendererFile` now rejects decoded null bytes before path resolution and covers `%00` in tests. Classified as P2.

packages/desktop-electron/src/main/store.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+describe("desktop store", () => {
+  test("does not instantiate electron-store at module import time", async () => {
+    const { getStore } = await import("./store")
+
+    expect(constructorCalls).toBe(0)
+    getStore()
+    expect(constructorCalls).toBe(1)
+    getStore()
+    expect(constructorCalls).toBe(1)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The test now uses a unique store name and resets constructor count, so prior default-store cache state cannot invalidate the assertion. Classified as P1 because it was failing CI.

packages/desktop-electron/src/main/renderer-protocol.ts
Astro-Han: Fixed. Wrapped `decodeURIComponent` and return `null` on `URIError`, preserving the protocol handler fallback path. Classified as P1.

packages/desktop-electron/src/main/renderer-protocol.ts
Astro-Han: Fixed. Empty and slash-only renderer URLs now resolve to `index.html`, with coverage for `pawwork-renderer://renderer`. Classified as P2.

packages/desktop-electron/src/main/renderer-protocol.ts
Astro-Han: Fixed. Removed `rendererRootFromMain` and inlined the root join in `windows.ts`. Classified as P3.

packages/desktop-electron/src/main/windows.ts
Astro-Han: Fixed. Split pre-ready scheme registration into `registerRendererScheme()` and call it synchronously in `setupApp()` before `app.whenReady()`. Classified as P2.

packages/desktop-electron/src/main/windows-protocol.test.ts
Astro-Han: Partially fixed. The security-critical window preferences now go through a pure `rendererWebPreferences()` helper with behavior coverage, avoiding brittle Electron imports in Bun. I kept the protocol and IPC string smoke tests because importing Electron main modules in these Bun unit tests previously caused mock leakage and CJS named-export failures. Classified as P2 for the window preference contract, P3 for the remaining smoke-test limitation.

packages/desktop-electron/src/main/server.test.ts
Astro-Han: Fixed. The server test now imports `rendererOrigin` from `./renderer-protocol`. Classified as P2.

packages/desktop-electron/src/renderer/startup-state.ts
Astro-Han: Fixed. Startup IPC fallback now logs `console.warn("[desktop] startup IPC failed", error)` and has test coverage. Classified as P2.

packages/desktop-electron/src/renderer/startup-state.ts
Astro-Han: Fixed. `defaultApi()` now uses `window.api ?? closedApi`, with coverage for preload not exposing `api`. Classified as P3.

packages/desktop-electron/src/renderer/startup-state.ts
Astro-Han: Fixed. Replaced the eager exported singleton with lazy `getStartupState()` and updated consumers. Classified as P3.

packages/desktop-electron/src/renderer/index.tsx
Astro-Han: Left unchanged. The startup IPC gate is intentional so updater, WSL, and initial deep links are loaded before renderer bootstrap. The possible first-paint cost is noted but not enough to justify changing flow in this PR. Classified as P3.

packages/desktop-electron/src/main/server.ts
Astro-Han: Left unchanged. Repeated `getStore()` calls hit a cached Map lookup and keep lazy initialization explicit at the call sites. This is aesthetic, not a correctness issue. Classified as P3.

packages/desktop-electron/src/renderer/startup-state.ts
Astro-Han: Fixed. `pushPendingDeepLinks` now mutates the pending array with `push(...urls)` instead of allocating a copy. Classified as P3.

PR #155, opened 2026-04-22
fix: close opencode upstream intake

### LINKED ISSUE #92: [Feature] Sync OpenCode upstream v1.14.20 intake

Astro-Han:
## Summary

- Add content-based attachment MIME sniffing so `read` promotes disguised image/PDF files while keeping SVG/text paths readable.
- Sync the remaining upstream runtime/plugin tail: GitHub Copilot Anthropic tool streaming opt-out, workspace adaptor env propagation, and skill discovery abort handling.
- Port the small UI primitive fixes for assistant parent lookup, review diff renderability, and tool accordion sticky offsets.

## Why

This is the final scoped #92 closeout after the session runtime lane in #140 landed. Desktop hardening remains tracked separately in #141/#152 and is not blocking this upstream intake closeout.

## Related Issue

Closes #92

## How To Verify

```bash
cd packages/opencode
bun test test/tool/read.test.ts test/session/message-v2.test.ts test/plugin/github-copilot-models.test.ts test/plugin/workspace-adaptor.test.ts
cd ../ui
bun test src/components/message-part-stale.test.ts src/components/session-turn-parent.test.ts src/components/session-review-diff-render.test.ts
cd ../..
bun turbo typecheck
bun turbo test:ci
git diff --check
```

## Screenshots or Recordings

Not applicable. The UI changes are narrow primitive guards and sticky offset constants covered by focused source tests plus the app build in `test:ci`.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Copilot behavior: correct handling for Anthropic-backed models and headers
  * Sticky accordion offset alignment in UI components
  * Session review diff visibility and per-file interaction states
  * Assistant threading now includes relevant follow-up assistant messages

* **Improvements**
  * Improved media/attachment detection via file-signature sniffing and stricter binary checks
  * Safer PDF/image handling with size caps and proper data-URL responses
  * Workspace creation now propagates runtime environment (auth, workspace id, OTEL) to spawned processes
  * File discovery respects cancellation for faster aborts

* **Tests**
  * New tests for Copilot model behavior, media sniffing, workspace env propagation, and session/message handling

### DIFF
--- a/packages/opencode/src/control-plane/types.ts
+++ b/packages/opencode/src/control-plane/types.ts
@@ -26,7 +26,8 @@ export type Target =
 
 export type Adaptor = {
   configure(input: WorkspaceInfo): WorkspaceInfo | Promise<WorkspaceInfo>
-  create(config: WorkspaceInfo, from?: WorkspaceInfo): Promise<void>
+  // from is reserved for future workspace copy flows; core does not pass it today.
+  create(config: WorkspaceInfo, env?: Record<string, string>, from?: WorkspaceInfo): Promise<void>
   remove(config: WorkspaceInfo): Promise<void>
   target(config: WorkspaceInfo): Target | Promise<Target>
 }
--- a/packages/opencode/src/control-plane/workspace.ts
+++ b/packages/opencode/src/control-plane/workspace.ts
@@ -11,6 +11,8 @@ import { Filesystem } from "@/util/filesystem"
 import { ProjectID } from "@/project/schema"
 import { Instance } from "@/project/instance"
 import { Plugin } from "@/plugin"
+import { Auth } from "@/auth"
+import { AppRuntime } from "@/effect/app-runtime"
 import { WorkspaceTable } from "./workspace.sql"
 import { getAdaptor, getBuiltinAdaptor, ownerKey } from "./adaptors"
 import { WorkspaceInfo } from "./types"
@@ -92,8 +94,8 @@ export namespace Workspace {
 
     const candidates = [
       ...new Set(
-        [input.hint, input.owner, projectWorktree, ...project.sandboxes].filter(
-          (value): value is string => Boolean(value),
+        [input.hint, input.owner, projectWorktree, ...project.sandboxes].filter((value): value is string =>
+          Boolean(value),
         ),
       ),
     ]
@@ -137,7 +139,9 @@ export namespace Workspace {
     throw lastError
   }
 
-  export async function resolveAdaptor(input: Pick<StoredInfo, "projectID" | "type" | "owner"> & { hint?: string | null }) {
+  export async function resolveAdaptor(
+    input: Pick<StoredInfo, "projectID" | "type" | "owner"> & { hint?: string | null },
+  ) {
     const hint =
       input.hint ??
       (() => {
@@ -200,7 +204,17 @@ export namespace Workspace {
         .run()
     })
 
-    await adaptor.create(config)
+    const env = Object.fromEntries(
+      Object.entries({
+        OPENCODE_AUTH_CONTENT: JSON.stringify(await AppRuntime.runPromise(Auth.Service.use((auth) => auth.all()))),
+        OPENCODE_WORKSPACE_ID: info.id,
+        OPENCODE_EXPERIMENTAL_WORKSPACES: "true",
+        OTEL_EXPORTER_OTLP_HEADERS: process.env.OTEL_EXPORTER_OTLP_HEADERS,
+        OTEL_EXPORTER_OTLP_ENDPOINT: process.env.OTEL_EXPORTER_OTLP_ENDPOINT,
+        OTEL_RESOURCE_ATTRIBUTES: process.env.OTEL_RESOURCE_ATTRIBUTES,
+      }).filter(([, value]) => value !== undefined),
+    ) as Record<string, string>
+    await adaptor.create(config, env)
 
     startSync({ space: info })
 
--- a/packages/opencode/src/plugin/github-copilot/copilot.ts
+++ b/packages/opencode/src/plugin/github-copilot/copilot.ts
@@ -10,6 +10,7 @@ import { MessageV2 } from "@/session/message-v2"
 const log = Log.create({ service: "plugin.copilot" })
 
 const CLIENT_ID = "Ov23li8tweQw6odWQebz"
+const COPILOT_ANTHROPIC_NPM = "@ai-sdk/anthropic"
 // Add a small safety buffer when polling to avoid hitting the server
 // slightly too early due to clock skew / timer drift.
 const OAUTH_POLLING_SAFETY_MARGIN_MS = 3000 // 3 seconds
@@ -329,16 +330,23 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
     },
     "chat.params": async (incoming, output) => {
       if (!incoming.model.providerID.includes("github-copilot")) return
+      const isCopilotAnthropic = incoming.model.api.npm === COPILOT_ANTHROPIC_NPM
 
       // Match github copilot cli, omit maxOutputTokens for gpt models
       if (incoming.model.api.id.includes("gpt")) {
         output.maxOutputTokens = undefined
       }
+
+      // Copilot's /v1/messages shim rejects the eager_input_streaming field.
+      if (isCopilotAnthropic) {
+        output.options.toolStreaming = false
+      }
     },
     "chat.headers": async (incoming, output) => {
       if (!incoming.model.providerID.includes("github-copilot")) return
+      const isCopilotAnthropic = incoming.model.api.npm === COPILOT_ANTHROPIC_NPM
 
-      if (incoming.model.api.npm === "@ai-sdk/anthropic") {
+      if (isCopilotAnthropic) {
         output.headers["anthropic-beta"] = "interleaved-thinking-2025-05-14"
       }
 
--- a/packages/opencode/src/tool/read.ts
+++ b/packages/opencode/src/tool/read.ts
@@ -1,7 +1,6 @@
 import z from "zod"
-import { Effect, Scope } from "effect"
+import { Effect, Option, Scope } from "effect"
 import { createReadStream } from "fs"
-import { open } from "fs/promises"
 import * as path from "path"
 import { createInterface } from "readline"
 import { Tool } from "./tool"
@@ -11,12 +10,16 @@ import DESCRIPTION from "./read.txt"
 import { Instance } from "../project/instance"
 import { assertExternalDirectoryEffect } from "./external-directory"
 import { Instruction } from "../session/instruction"
+import { isImageAttachment, isPdfAttachment, sniffAttachmentMime } from "../util/media"
 
 const DEFAULT_READ_LIMIT = 2000
 const MAX_LINE_LENGTH = 2000
 const MAX_LINE_SUFFIX = `... (line truncated to ${MAX_LINE_LENGTH} chars)`
 const MAX_BYTES = 50 * 1024
 const MAX_BYTES_LABEL = `${MAX_BYTES / 1024} KB`
+const MAX_ATTACHMENT_BYTES = 5 * 1024 * 1024
+const MAX_ATTACHMENT_BYTES_LABEL = `${MAX_ATTACHMENT_BYTES / 1024 / 1024} MB`
+const SAMPLE_BYTES = 4096
 
 const parameters = z.object({
   filePath: z.string().describe("The absolute path to the file or directory to read"),
@@ -75,6 +78,18 @@ export const ReadTool = Tool.define(
 
     const warm = Effect.fn("ReadTool.warm")(function* (filepath: string) {
       yield* lsp.touchFile(filepath, false).pipe(Effect.ignore, Effect.forkIn(scope))
+    })
+
+    const readSample = Effect.fn("ReadTool.sample")(function* (filepath: string, fileSize: number) {
+      if (fileSize === 0) return new Uint8Array()
+
+      return yield* Effect.scoped(
+        Effect.gen(function* () {
+          const file = yield* fs.open(filepath, { flag: "r" })
+          const bytes = yield* file.readAlloc(Math.min(SAMPLE_BYTES, fileSize))
+          return Option.getOrElse(bytes, () => new Uint8Array())
+        }),
+      )
     })
 
     const run = Effect.fn("ReadTool.execute")(function* (params: z.infer<typeof parameters>, ctx: Tool.Context) {
@@ -142,10 +157,23 @@ export const ReadTool = Tool.define(
 
       const loaded = yield* instruction.resolve(ctx.messages, filepath, ctx.messageID)
 
-      const mime = AppFileSystem.mimeType(filepath)
-      const isImage = mime.startsWith("image/") && mime !== "image/svg+xml" && mime !== "image/vnd.fastbidsheet"
-      const isPdf = mime === "application/pdf"
+      if (isBinaryByExt(filepath) && !shouldSniffBeforeBinaryExt(filepath)) {
+        return yield* Effect.fail(
+          new Error(`Cannot read binary file (extension: ${path.extname(filepath).toLowerCase()}): ${filepath}`),
+        )
+      }
+
+      const sample = yield* readSample(filepath, Number(stat.size))
+      const mime = sniffAttachmentMime(sample, AppFileSystem.mimeType(filepath))
+      const isImage = isImageAttachment(mime)
+      const isPdf = isPdfAttachment(mime)
       if (isImage || isPdf) {
+        if (Number(stat.size) > MAX_ATTACHMENT_BYTES) {
+          return yield* Effect.fail(
+            new Error(`Cannot read attachment larger than ${MAX_ATTACHMENT_BYTES_LABEL}: ${filepath}`),
+          )
+        }
+
         const msg = `${isImage ? "Image" : "PDF"} read successfully`
         return {
           title,
@@ -165,8 +193,8 @@ export const ReadTool = Tool.define(
         }
       }
 
-      if (yield* Effect.promise(() => isBinaryFile(filepath, Number(stat.size)))) {
-        return yield* Effect.fail(new Error(`Cannot read binary file: ${filepath}`))
+      if (isBinaryFile(filepath, sample)) {
+        return yield* Effect.fail(new Error(`Cannot read binary file (content inspection): ${filepath}`))
       }
 
       const file = yield* Effect.promise(() =>
@@ -262,7 +290,29 @@ async function lines(filepath: string, opts: { limit: number; offset: number }) 
   return { raw, count, cut, more, offset: opts.offset }
 }
 
-async function isBinaryFile(filepath: string, fileSize: number): Promise<boolean> {
+function isBinaryFile(filepath: string, sample: Uint8Array): boolean {
+  if (isBinaryByExt(filepath)) return true
+
+  if (sample.byteLength === 0) return false
+
+  let nonPrintableCount = 0
+  for (let i = 0; i < sample.byteLength; i++) {
+    if (sample[i] === 0) return true
+    if (sample[i] < 9 || (sample[i] > 13 && sample[i] < 32)) {
+      nonPrintableCount++
+    }
+  }
+  // If >30% non-printable characters, consider it binary
+  return nonPrintableCount / sample.byteLength > 0.3
+}
+
+function shouldSniffBeforeBinaryExt(filepath: string): boolean {
+  const ext = path.extname(filepath).toLowerCase()
+  // These common generic binary extensions may still contain renamed image/PDF attachments.
+  return ext === ".bin" || ext === ".dat"
+}
+
+function isBinaryByExt(filepath: string): boolean {
   const ext = path.extname(filepath).toLowerCase()
   // binary check for common non-text extensions
   switch (ext) {
@@ -296,28 +346,6 @@ async function isBinaryFile(filepath: string, fileSize: number): Promise<boolean
     case ".pyo":
       return true
     default:
-      break
-  }
-
-  if (fileSize === 0) return false
-
-  const fh = await open(filepath, "r")
-  try {
-    const sampleSize = Math.min(4096, fileSize)
-    const bytes = Buffer.alloc(sampleSize)
-    const result = await fh.read(bytes, 0, sampleSize, 0)
-    if (result.bytesRead === 0) return false
-
-    let nonPrintableCount = 0
-    for (let i = 0; i < result.bytesRead; i++) {
-      if (bytes[i] === 0) return true
-      if (bytes[i] < 9 || (bytes[i] > 13 && bytes[i] < 32)) {
-        nonPrintableCount++
-      }
-    }
-    // If >30% non-printable characters, consider it binary
-    return nonPrintableCount / result.bytesRead > 0.3
-  } finally {
-    await fh.close()
+      return false
   }
 }
--- a/packages/opencode/src/util/media.ts
+++ b/packages/opencode/src/util/media.ts
@@ -1,0 +1,49 @@
+const startsWith = (bytes: Uint8Array, prefix: number[]) =>
+  bytes.length >= prefix.length && prefix.every((value, index) => bytes[index] === value)
+const startsWithAt = (bytes: Uint8Array, offset: number, prefix: number[]) =>
+  bytes.length >= offset + prefix.length && prefix.every((value, index) => bytes[offset + index] === value)
+
+const ascii = (value: string) => [...value].map((char) => char.charCodeAt(0))
+const brand = (bytes: Uint8Array, offset: number) => String.fromCharCode(...bytes.slice(offset, offset + 4))
+const u32be = (bytes: Uint8Array, offset: number) =>
+  bytes.length >= offset + 4
+    ? ((bytes[offset] << 24) | (bytes[offset + 1] << 16) | (bytes[offset + 2] << 8) | bytes[offset + 3]) >>> 0
+    : 0
+
+export function isPdfAttachment(mime: string) {
+  return mime === "application/pdf"
+}
+
+export function isMedia(mime: string) {
+  return mime.startsWith("image/") || isPdfAttachment(mime)
+}
+
+export function isImageAttachment(mime: string) {
+  return mime.startsWith("image/") && mime !== "image/svg+xml" && mime !== "image/vnd.fastbidsheet"
+}
+
+export function sniffAttachmentMime(bytes: Uint8Array, fallback: string) {
+  if (startsWith(bytes, [0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a])) return "image/png"
+  if (startsWith(bytes, [0xff, 0xd8, 0xff])) return "image/jpeg"
+  if (startsWith(bytes, [0x47, 0x49, 0x46, 0x38])) return "image/gif"
+  if (startsWith(bytes, [0x42, 0x4d])) return "image/bmp"
+  if (startsWith(bytes, [0x49, 0x49, 0x2a, 0x00]) || startsWith(bytes, [0x4d, 0x4d, 0x00, 0x2a])) return "image/tiff"
+  if (startsWith(bytes, [0x25, 0x50, 0x44, 0x46, 0x2d])) return "application/pdf"
+  if (
+    startsWith(bytes, [0x52, 0x49, 0x46, 0x46]) &&
+    bytes.length >= 12 &&
+    startsWith(bytes.slice(8, 12), [0x57, 0x45, 0x42, 0x50])
+  )
+    return "image/webp"
+  if (startsWithAt(bytes, 4, ascii("ftyp"))) {
+    const boxSize = u32be(bytes, 0)
+    const limit = Math.min(boxSize > 0 ? boxSize : bytes.length, bytes.length)
+    const brands = []
+    for (let offset = 8; offset + 4 <= limit; offset += 4) {
+      brands.push(brand(bytes, offset))
+    }
+    if (brands.some((item) => item === "avif" || item === "avis")) return "image/avif"
+    if (brands.some((item) => ["heic", "heix", "hevc", "hevx", "mif1", "msf1"].includes(item))) return "image/heic"
+  }
+  return fallback
+}
--- a/packages/opencode/test/plugin/github-copilot-models.test.ts
+++ b/packages/opencode/test/plugin/github-copilot-models.test.ts
@@ -161,3 +161,138 @@ test("remaps fallback oauth model urls to the enterprise host", async () => {
   expect(models.claude.api.url).toBe("https://copilot-api.ghe.example.com")
   expect(models.claude.api.npm).toBe("@ai-sdk/github-copilot")
 })
+
+test("disables anthropic tool streaming for github copilot chat params", async () => {
+  const hooks = await CopilotAuthPlugin({
+    client: {} as never,
+    project: {} as never,
+    directory: "",
+    worktree: "",
+    experimental_workspace: {
+      register() {},
+    },
+    serverUrl: new URL("https://example.com"),
+    $: {} as never,
+  })
+
+  const output = { temperature: 0, topP: 1, topK: 0, options: {} as Record<string, unknown> }
+  await hooks["chat.params"]?.(
+    {
+      model: {
+        providerID: "github-copilot",
+        id: "claude",
+        api: { id: "claude-sonnet-4.5", npm: "@ai-sdk/anthropic" },
+      },
+    } as never,
+    output as never,
+  )
+
+  expect(output.options).toMatchObject({ toolStreaming: false })
+})
+
+test("keeps tool streaming untouched outside github copilot anthropic chat params", async () => {
+  const hooks = await CopilotAuthPlugin({
+    client: {} as never,
+    project: {} as never,
+    directory: "",
+    worktree: "",
+    experimental_workspace: {
+      register() {},
+    },
+    serverUrl: new URL("https://example.com"),
+    $: {} as never,
+  })
+
+  const copilotOpenAI = { temperature: 0, topP: 1, topK: 0, options: {} as Record<string, unknown> }
+  await hooks["chat.params"]?.(
+    {
+      model: {
+        providerID: "github-copilot",
+        id: "gpt",
+        api: { id: "gpt-5", npm: "@ai-sdk/openai" },
+      },
+    } as never,
+    copilotOpenAI as never,
+  )
+
+  const anthropic = { temperature: 0, topP: 1, topK: 0, options: {} as Record<string, unknown> }
+  await hooks["chat.params"]?.(
+    {
+      model: {
+        providerID: "anthropic",
+        id: "claude",
+        api: { id: "claude-sonnet-4.5", npm: "@ai-sdk/anthropic" },
+      },
+    } as never,
+    anthropic as never,
+  )
+
+  expect(copilotOpenAI.options).not.toHaveProperty("toolStreaming")
+  expect(anthropic.options).not.toHaveProperty("toolStreaming")
+})
+
+test("sets anthropic beta header only for github copilot anthropic chat headers", async () => {
+  const hooks = await CopilotAuthPlugin({
+    client: {
+      session: {
+        message: async () => {
+          throw new Error("skip")
+        },
+        get: async () => {
+          throw new Error("skip")
+        },
+      },
+    } as never,
+    project: {} as never,
+    directory: "",
+    worktree: "",
+    experimental_workspace: {
+      register() {},
+    },
+    serverUrl: new URL("https://example.com"),
+    $: {} as never,
+  })
+
+  const copilotAnthropic = { headers: {} as Record<string, string> }
+  await hooks["chat.headers"]?.(
+    {
+      model: {
+        providerID: "github-copilot",
+        id: "claude",
+        api: { id: "claude-sonnet-4.5", npm: "@ai-sdk/anthropic" },
+      },
+      message: { sessionID: "s", id: "m" },
+    } as never,
+    copilotAnthropic as never,
+  )
+
+  const copilotOpenAI = { headers: {} as Record<string, string> }
+  await hooks["chat.headers"]?.(
+    {
+      model: {
+        providerID: "github-copilot",
+        id: "gpt",
+        api: { id: "gpt-5", npm: "@ai-sdk/openai" },
+      },
+      message: { sessionID: "s", id: "m" },
+    } as never,
+    copilotOpenAI as never,
+  )
+
+  const anthropic = { headers: {} as Record<string, string> }
+  await hooks["chat.headers"]?.(
+    {
+      model: {
+        providerID: "anthropic",
+        id: "claude",
+        api: { id: "claude-sonnet-4.5", npm: "@ai-sdk/anthropic" },
+      },
+      message: { sessionID: "s", id: "m" },
+    } as never,
+    anthropic as never,
+  )
+
+  expect(copilotAnthropic.headers["anthropic-beta"]).toBe("interleaved-thinking-2025-05-14")
+  expect(copilotOpenAI.headers).not.toHaveProperty("anthropic-beta")
+  expect(anthropic.headers).not.toHaveProperty("anthropic-beta")
+})
--- a/packages/opencode/test/plugin/workspace-adaptor.test.ts
+++ b/packages/opencode/test/plugin/workspace-adaptor.test.ts
@@ -85,8 +85,8 @@ async function pluginProject() {
           "    configure(input) {",
           `      return { ...input, name: "plug", branch: "plug/main", directory: ${JSON.stringify(space)} }`,
           "    },",
-          "    async create(input) {",
-          `      await Bun.write(${JSON.stringify(mark)}, JSON.stringify(input))`,
+          "    async create(input, env) {",
+          `      await Bun.write(${JSON.stringify(mark)}, JSON.stringify({ input, env }))`,
           "    },",
           "    async remove() {},",
           "    target(input) {",
@@ -141,13 +141,25 @@ describe("plugin.workspace", () => {
     expect(info.branch).toBe("plug/main")
     expect(info.directory).toBe(tmp.extra.space)
     expect(info.extra).toEqual({ key: "value" })
-    expect(JSON.parse(await Bun.file(tmp.extra.mark).text())).toMatchObject({
+    const created = JSON.parse(await Bun.file(tmp.extra.mark).text())
+    expect(created.input).toMatchObject({
       type: tmp.extra.type,
       name: "plug",
       branch: "plug/main",
       directory: tmp.extra.space,
       extra: { key: "value" },
     })
+    expect(created.env.OPENCODE_WORKSPACE_ID).toBe(info.id)
+    expect(created.env.OPENCODE_EXPERIMENTAL_WORKSPACES).toBe("true")
+    const otelKeys = ["OTEL_EXPORTER_OTLP_HEADERS", "OTEL_EXPORTER_OTLP_ENDPOINT", "OTEL_RESOURCE_ATTRIBUTES"] as const
+    for (const key of otelKeys) {
+      const expected = process.env[key]
+      if (expected === undefined) expect(created.env).not.toHaveProperty(key)
+      else expect(created.env[key]).toBe(expected)
+    }
+    const auth = JSON.parse(created.env.OPENCODE_AUTH_CONTENT)
+    expect(typeof auth).toBe("object")
+    expect(auth).not.toBeNull()
     await waitFor(() => {
       const status = Workspace.status().find((item) => item.workspaceID === info.id)
       return status !== undefined && status.status !== "connecting"
--- a/packages/opencode/test/tool/read.test.ts
+++ b/packages/opencode/test/tool/read.test.ts
@@ -17,6 +17,12 @@ import { provideInstance, tmpdirScoped } from "../fixture/fixture"
 import { testEffect } from "../lib/effect"
 
 const FIXTURES_DIR = path.join(import.meta.dir, "fixtures")
+const ftyp = (...brands: string[]) => {
+  const content = Buffer.concat([Buffer.from("ftyp"), ...brands.map((item) => Buffer.from(item))])
+  const size = Buffer.alloc(4)
+  size.writeUInt32BE(size.length + content.length)
+  return Buffer.concat([size, content])
+}
 
 afterEach(async () => {
   await Instance.disposeAll()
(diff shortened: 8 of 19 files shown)


packages/opencode/src/control-plane/workspace.ts
```diff
@@ -200,7 +202,15 @@ export namespace Workspace {
         .run()
     })
 
-    await adaptor.create(config)
+    const env = {
+      OPENCODE_AUTH_CONTENT: JSON.stringify(await AppRuntime.runPromise(Auth.Service.use((auth) => auth.all()))),
```
Astro-Han: **[P1][security-ish]** `OPENCODE_AUTH_CONTENT` serializes **every** provider's auth blob into an env var handed to whatever adaptor is registered — including third-party plugin adaptors the user may have installed for unrelated reasons. There is no per-provider scoping and no opt-in from the adaptor side; a plugin author who only wanted to implement `create(config)` now transparently receives every credential.

At minimum: document this on the `WorkspaceAdaptor.create` contract in `packages/plugin/src/index.ts` so plugin authors know what they're being handed. Ideally, narrow the blob to the auth entries relevant to `config` (or the currently selected provider), or require adaptors to opt in via a separate parameter shape. This keeps the blast radius of a compromised/curious plugin bounded.

packages/opencode/src/control-plane/workspace.ts
```diff
@@ -200,7 +202,15 @@ export namespace Workspace {
...
+      OPENCODE_WORKSPACE_ID: config.id,
+      OPENCODE_EXPERIMENTAL_WORKSPACES: "true",
+      OTEL_EXPORTER_OTLP_HEADERS: process.env.OTEL_EXPORTER_OTLP_HEADERS,
+      OTEL_EXPORTER_OTLP_ENDPOINT: process.env.OTEL_EXPORTER_OTLP_ENDPOINT,
+      OTEL_RESOURCE_ATTRIBUTES: process.env.OTEL_RESOURCE_ATTRIBUTES,
```
Astro-Han: **[P2]** `Record<string, string | undefined>` leaks OTEL env variables through even when they're `undefined` in the parent process. Most adaptors will forward this straight to `child_process.spawn({ env })`, and Node's behavior for `undefined` env values is version-dependent (some stringify to `"undefined"`, some drop the key). Cleaner to filter undefineds at the boundary:

```ts
const env = Object.fromEntries(
  Object.entries({
    OPENCODE_AUTH_CONTENT: ...,
    OPENCODE_WORKSPACE_ID: config.id,
    OPENCODE_EXPERIMENTAL_WORKSPACES: "true",
    OTEL_EXPORTER_OTLP_HEADERS: process.env.OTEL_EXPORTER_OTLP_HEADERS,
    OTEL_EXPORTER_OTLP_ENDPOINT: process.env.OTEL_EXPORTER_OTLP_ENDPOINT,
    OTEL_RESOURCE_ATTRIBUTES: process.env.OTEL_RESOURCE_ATTRIBUTES,
  }).filter(([, v]) => v !== undefined),
) as Record<string, string>
```

This also lets the plugin API type `env` as `Record<string, string>` — a nicer contract for adaptor authors.

packages/plugin/src/index.ts
```diff
@@ -49,7 +49,7 @@ export type WorkspaceAdaptor = {
   name: string
   description: string
   configure(config: WorkspaceInfo): WorkspaceInfo | Promise<WorkspaceInfo>
-  create(config: WorkspaceInfo, from?: WorkspaceInfo): Promise<void>
+  create(config: WorkspaceInfo, env: Record<string, string | undefined>, from?: WorkspaceInfo): Promise<void>
```
Astro-Han: **[P2][API-compat]** This is a hard-breaking TypeScript signature change on the exported `WorkspaceAdaptor.create`. Any plugin consumer that previously typed their adaptor as `create(config)` will fail `strict`/`noImplicitAny` compilation against the new `@opencode-ai/plugin` release — note that even the in-repo `packages/plugin/src/example-workspace.ts` still uses the one-arg form and happens to work only because TS allows omitting trailing parameters at the call site.

Either: (a) call it out in release notes for the plugin package, (b) make `env` optional (`env?:`) so existing adaptors keep compiling, or (c) ship a migration note in the plugin README.

packages/ui/src/components/session-review.tsx
```diff
@@ -493,16 +493,18 @@ export const SessionReview = (props: SessionReviewProps) => {
                       commentsUi.onLineSelectionEnd(range)
                     }
 
+                    const diffCanRender = () => diff.additions !== 0 || diff.deletions !== 0
+
                     return (
                       <Accordion.Item
-                        value={file}
+                        value={diffCanRender() ? file : null!}
```
Astro-Han: **[P1]** `value={diffCanRender() ? file : null!}` uses the non-null-assertion operator to lie to TypeScript. At runtime Kobalte's `Accordion.Item` receives `null` as its value — and if a review contains two zero-diff files (e.g. both renamed, or both metadata-only), **they now share the same `null` key** and the accordion's internal selection map can't tell them apart. A `multiple`-mode accordion with duplicate values tends to open/close them in lockstep.

Safer to keep the unique `file` value on every item and rely on `disabled={!diffCanRender()}` alone (you already do, two lines below), or split the zero-diff case into a plain non-accordion row rendered as a sibling.

packages/ui/src/components/session-review.tsx
```diff
@@ -493,16 +493,18 @@ export const SessionReview = (props: SessionReviewProps) => {
...
                         id={diffId(file)}
                         data-file={file}
                         data-slot="session-review-accordion-item"
                         data-selected={props.focusedFile === file ? "" : undefined}
                       >
                         <StickyAccordionHeader>
-                          <Accordion.Trigger>
+                          <Accordion.Trigger disabled={!diffCanRender()} class="cursor-default">
```
Astro-Han: **[P2]** `disabled={!diffCanRender()} class="cursor-default"` gives off a mixed signal: the control is disabled, but the cursor says "nothing special, go ahead." Native disabled state is `cursor: not-allowed`; overriding it to `default` usually means the author didn't actually want `disabled` semantics (e.g. they want the row to look normal but be inert). Either:

- Drop the cursor override and let the standard disabled cursor show, or
- Replace `disabled` with `aria-disabled="true"` + a no-op click handler, keeping the appearance you want without confusing assistive tech.

Minor, but worth picking one.

packages/ui/src/components/session-turn.tsx
```diff
@@ -267,17 +267,9 @@ export function SessionTurn(
...
-      for (let i = index + 1; i < messages.length; i++) {
-        const item = messages[i]
-        if (!item) continue
-        if (item.role === "user") break
-        if (item.role === "assistant" && item.parentID === msg.id) result.push(item as AssistantMessage)
-      }
-      return result
+      if (messageIndex() < 0) return emptyAssistant
+
+      return messages.filter((item): item is AssistantMessage => item.role === "assistant" && item.parentID === msg.id)
```
Astro-Han: **[P2][perf/correctness trade-off]** The old scan was `O(turn-length)` with a hard stop at the next `user` message. This rewrite is `O(N)` over the entire `messages` array, per `createMemo` recomputation, across every turn — for a long session (say, 300 messages across 10 turns) the sum of work is quadratic in turn count.

The fix is correct in intent (the old boundary missed valid `parentID` matches). But two questions worth confirming before accepting the unbounded scan:

1. Is `parentID` guaranteed to point to the `msg.id` that is *earlier* in the list? If so, starting the scan at `messageIndex() + 1` is still correct and keeps the optimization.
2. If assistant messages *can* arrive after the next user turn, under what condition? A short comment here would save the next maintainer from re-deriving this on sight of a hot-path profile.

One-line fix if (1) holds:

```ts
return messages.slice(messageIndex() + 1).filter(
  (item): item is AssistantMessage => item.role === "assistant" && item.parentID === msg.id,
)
```

packages/opencode/src/tool/read.ts
```diff
@@ -142,9 +155,10 @@ export const ReadTool = Tool.define(
 
       const loaded = yield* instruction.resolve(ctx.messages, filepath, ctx.messageID)
 
-      const mime = AppFileSystem.mimeType(filepath)
-      const isImage = mime.startsWith("image/") && mime !== "image/svg+xml" && mime !== "image/vnd.fastbidsheet"
-      const isPdf = mime === "application/pdf"
+      const sample = yield* readSample(filepath, Number(stat.size))
```
Astro-Han, line 166: **[P2]** `readSample` now runs unconditionally, *before* the extension-based binary short-circuit inside `isBinaryFile`. For any file with a known-binary extension (`.zip`, `.exe`, `.so`, `.class`, etc.), we still pay 4 KB of disk I/O for a sniff whose result is never consumed — the `isBinary` path returns early, and the attachment path is unreachable because none of the magic-byte signatures overlap with those extensions.

Either:

- Move the extension switch out of `isBinaryFile` into a separate `isBinaryByExt(filepath)` and short-circuit before `readSample`, or
- Accept the minor regression and add a comment that the sample is unconditional for simplicity.

Not worth blocking on, but cheap to fix.

packages/opencode/src/plugin/github-copilot/copilot.ts
```diff
@@ -334,6 +334,9 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
       if (incoming.model.api.id.includes("gpt")) {
         output.maxOutputTokens = undefined
       }
+      if (incoming.model.api.npm === "@ai-sdk/anthropic") {
+        output.options.toolStreaming = false
```
Astro-Han, line 342: **[P2]** `@ai-sdk/anthropic` is hardcoded in a plain branch right next to another `incoming.model.api.npm === "@ai-sdk/anthropic"` check in `chat.headers` (line 344). If a future Copilot npm adapter also needs tool-streaming disabled (history suggests `@ai-sdk/google` and friends have had similar quirks), this will silently regress.

Two cheap hardens:

1. Lift a `const isCopilotAnthropic = incoming.model.api.npm === "@ai-sdk/anthropic"` to the top and reuse it in both hooks.
2. Add a one-line anchor comment explaining **why** tool streaming is disabled (Copilot proxy stripping stream chunks? SDK bug?) — the next maintainer will thank you when deciding whether to remove the workaround.

packages/opencode/src/util/media.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (startsWith(bytes, [0xff, 0xd8, 0xff])) return "image/jpeg"
+  if (startsWith(bytes, [0x47, 0x49, 0x46, 0x38])) return "image/gif"
+  if (startsWith(bytes, [0x42, 0x4d])) return "image/bmp"
+  if (startsWith(bytes, [0x25, 0x50, 0x44, 0x46, 0x2d])) return "application/pdf"
+  if (
+    startsWith(bytes, [0x52, 0x49, 0x46, 0x46]) &&
+    bytes.length >= 12 &&
+    startsWith(bytes.slice(8, 12), [0x57, 0x45, 0x42, 0x50])
+  )
+    return "image/webp"
```
Astro-Han, line 37: **[P3]** The PR stated goal is to promote "disguised image/PDF files." The signature list covers PNG/JPEG/GIF/BMP/PDF/WebP, but skips three formats that are extremely common on modern devices:

- **TIFF**: `49 49 2A 00` (little-endian) / `4D 4D 00 2A` (big-endian)
- **HEIC / HEIF**: `....ftypheic` / `ftyphevc` at offset 4
- **AVIF**: `....ftypavif` at offset 4

If a user drags a renamed `.heic` off an iPhone today, the sniff returns the fallback mime (often `application/octet-stream`), the attachment path doesn't trigger, and the file hits `isBinaryFile` and fails. Either add these signatures or note the supported set in a comment so the gap is explicit.

packages/opencode/test/plugin/workspace-adaptor.test.ts
```diff
@@ -141,13 +141,17 @@ describe("plugin.workspace", () => {
...
+    expect(created.input).toMatchObject({
       type: tmp.extra.type,
       name: "plug",
       branch: "plug/main",
       directory: tmp.extra.space,
       extra: { key: "value" },
     })
+    expect(created.env.OPENCODE_WORKSPACE_ID).toBe(info.id)
+    expect(created.env.OPENCODE_EXPERIMENTAL_WORKSPACES).toBe("true")
+    expect(JSON.parse(created.env.OPENCODE_AUTH_CONTENT)).toEqual({})
```
Astro-Han: **[P3]** `toEqual({})` asserts the test runtime has **zero** auth entries. The moment another test helper (or a future `beforeAll`) seeds an auth fixture, this explodes without telling the reader what the meaningful contract is. Prefer a shape assertion:

```ts
const auth = JSON.parse(created.env.OPENCODE_AUTH_CONTENT)
expect(typeof auth).toBe("object")
expect(auth).not.toBeNull()
```

Or if the intent is "env-propagated auth is whatever `auth.all()` returned at the time," call `Auth.Service.use(a => a.all())` inside the test and compare against that.

packages/ui/src/components/session-turn-parent.test.ts
```diff
@@ -0,0 +1,12 @@
+import { expect, test } from "bun:test"
```
Astro-Han, line 1: **[P3][test-quality]** This (and `session-review-diff-render.test.ts`, and the new case added to `message-part-stale.test.ts`) asserts against `readFileSync`'d source strings. That's fine as a regression pin for a specific refactor, but each new source-match test compounds the tax: any stylistic refactor (line-wrapping, variable renames, even `const` → `let`) breaks the test without flagging a real regression, and no real regression in the `parentID` lookup logic would be caught by a substring check.

Not blocking — the pattern is already established in this directory — but a thin render test via `@solidjs/testing-library` that mounts `SessionTurn` with a crafted `allMessages()` and asserts the returned `assistantMessages` signal would actually defend the invariant.

Astro-Han (commented):
Nitpicky review of the upstream-intake closeout. Findings labeled P0–P3; nothing looks blocking for the stated scope, but a few of the P1/P2 items (auth env fanout, `null!` on Accordion.Item, and the unbounded turn scan) feel worth a second look before merge.

- P1: 3
- P2: 4
- P3: 3

packages/opencode/src/control-plane/workspace.ts
Astro-Han: Documented the workspace adaptor env contract in packages/plugin/src/index.ts, including that OPENCODE_AUTH_CONTENT is a serialized provider auth blob and must be treated as secret. I did not narrow the auth blob in this PR because WorkspaceInfo does not carry provider selection, so that would be a separate plugin contract change rather than a safe #92 tail fix.

packages/opencode/src/control-plane/workspace.ts
Astro-Han: Fixed. The env object now filters undefined values at the boundary and is typed as Record<string, string>.

packages/plugin/src/index.ts
Astro-Han: Fixed. The plugin API now makes env optional and documents the secret-bearing env contract, so existing one-argument adaptors continue to type cleanly.

packages/ui/src/components/session-review.tsx
Astro-Han: Fixed. Accordion.Item now keeps value={file} for every row and relies on the disabled trigger for non-renderable diffs, avoiding duplicate null values.

packages/ui/src/components/session-review.tsx
Astro-Han: Fixed. Removed the cursor override and kept native disabled semantics.

packages/ui/src/components/session-turn.tsx
Astro-Han: Fixed. The lookup now starts at messageIndex() + 1 and still filters by parentID, with a short comment explaining why it no longer stops at the next user message.

packages/opencode/src/tool/read.ts
Astro-Han: Fixed. Split extension-based binary detection out and short-circuit known binary extensions before sampling, while still allowing .bin and .dat to be sniffed for disguised image/PDF attachments before rejecting them as binary.

packages/opencode/src/plugin/github-copilot/copilot.ts
Astro-Han: Fixed. Reused an isCopilotAnthropic local in both hooks and added the upstream reason: Copilot's messages shim rejects the eager_input_streaming field injected by @ai-sdk/anthropic tool streaming.

packages/opencode/src/util/media.ts
Astro-Han: Fixed. Added TIFF, HEIC, and AVIF content sniffing plus coverage for renamed .bin/.dat files.

packages/opencode/test/plugin/workspace-adaptor.test.ts
Astro-Han: Fixed. The test now checks OPENCODE_AUTH_CONTENT as a non-null object shape instead of assuming the auth store is empty, and verifies exported env values are strings.

packages/ui/src/components/session-turn-parent.test.ts
Astro-Han: Leaving this as a source-level regression pin for this PR. This directory already uses that lightweight pattern, and adding a Solid render harness would be larger than the #92 closeout tail. I did tighten the source assertion to cover the new slice-based parentID invariant.

Astro-Han:
@coderabbitai review

packages/opencode/src/util/media.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (startsWithAt(bytes, 4, ascii("ftyp"))) {
+    if (startsWithAt(bytes, 8, ascii("avif"))) return "image/avif"
+    if (startsWithAt(bytes, 8, ascii("heic")) || startsWithAt(bytes, 8, ascii("hevc"))) return "image/heic"
+  }
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The ftyp sniff now reads the box size and scans all 4-byte brand entries in the box, so compatible-brand-only HEIF/AVIF files are detected. Added regression coverage for isom major brand files that advertise heix, hevx, or avis only in the compatible brand list.

packages/opencode/src/util/media.ts
```diff
@@ -0,0 +1,48 @@
+const startsWith = (bytes: Uint8Array, prefix: number[]) => prefix.every((value, index) => bytes[index] === value)
```
Astro-Han: test comment - please ignore

packages/opencode/src/control-plane/types.ts
```diff
@@ -26,7 +26,7 @@ export type Target =
 
 export type Adaptor = {
   configure(input: WorkspaceInfo): WorkspaceInfo | Promise<WorkspaceInfo>
-  create(config: WorkspaceInfo, from?: WorkspaceInfo): Promise<void>
+  create(config: WorkspaceInfo, env?: Record<string, string>, from?: WorkspaceInfo): Promise<void>
```
Astro-Han, line 30: **[P2][dead param]** The `from?` parameter on `Adaptor.create` is never wired up — `workspace.ts` calls `adaptor.create(config, env)` with no third argument. This was dead code before this PR and still is. Either remove `from?` from both `Adaptor` and `WorkspaceAdaptor` (they're in lockstep), or add a `// TODO: reserved for future use` so adaptor authors don't wonder what to do with it.

packages/opencode/src/control-plane/workspace.ts
```diff
@@ -200,7 +204,17 @@ export namespace Workspace {
         .run()
     })
 
-    await adaptor.create(config)
+    const env = Object.fromEntries(
+      Object.entries({
+        OPENCODE_AUTH_CONTENT: JSON.stringify(await AppRuntime.runPromise(Auth.Service.use((auth) => auth.all()))),
```
Astro-Han, line 209: **[P1][security re-raise]** The previous review noted that `OPENCODE_AUTH_CONTENT` serializes *all* provider credentials into an env var. The fix was doc-only (JSDoc on `WorkspaceAdaptor.create`). Re-raising because the blast radius is credential exposure to untrusted plugin code — documentation is advisory, not enforceable. Consider scoping to only the auth entries relevant to `config.type`'s provider, or logging a one-time warning when a non-builtin adaptor receives `OPENCODE_AUTH_CONTENT`. At minimum, track as a follow-up issue so it doesn't get lost after merge.

packages/opencode/src/util/media.ts
Astro-Han: **[P1][crash risk]** `startsWith` does not bounds-check `bytes.length` against `prefix.length`. `startsWithAt` (line 2) checks `bytes.length >= offset + prefix.length` but `startsWith` doesn't. If a future caller passes a sample shorter than the prefix, `bytes[index]` reads `undefined` and silently returns `false` — no error, just a wrong result. The asymmetry is a landmine. Fix:

```ts
const startsWith = (bytes: Uint8Array, prefix: number[]) =>
  bytes.length >= prefix.length && prefix.every((value, index) => bytes[index] === value)
```

packages/opencode/src/util/media.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (startsWith(bytes, [0x25, 0x50, 0x44, 0x46, 0x2d])) return "application/pdf"
+  if (
+    startsWith(bytes, [0x52, 0x49, 0x46, 0x46]) &&
+    bytes.length >= 12 &&
+    startsWith(bytes.slice(8, 12), [0x57, 0x45, 0x42, 0x50])
+  )
+    return "image/webp"
+  if (startsWithAt(bytes, 4, ascii("ftyp"))) {
+    const boxSize = u32be(bytes, 0)
+    const limit = Math.min(boxSize || bytes.length, bytes.length)
```
Astro-Han: **[P2][subtle bug]** `boxSize || bytes.length` — if `u32be` returns `0` (sample too short for the box-size field), `0 || bytes.length` falls through to `bytes.length`, which happens to be correct. But a genuine `boxSize` of 0 (degenerate but per ISO 14496-12 not forbidden) would also fall through, scanning the entire sample as brands. Use `boxSize > 0 ? boxSize : bytes.length` or add a comment explaining the `0` fallback.

packages/opencode/src/plugin/github-copilot/copilot.ts
```diff
@@ -329,16 +329,23 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
...
       }
+
+      // Copilot's /v1/messages shim rejects the eager_input_streaming field.
+      if (isCopilotAnthropic) {
+        output.options.toolStreaming = false
+      }
     },
     "chat.headers": async (incoming, output) => {
       if (!incoming.model.providerID.includes("github-copilot")) return
+      const isCopilotAnthropic = incoming.model.api.npm === "@ai-sdk/anthropic"
```
Astro-Han: **[P3][DRY]** `isCopilotAnthropic` is computed independently in `chat.params` (line 332) and `chat.headers` (line 346). The string `"@ai-sdk/anthropic"` is duplicated. If Copilot adds a second Anthropic npm adapter, both hooks need updating. Extract a `const COPILOT_ANTHROPIC_NPM = "@ai-sdk/anthropic"` at the top of the plugin.

packages/opencode/test/plugin/github-copilot-models.test.ts
```diff
@@ -161,3 +161,31 @@ test("remaps fallback oauth model urls to the enterprise host", async () => {
   expect(models.claude.api.url).toBe("https://copilot-api.ghe.example.com")
   expect(models.claude.api.npm).toBe("@ai-sdk/github-copilot")
 })
+
+test("disables anthropic tool streaming for github copilot chat params", async () => {
```
Astro-Han, line 165: **[P2][test gap]** This test only verifies the positive case (Copilot + Anthropic -> toolStreaming: false). Missing:

1. **Non-Anthropic Copilot model** (e.g. `api.npm = "@ai-sdk/openai"`) should *not* set `toolStreaming: false`.
2. **Non-Copilot Anthropic model** (e.g. `providerID: "anthropic"`) should also *not* set it.

At least one negative case is needed to lock in the AND condition.

packages/opencode/test/tool/read.test.ts
```diff
@@ -421,6 +427,80 @@ describe("tool.read truncation", () => {
...
+      expect(result.output).toContain("PDF read successfully")
+      expect(result.attachments?.[0]?.mime).toBe("application/pdf")
+      expect(result.attachments?.[0]?.url).toStartWith("data:application/pdf;base64,")
+    }),
+  )
+
+  it.live("detects modern image attachment media from file contents", () =>
+    Effect.gen(function* () {
+      const dir = yield* tmpdirScoped()
+      yield* put(path.join(dir, "photo.bin"), Buffer.from([0x49, 0x49, 0x2a, 0x00]))
```
Astro-Han: **[P2][test gap]** The TIFF fixture is only 4 bytes. Too small to exercise the full attachment path (base64 encoding, data URL construction). Pad to 8+ bytes to exercise `readSample`'s `Math.min(SAMPLE_BYTES, fileSize)` more realistically.

packages/opencode/test/plugin/workspace-adaptor.test.ts
```diff
@@ -141,13 +141,22 @@ describe("plugin.workspace", () => {
     expect(info.branch).toBe("plug/main")
     expect(info.directory).toBe(tmp.extra.space)
     expect(info.extra).toEqual({ key: "value" })
-    expect(JSON.parse(await Bun.file(tmp.extra.mark).text())).toMatchObject({
+    const created = JSON.parse(await Bun.file(tmp.extra.mark).text())
+    expect(created.input).toMatchObject({
       type: tmp.extra.type,
       name: "plug",
       branch: "plug/main",
```
Astro-Han, line 148: **[P3][fragile]** `for (const value of Object.values(created.env))` asserts all env values are strings. But `OPENCODE_AUTH_CONTENT` is `JSON.stringify(...)` by construction. The interesting invariant is that OTEL keys are *absent* when not set. Consider asserting key absence instead, or remove the noise.

packages/opencode/src/tool/read.ts
```diff
@@ -142,9 +155,14 @@ export const ReadTool = Tool.define(
 
       const loaded = yield* instruction.resolve(ctx.messages, filepath, ctx.messageID)
 
-      const mime = AppFileSystem.mimeType(filepath)
-      const isImage = mime.startsWith("image/") && mime !== "image/svg+xml" && mime !== "image/vnd.fastbidsheet"
-      const isPdf = mime === "application/pdf"
+      if (isBinaryByExt(filepath) && !shouldSniffBeforeBinaryExt(filepath)) {
```
Astro-Han, line 160: **[P2][UX]** Before this PR, a binary-extension file would still get content-sampled by `isBinaryFile`'s byte heuristic. Now `isBinaryByExt` short-circuits with an error. The message `"Cannot read binary file"` is identical to the content-detected one (line 186). A user can't tell whether rejection was due to extension or content. Consider differentiating: `"Cannot read binary file (extension: .zip)"` vs `"Cannot read binary file (content inspection)"`.

packages/opencode/src/tool/read.ts
```diff
@@ -142,9 +155,14 @@ export const ReadTool = Tool.define(
...
+        return yield* Effect.fail(new Error(`Cannot read binary file: ${filepath}`))
+      }
+
+      const sample = yield* readSample(filepath, Number(stat.size))
```
Astro-Han, line 166: **[P1][memory]** When a sniffed image/PDF is detected, `readSample` just read 4 KB into `sample`, then `fs.readFile(filepath)` at line 180 reads the *entire file* to base64-encode it. For large files (e.g. a 20 MB RAW photo renamed to `.bin`), the full content is held multiple times in the JS heap (readSample Uint8Array + readFile Buffer + base64 string). The `sample` variable is unused after `sniffAttachmentMime` returns — consider scoping it for GC, or add a comment.

Also: **no file-size cap** on image/PDF attachments. A 100 MB PDF will be eagerly base64-encoded into a data URL, likely OOMing. Consider adding `MAX_ATTACHMENT_BYTES` before the readFile at line 180.

packages/opencode/src/tool/read.ts
```diff
@@ -262,7 +280,28 @@ async function lines(filepath: string, opts: { limit: number; offset: number })
...
+    if (sample[i] === 0) return true
+    if (sample[i] < 9 || (sample[i] > 13 && sample[i] < 32)) {
+      nonPrintableCount++
+    }
+  }
+  // If >30% non-printable characters, consider it binary
+  return nonPrintableCount / sample.byteLength > 0.3
+}
+
+function shouldSniffBeforeBinaryExt(filepath: string): boolean {
```
Astro-Han, line 309: **[P3][coupling]** `shouldSniffBeforeBinaryExt` allows `.bin` and `.dat` to bypass the extension short-circuit. But those extensions are *also* in `isBinaryByExt`'s switch (lines 337-338). The two lists must stay in sync: if someone adds to `shouldSniffBeforeBinaryExt` but forgets `isBinaryByExt`, the sniff-override becomes dead code. A comment explaining the coupling would help.

packages/opencode/src/util/media.ts
Astro-Han: Ignored as requested.

packages/opencode/src/control-plane/types.ts
Astro-Han: Handled with a reserved-for-future-use comment on both the internal Adaptor.create and public WorkspaceAdaptor.create contracts. I kept the parameter for compatibility instead of removing it in this PR.

packages/opencode/src/control-plane/workspace.ts
Astro-Han: Not changing auth scoping in #155 because WorkspaceInfo does not carry provider selection, and narrowing or opt-in would be a separate plugin contract design. Created follow-up #173 to track scoped or opt-in OPENCODE_AUTH_CONTENT propagation.

packages/opencode/src/util/media.ts
Astro-Han: Fixed. startsWith now checks bytes.length before comparing the prefix.

packages/opencode/src/util/media.ts
Astro-Han: Fixed. The ftyp box limit now uses boxSize > 0 ? boxSize : bytes.length so the zero-size fallback is explicit.

packages/opencode/src/plugin/github-copilot/copilot.ts
Astro-Han: Fixed. Extracted COPILOT_ANTHROPIC_NPM and reused it in both hooks.

packages/opencode/test/plugin/github-copilot-models.test.ts
Astro-Han: Fixed. Added negative coverage for GitHub Copilot with a non-Anthropic adapter and non-Copilot Anthropic, both asserting toolStreaming is untouched.

packages/opencode/test/tool/read.test.ts
Astro-Han: Fixed. The TIFF fixture is now padded beyond the 4-byte signature.

packages/opencode/test/plugin/workspace-adaptor.test.ts
Astro-Han: Fixed. Replaced the broad Object.values assertion with explicit absence checks for unset OTEL env keys.

packages/opencode/src/tool/read.ts
Astro-Han: Fixed. Extension-based and content-inspection binary rejections now produce distinct messages, with test coverage for both paths.

packages/opencode/src/tool/read.ts
Astro-Han: Fixed. Sniffed media attachments now have a 5 MB cap before eager base64 encoding, with coverage for oversized renamed attachments.

packages/opencode/src/tool/read.ts
Astro-Han: Fixed. Added a coupling comment on the .bin/.dat sniff override so future changes keep it aligned with extension-based binary detection.

packages/opencode/test/plugin/workspace-adaptor.test.ts
```diff
@@ -141,13 +141,22 @@ describe("plugin.workspace", () => {
...
       directory: tmp.extra.space,
       extra: { key: "value" },
     })
+    expect(created.env.OPENCODE_WORKSPACE_ID).toBe(info.id)
+    expect(created.env.OPENCODE_EXPERIMENTAL_WORKSPACES).toBe("true")
+    expect(created.env).not.toHaveProperty("OTEL_EXPORTER_OTLP_HEADERS")
+    expect(created.env).not.toHaveProperty("OTEL_EXPORTER_OTLP_ENDPOINT")
+    expect(created.env).not.toHaveProperty("OTEL_RESOURCE_ATTRIBUTES")
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The OTEL assertions now compare against the current process.env value for each forwarded key, so the test passes whether the runner exports OTEL variables or leaves them unset.

Astro-Han:
Handled the CodeRabbit outside-diff review body items from the latest review:

- Fixed the session review renderability gate. Media diffs now count as renderable, expand-all only opens renderable files, and controlled open state cannot mount content for metadata-only rows.
- Left the repeated OPENCODE_AUTH_CONTENT scoping item out of #155 by design. That needs a plugin contract change, and is tracked in #173.

packages/opencode/src/control-plane/workspace.ts
```diff
@@ -200,7 +204,17 @@ export namespace Workspace {
...
+        OPENCODE_WORKSPACE_ID: config.id,
+        OPENCODE_EXPERIMENTAL_WORKSPACES: "true",
```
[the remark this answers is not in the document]
Astro-Han: Fixed. OPENCODE_WORKSPACE_ID now comes from the persisted workspace record via info.id rather than the mutable configure result.

packages/opencode/test/plugin/github-copilot-models.test.ts
```diff
@@ -161,3 +161,72 @@ test("remaps fallback oauth model urls to the enterprise host", async () => {
...
+        id: "claude",
+        api: { id: "claude-sonnet-4.5", npm: "@ai-sdk/anthropic" },
+      },
+    } as never,
+    anthropic as never,
+  )
+
+  expect(copilotOpenAI.options).not.toHaveProperty("toolStreaming")
+  expect(anthropic.options).not.toHaveProperty("toolStreaming")
+})
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Added chat.headers coverage for Copilot Anthropic setting anthropic-beta, plus negative checks for Copilot OpenAI and non-Copilot Anthropic.

packages/ui/src/components/session-review.tsx
```diff
@@ -261,8 +263,14 @@ export const SessionReview = (props: SessionReviewProps) => {
...
+  const renderableFiles = createMemo(() =>
+    items()
+      .filter(canRenderDiff)
+      .map((diff) => diff.file),
+  )
+
   const handleExpandOrCollapseAll = () => {
-    const next = open().length > 0 ? [] : files()
+    const next = open().length > 0 ? [] : renderableFiles()
     handleChange(next)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The bulk toggle and label now use hasOpenRenderableFiles(), so non-renderable open entries do not force the button into collapse-all mode.

Astro-Han:
@coderabbitai review

Astro-Han:
@coderabbitai review

PR #157, opened 2026-04-22
fix: isolate release metadata downloads

Astro-Han:
## Summary

Keep live updater metadata downloads in a separate temp directory from the predownloaded release metadata cache.

## Why

The v0.2.6 macOS x64 finalize run completed notarization and artifact verification, then failed while finalizing updater metadata because `gh release download` refused to write `latest-mac.yml` over a preexisting cached file. The release draft was repaired manually, but the workflow needs this fix so future finalize runs do not fail at the same step.

## Related Issue

No issue. Found during the v0.2.6 release workflow.

## How To Verify

```bash
bun --cwd packages/desktop-electron test scripts/release-metadata-contract.test.ts
```

## Screenshots or Recordings

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for merging live and predownloaded release metadata.
  * Enhanced download validation to prevent unintended file overwrites.

* **Chores**
  * Refined internal release metadata handling scripts.

### DIFF
--- a/packages/desktop-electron/scripts/finalize-latest-yml.ts
+++ b/packages/desktop-electron/scripts/finalize-latest-yml.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env bun
 
 import { $ } from "bun"
-import { mkdir } from "node:fs/promises"
+import { mkdtemp } from "node:fs/promises"
 import path from "path"
 
 const dir = process.env.LATEST_YML_DIR!
@@ -133,16 +133,15 @@ async function downloadExisting(tag: string, filename: string) {
     configured ? await readFile(path.join(configured, filename)) : undefined,
   )
 
-  const existingDir = path.join(tmp, "existing-latest-yml")
-  await mkdir(existingDir, { recursive: true })
+  const liveDir = await mkdtemp(path.join(tmp, "live-latest-yml-"))
   try {
-    await $`gh release download ${tag} --pattern ${filename} --dir ${existingDir} --repo ${repo}`.quiet()
+    await $`gh release download ${tag} --pattern ${filename} --dir ${liveDir} --repo ${repo} --clobber`.quiet()
   } catch (error) {
     const message = shellErrorText(error)
     if (isMissingAssetError(message)) return cached
     throw new Error(`Failed to download existing ${filename}: ${message}`)
   }
-  const live = assertSameVersion("GitHub release", filename, await readFile(path.join(existingDir, filename)))
+  const live = assertSameVersion("GitHub release", filename, await readFile(path.join(liveDir, filename)))
   return mergeLatest(cached, live)
 }
 
--- a/packages/desktop-electron/scripts/release-metadata-contract.test.ts
+++ b/packages/desktop-electron/scripts/release-metadata-contract.test.ts
@@ -114,6 +114,32 @@ describe("release metadata finalizer", () => {
     expect(latestMac).toContain("PawWork-stale-x64.zip")
   })
 
+  test("does not collide when EXISTING_LATEST_YML_DIR is RUNNER_TEMP/existing-latest-yml", async () => {
+    const root = mkdtempSync(join(tmpdir(), "pawwork-release-metadata-"))
+    roots.push(root)
+    const latestDir = join(root, "latest-yml")
+    const runnerTemp = join(root, "runner")
+    const binDir = join(root, "bin")
+    const snapshotDir = join(runnerTemp, "existing-latest-yml")
+    mkdirSync(runnerTemp, { recursive: true })
+    mkdirSync(binDir, { recursive: true })
+    mkdirSync(snapshotDir, { recursive: true })
+    writeFakeGh(binDir, {
+      "latest-mac.yml": "PawWork-live-x64.zip",
+    })
+    writeLatest(snapshotDir, "latest-mac.yml", "PawWork-snapshot-x64.zip")
+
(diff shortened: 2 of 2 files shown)


packages/desktop-electron/scripts/finalize-latest-yml.ts
```diff
@@ -133,7 +133,7 @@ async function downloadExisting(tag: string, filename: string) {
     configured ? await readFile(path.join(configured, filename)) : undefined,
   )
 
-  const existingDir = path.join(tmp, "existing-latest-yml")
+  const existingDir = path.join(tmp, "live-latest-yml")
```
Astro-Han: **P2 — brittle collision avoidance.**

The fix works only because `.github/workflows/build.yml` hardcodes `$RUNNER_TEMP/existing-latest-yml` and this file hardcodes `$RUNNER_TEMP/live-latest-yml`. That contract is invisible — a future edit to either path re-introduces the exact `already exists (use clobber ...)` failure that caused the v0.2.6 finalize to fall over.

Safer: make the live dir unique by construction so it can never equal `EXISTING_LATEST_YML_DIR`, e.g.

```ts
import { mkdtemp } from "node:fs/promises"
// ...
const existingDir = await mkdtemp(path.join(tmp, "live-latest-yml-"))
```

Then the script owns uniqueness instead of relying on the workflow author to remember.

Astro-Han: **P3 — stale variable name.**

The directory is now `live-latest-yml`, but the binding is still `existingDir`. Given that `cached` (the predownloaded snapshot from `EXISTING_LATEST_YML_DIR`) and the live-downloaded file are distinct concepts in this function, renaming to `liveDir` would make the `cached` vs live split self-documenting and avoid confusion with the `EXISTING_LATEST_YML_DIR` label used on line 131.

packages/desktop-electron/scripts/release-metadata-contract.test.ts
```diff
@@ -114,6 +114,32 @@ describe("release metadata finalizer", () => {
     expect(latestMac).toContain("PawWork-stale-x64.zip")
   })
 
+  test("downloads live metadata outside the predownloaded runner cache", async () => {
```
Astro-Han: **P3 — test name doesn't spell out the regression.**

`"downloads live metadata outside the predownloaded runner cache"` describes the implementation, not the failure mode. The scenario this pins down is: `EXISTING_LATEST_YML_DIR` is set to the same `$RUNNER_TEMP/existing-latest-yml` path the CI job pre-populates, and the finalize step must not collide when it calls `gh release download`. A name like `"does not collide when EXISTING_LATEST_YML_DIR is RUNNER_TEMP/existing-latest-yml"` would make the regression obvious from the test name alone.

Astro-Han (commented):
Nitpick review. Fix is effective, but worth hardening.

- **P2** — collision avoidance is implicit/brittle (see inline comment on `finalize-latest-yml.ts:136`).
- **P3** — variable name no longer matches the renamed path (`finalize-latest-yml.ts:136`).
- **P3** — new test name does not spell out the regression it guards (`release-metadata-contract.test.ts:117`).

Nothing blocking.

packages/desktop-electron/scripts/release-metadata-contract.test.ts
```diff
@@ -291,7 +317,12 @@ function writeFakeGh(binDir: string, downloads: Record<string, string | FixtureF
       "    console.error(`no matches found for ${pattern}`)",
       "    process.exit(1)",
       "  }",
-      "  writeFileSync(join(dir, pattern), latestYml(downloads[pattern]), 'utf8')",
+      "  const target = join(dir, pattern)",
+      "  if (existsSync(target)) {",
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The fake `gh release download` now respects `--clobber` and `-c` before failing on an existing target file.

packages/desktop-electron/scripts/finalize-latest-yml.ts
```diff
@@ -133,7 +133,7 @@ async function downloadExisting(tag: string, filename: string) {
...
   await mkdir(existingDir, { recursive: true })
   try {
     await $`gh release download ${tag} --pattern ${filename} --dir ${existingDir} --repo ${repo}`.quiet()
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The live download now uses a unique mkdtemp directory and passes `--clobber` to `gh release download` for idempotency.

packages/desktop-electron/scripts/finalize-latest-yml.ts
Astro-Han: Fixed. The script now creates the live metadata download directory with `mkdtemp`, so uniqueness is owned by the finalizer instead of relying on a workflow path convention.

Astro-Han: Fixed. Renamed the live download binding from `existingDir` to `liveDir` so it is distinct from the cached `EXISTING_LATEST_YML_DIR` snapshot.

packages/desktop-electron/scripts/release-metadata-contract.test.ts
Astro-Han: Fixed. Renamed the regression test to state the failure mode: it must not collide when `EXISTING_LATEST_YML_DIR` is `$RUNNER_TEMP/existing-latest-yml`.

PR #160, opened 2026-04-22
fix(desktop): avoid main util ts import

### LINKED ISSUE #159: [Bug] Packaged app crashes on startup when main process loads @opencode-ai/util TypeScript source
## What happened?
PawWork v0.2.6 packaged macOS app crashes on startup with a main-process JavaScript error before the normal UI opens:

```text
Uncaught Exception:
Error [ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING]: Stripping types is currently unsupported for files under node_modules, for "file:///Applications/PawWork.app/Contents/Resources/app.asar/node_modules/@opencode-ai/util/src/file-extensions.ts"
    at stripTypeScriptModuleTypes (node:internal/modules/typescript:183:11)
    at ModuleLoader.<anonymous> (node:internal/modules/esm/translators:650:29)
    at #translate (node:internal/modules/esm/loader:473:20)
    at afterLoad (node:internal/modules/esm/loader:529:29)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:534:12)
    at #getOrCreateModuleJobAfterResolve (node:internal/modules/esm/loader:577:36)
    at afterResolve (node:internal/modules/esm/loader:625:52)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:631:12)
    at ModuleJob.syncLink (node:internal/modules/esm/module_job:160:33)
    at ModuleJob.link (node:internal/modules/esm/module_job:245:17)
```

## Confirmed root cause
The v0.2.6 packaged `app.asar/out/main/index.js` contains a runtime main-process import of `@opencode-ai/util/file-extensions`. The packaged `node_modules/@opencode-ai/util/package.json` exports `./*` to `./src/*.ts`, so Electron resolves that import to `node_modules/@opencode-ai/util/src/file-extensions.ts`. Node's TypeScript stripping refuses files under `node_modules`, which makes the main process crash during startup.

## Steps to reproduce
1. Install or open PawWork v0.2.6 from the packaged macOS app bundle or DMG.
2. Launch PawWork.
3. The app shows the main-process JavaScript error above before the normal UI opens.

The packaged app should start normally. Electron main-process runtime code should not load `.ts` source files from `node_modules`.

## PawWork version
v0.2.6 packaged macOS app.

## OS version
macOS 26.3.1 on the local machine where the error was observed.

Yes. Opening the v0.2.6 packaged app reproduces the crash.

## Fix status
Fix PR: #160. The PR body includes `Closes #159`, so merging that PR should close this issue.

## Follow-up
The direct startup crash is tracked here. The CI and release guard gap that allowed this packaged-only failure to escape is tracked separately in #162.

Astro-Han:
## Summary

Avoid importing `@opencode-ai/util/file-extensions` from the Electron main process by moving the attachment MIME lookup into desktop main code. Bump release packages to `0.2.7` for the startup-crash hotfix.

## Why

The published `v0.2.6` packaged app crashes on startup because the main process loads `node_modules/@opencode-ai/util/src/file-extensions.ts`, and Node refuses TypeScript stripping for files under `node_modules`. The broken `v0.2.6` GitHub Release has been deleted to prevent accidental downloads; `v0.2.7` should replace it as the latest stable release.

## Related Issue

Closes #159.

## How To Verify

```bash
bun --cwd packages/desktop-electron test src/main/attachment-mime.test.ts
bun --cwd packages/desktop-electron build
rg -n "@opencode-ai/util/file-extensions|node_modules/@opencode-ai/util/src/file-extensions" packages/desktop-electron/out || true
bun install --frozen-lockfile
```

## Screenshots or Recordings

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Bumped package versions to 0.2.7 across the monorepo.

* **New Features**
  * Added attachment file type → MIME detection to improve file handling and validation.

* **Tests**
  * Added tests covering attachment MIME detection, extension handling, and unsupported/edge cases.

### DIFF
--- a/packages/desktop-electron/src/main/attachment-mime.test.ts
+++ b/packages/desktop-electron/src/main/attachment-mime.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, test } from "bun:test"
+
+import { IMAGE_EXTS } from "@opencode-ai/util/file-extensions"
+
+import { attachmentPathMime, MIME_BY_EXTENSION } from "./attachment-mime"
+
+describe("attachmentPathMime", () => {
+  test("returns MIME types for direct attachment files", () => {
+    expect(attachmentPathMime("/tmp/image.gif")).toBe("image/gif")
+    expect(attachmentPathMime("/tmp/image.jpeg")).toBe("image/jpeg")
+    expect(attachmentPathMime("/tmp/image.png")).toBe("image/png")
+    expect(attachmentPathMime("/tmp/image.JPG")).toBe("image/jpeg")
+    expect(attachmentPathMime("/tmp/image.webp")).toBe("image/webp")
+    expect(attachmentPathMime("/tmp/document.pdf")).toBe("application/pdf")
+  })
+
+  test("returns MIME types using an injected extension reader", () => {
+    const extname = (filepath: string) => filepath.slice(filepath.lastIndexOf("."))
+
+    expect(attachmentPathMime("C:\\tmp\\image.PNG", extname)).toBe("image/png")
+  })
+
+  test("keeps image MIME entries in sync with the renderer contract", () => {
+    const imageEntries = [...MIME_BY_EXTENSION.entries()]
+      .filter(([extension]) => extension !== "pdf")
+      .sort(([left], [right]) => left.localeCompare(right))
+
+    const rendererImageEntries = [...IMAGE_EXTS.entries()].sort(([left], [right]) => left.localeCompare(right))
+
+    expect(imageEntries).toEqual(rendererImageEntries)
+  })
+
+  test("returns undefined for unsupported extensions", () => {
+    expect(attachmentPathMime("/tmp/archive.zip")).toBeUndefined()
+  })
+
+  test("returns undefined for missing extensions", () => {
+    expect(attachmentPathMime("/tmp/no-extension")).toBeUndefined()
+  })
+})
--- a/packages/desktop-electron/src/main/attachment-mime.ts
+++ b/packages/desktop-electron/src/main/attachment-mime.ts
@@ -1,0 +1,16 @@
+import path from "node:path"
+
+// Keep image entries in sync with packages/util/src/file-extensions.ts::IMAGE_EXTS.
+export const MIME_BY_EXTENSION = new Map([
+  ["gif", "image/gif"],
+  ["jpeg", "image/jpeg"],
+  ["jpg", "image/jpeg"],
+  ["pdf", "application/pdf"],
+  ["png", "image/png"],
+  ["webp", "image/webp"],
+])
+
+export function attachmentPathMime(filepath: string, extname = path.extname) {
+  const suffix = extname(filepath).slice(1).toLowerCase()
+  return MIME_BY_EXTENSION.get(suffix)
+}
--- a/bun.lock
+++ b/bun.lock
@@ -26,7 +26,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "0.2.6",
+      "version": "0.2.7",
       "dependencies": {
         "@kobalte/core": "catalog:",
         "@opencode-ai/sdk": "workspace:*",
@@ -80,7 +80,7 @@
     },
     "packages/desktop-electron": {
       "name": "@opencode-ai/desktop-electron",
-      "version": "0.2.6",
+      "version": "0.2.7",
       "dependencies": {
         "@opencode-ai/util": "workspace:*",
         "effect": "catalog:",
@@ -141,7 +141,7 @@
     },
     "packages/opencode": {
       "name": "opencode",
-      "version": "0.2.6",
+      "version": "0.2.7",
       "bin": {
         "opencode": "./bin/opencode",
       },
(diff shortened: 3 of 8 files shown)


packages/desktop-electron/src/main/attachment-mime.ts
```diff
@@ -0,0 +1,15 @@
+import path from "node:path"
+
+const IMAGE_MIME_BY_EXTENSION = new Map([
+  ["gif", "image/gif"],
+  ["jpeg", "image/jpeg"],
+  ["jpg", "image/jpeg"],
+  ["png", "image/png"],
+  ["webp", "image/webp"],
```
Astro-Han, line 10: **P2 — drift risk with the renderer's canonical list.** `IMAGE_MIME_BY_EXTENSION` is a verbatim copy of `IMAGE_EXTS` in `packages/util/src/file-extensions.ts`. The renderer (`packages/app/src/components/prompt-input/files.ts`, `attachment-routing.ts`) still imports the util map to decide what MIME to send to main, and `ipc.ts:204` rejects the request when `attachmentPathMime(normalized) !== mime`. If the util map grows a new image format and this copy isn't updated in lockstep, renderer-approved attachments will silently return `null` from `read-file-data-url` with no user-visible error.

At minimum, add a short comment here pointing at `packages/util/src/file-extensions.ts::IMAGE_EXTS` so a future editor sees the contract. A stronger fix: a contract test that imports both and asserts equality, so the mismatch fails CI rather than production.

packages/desktop-electron/src/main/attachment-mime.test.ts
```diff
@@ -0,0 +1,16 @@
+import { describe, expect, test } from "bun:test"
+
+import { attachmentPathMime } from "./attachment-mime"
+
+describe("attachmentPathMime", () => {
+  test("returns MIME types for direct attachment files", () => {
+    expect(attachmentPathMime("/tmp/image.png")).toBe("image/png")
+    expect(attachmentPathMime("/tmp/image.JPG")).toBe("image/jpeg")
+    expect(attachmentPathMime("/tmp/document.pdf")).toBe("application/pdf")
```
Astro-Han, line 14: **P3 — coverage gap.** The first test exercises `png`, `JPG` (case-insensitivity), and `pdf`, but three of the five entries in `IMAGE_MIME_BY_EXTENSION` (`gif`, `webp`, `jpeg` — distinct from `jpg`) are never asserted. A future refactor that drops one of them would slip through. Adding one-line assertions for each keeps the map honest.

packages/desktop-electron/src/main/attachment-mime.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  })
+
+  test("ignores unsupported extensions", () => {
+    expect(attachmentPathMime("/tmp/archive.zip")).toBeUndefined()
+    expect(attachmentPathMime("/tmp/no-extension")).toBeUndefined()
+  })
```
Astro-Han, line 39: **P3 — test name conflates two branches.** `ignores unsupported extensions` covers both "extension present but unsupported" (`archive.zip`) and "no extension at all" (`no-extension`), which hit different code paths (`Map.get("zip")` vs `path.extname("...") === ""`). Splitting into two tests, or renaming to something like `returns undefined for unsupported or missing extensions`, makes the intent match what's actually asserted.

Astro-Han (commented):
Three nits on the hotfix. Approach is fine — moving the narrow slice into main's source tree sidesteps the Node-refuses-TS-in-node_modules failure. Flagging drift risk and test gaps below.

One alternative worth a thought for a follow-up (not this PR): `externalizeDeps: { exclude: ["@opencode-ai/util"] }` in `electron.vite.config.ts` would bundle the util module into main at build time, keeping a single source of truth with the renderer. Current approach is defensible for a hotfix.

packages/desktop-electron/src/main/attachment-mime.ts
```diff
@@ -0,0 +1,15 @@
...
+])
+
+export function attachmentPathMime(filepath: string) {
+  const suffix = path.extname(filepath).slice(1).toLowerCase()
+  if (suffix === "pdf") return "application/pdf"
+  return IMAGE_MIME_BY_EXTENSION.get(suffix)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Unified PDF into MIME_BY_EXTENSION and made the extension reader injectable while keeping the default path.extname behavior.

packages/desktop-electron/src/main/attachment-mime.test.ts
Astro-Han: Fixed. Split unsupported-extension and missing-extension behavior into separate tests.

packages/desktop-electron/src/main/attachment-mime.test.ts
Astro-Han: Fixed. The direct attachment test now asserts gif, jpeg, jpg case-insensitivity, png, webp, and pdf.

packages/desktop-electron/src/main/attachment-mime.ts
Astro-Han: Fixed. Added an explicit comment for the util IMAGE_EXTS contract and a contract test that compares the main-process image MIME entries against @opencode-ai/util/file-extensions.

PR #161, opened 2026-04-22
fix: preserve persisted cache crash diagnostics

### LINKED ISSUE #156: [Bug] Packaged app can crash with masked persisted workspace cache error
## What happened?
PawWork v0.2.5 packaged app showed a renderer error page while opening or rendering a session/workspace view:

```text
Error: Failed to create persisted cache
    at ensureChild (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/main-CpucilYi.js:68422:23)
    at Object.child (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/main-CpucilYi.js:68504:24)
    at file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/main-CpucilYi.js:69475:23
    at get data (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/main-CpucilYi.js:69738:16)
    at Object.fn (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/session-DGl7U4CN.js:25801:27)
    at runComputation (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/styles-C-a255bR.js:468:22)
    at updateComputation (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/styles-C-a255bR.js:451:3)
    at runTop (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/styles-C-a255bR.js:546:7)
    at runQueue (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/styles-C-a255bR.js:618:42)
    at completeUpdates (file:///Applications/PawWork.app/Contents/Resources/app.asar/out/renderer/assets/styles-C-a255bR.js:574:5)
```

Local source inspection maps the top frame to `packages/app/src/context/global-sync/child-store.ts`, where `ensureChild()` creates the workspace-level `vcs` persisted store:

```ts
const vcs = runWithOwner(input.owner, () =>
  persisted(
    Persist.workspace(directory, "vcs", ["vcs.v1"]),
    createStore({ value: undefined as VcsInfo | undefined }),
  ),
)
if (!vcs) throw new Error(input.translate("error.childStore.persistedCacheCreateFailed"))
```

`persisted()` normally returns a tuple, so the visible error is masking an earlier synchronous exception from inside the `runWithOwner(...persisted...)` callback. Solid's `runWithOwner()` can route callback errors through the owner error handling path and return `undefined`; PawWork then replaces the original cause with the generic translated error. The generic `Failed to create persisted cache` message is therefore the symptom, not the root cause.

This is separate from #145's startup hang: the latest local `~/Library/Logs/PawWork/main.log` shows the app reached `loading task finished`, `init step { phase: 'done' }`, and `server ready`. The main-process log did not include the renderer storage cause. The recurring updater `app-update.yml` warning in the same log appears unrelated to this renderer crash.

## Updated diagnosis
After extracting `/Applications/PawWork.app/Contents/Resources/app.asar` from the affected v0.2.5 install, the packaged renderer still uses the old OpenCode storage names:

```js
const GLOBAL_STORAGE = "opencode.global.dat"
return `opencode.workspace.${head}.${sum}.dat`
```

The storage namespace rename to `pawwork.global.dat` / `pawwork.workspace.*.dat` landed after v0.2.5 and should be included in v0.2.6. Therefore, do not attribute this specific observed v0.2.5 crash to the later `pawwork.*` storage rename. The relevant v0.2.5 question is what original synchronous exception was hidden behind `Failed to create persisted cache`.

The stronger current hypothesis is a bad workspace directory value reaching `ensureChild()`. In the packaged bundle, the path is:

```js
get data() {
  return current()[0]
}
```

where `current()` calls `globalSync.child(sdk.directory)`, and `ensureChild(directory)` only logs when `directory` is falsy:

```ts
if (!directory) console.error("No directory provided")
```

It then continues into `Persist.workspace(directory, "vcs", ...)`. `workspaceStorage(dir)` immediately calls `dir.slice(0, 12)`, so if `directory` is `undefined`, `null`, or otherwise not a string, it will synchronously throw before `persisted()` can return. That synchronous throw is then masked as `Failed to create persisted cache` by the `runWithOwner()` fallback path.

This does not prove the bad-directory hypothesis yet, because the original exception is currently hidden. It does make a broad app-data permission explanation less likely for this v0.2.5 report.

The session/workspace view should keep rendering, recover, or at minimum show/report the original failure with enough context to diagnose it. A generic `Failed to create persisted cache` error page is not actionable.

## What actually happened?
The packaged app entered a renderer error page before the session view could render. The visible error only identifies the first child-store cache (`vcs`) and hides the real synchronous failure.

## Investigation notes
- Installed app: `/Applications/PawWork.app`, version `0.2.5`.
- App data directory: `~/Library/Application Support/ai.pawwork.desktop`.
- The affected packaged renderer still uses `opencode.global.dat` and `opencode.workspace.*.dat`.
- The current `dev` / v0.2.6 line uses `pawwork.global.dat` and `pawwork.workspace.*.dat`, so storage namespace rename should already be covered for the next release line.
- Manual local probing showed the app data directory itself is writable; this does not prove the packaged renderer storage path is healthy, but it makes a broad directory-permission failure less likely.
- The most useful next diagnostic is preserving the original `cause`, `directory`, `storage`, and `key` when child-store persisted cache creation fails.
- No existing issue was found for the exact `Failed to create persisted cache` symptom.

## Suggested fix direction
- Add an explicit guard in `ensureChild(directory)` for non-string or empty directory values. Do not continue into `Persist.workspace(...)` after only logging `No directory provided`.
- Preserve and surface the original cause when `runWithOwner(...persisted...)` fails, including the directory, persistence target storage name, key, and original error/cause when available.
- Add a regression test for a bad directory value reaching `ensureChild()` so it fails with an actionable error instead of `Failed to create persisted cache`.
- Add a regression test that simulates workspace persisted cache initialization failure and verifies the diagnostic is not replaced by only `Failed to create persisted cache`.
- After the true cause is visible, decide whether the production fix belongs in the route/SDK directory lifecycle, child-store input validation, or desktop storage handling.

## PawWork version
v0.2.5 packaged macOS app.

## OS version
macOS on the local development machine.

Not yet. Observed from the packaged app error page and investigated from the stack trace plus local source and extracted packaged renderer bundle.

Astro-Han:
## Summary

Preserve the original cause and child-store cache context when workspace persisted cache creation fails. Invalid runtime workspace directory input now fails before persistence setup instead of continuing into `Persist.workspace(...)`.

## Why

Closes #156. The packaged app could show only `Failed to create persisted cache`, which hid the synchronous exception, storage file, key, cache kind, and workspace directory. The fix keeps the failure at the child-store layer and does not change storage namespace behavior.

This intentionally turns empty workspace directories into explicit child-store errors too. Previously `ensureChild("")` only logged and continued into a fallback workspace cache path, which masked bad upstream input.

## Related Issue

Closes #156

## How To Verify

```bash
(cd packages/app && bun test --preload ./happydom.ts ./src/context/sync-current-child.test.ts ./src/context/global-sync/child-store-error.test.ts ./src/context/global-sync/child-store.test.ts ./src/utils/persist.test.ts)
```

## Screenshots or Recordings

Not applicable. No visible UI layout change.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Better child-store persistence flow with optional custom persistence support and stricter directory validation.

* **Bug Fixes**
  * Improved error wrapping that preserves original causes and prevents partial/invalid cache publishing.

* **Tests**
  * Added tests covering error diagnostics, propagation, and persisted-cache failure scenarios.

### DIFF
--- a/packages/app/src/context/global-sync/child-store-error.test.ts
+++ b/packages/app/src/context/global-sync/child-store-error.test.ts
@@ -1,0 +1,144 @@
+import { describe, expect, test } from "bun:test"
+import { catchError, createRoot, getOwner } from "solid-js"
+import { createStore } from "solid-js/store"
+import {
+  ChildStoreError,
+  createPersistedChildCache,
+  validateChildStoreDirectory,
+} from "./child-store-error"
+
+const translate = (key: string) => {
+  if (key === "error.childStore.persistedCacheCreateFailed") return "Failed to create persisted cache"
+  return key
+}
+
+describe("child store diagnostics", () => {
+  test("rejects invalid workspace directories before persisted cache setup", () => {
+    for (const directory of [undefined, ""]) {
+      expect(() => validateChildStoreDirectory(directory)).toThrow("Invalid workspace directory for child store")
+
+      try {
+        validateChildStoreDirectory(directory)
+      } catch (error) {
+        expect(error).toBeInstanceOf(ChildStoreError)
+        expect((error as ChildStoreError).context).toEqual({
+          kind: "workspace",
+          directory,
+        })
+        continue
+      }
+
+      throw new Error("expected invalid directory to throw")
+    }
+  })
+
+  test("describes unserializable invalid directories without masking validation errors", () => {
+    const circular: { self?: unknown } = {}
+    circular.self = circular
+
+    for (const directory of [1n, circular]) {
+      try {
+        validateChildStoreDirectory(directory)
+      } catch (error) {
+        expect(error).toBeInstanceOf(ChildStoreError)
+        expect((error as ChildStoreError).message).toContain("<unserializable>")
+        expect((error as ChildStoreError).context.directory).toBe(directory)
+        continue
+      }
+
+      throw new Error("expected invalid directory to throw")
+    }
+  })
+
+  test("preserves the original persisted cache setup cause and context", () => {
+    const cause = new TypeError("storage init failed")
+    const target = {
+      storage: "pawwork.workspace.-tmp-project.abc123.dat",
+      key: "workspace:vcs",
+      legacy: ["vcs.v1"],
+    }
+
+    createRoot((dispose) => {
+      const owner = getOwner()
+      if (!owner) throw new Error("owner required")
+
+      try {
+        createPersistedChildCache({
+          owner,
+          directory: "/tmp/project",
+          kind: "vcs",
+          messageKey: "error.childStore.persistedCacheCreateFailed",
+          target,
+          store: createStore({ value: undefined as string | undefined }),
+          translate,
+          persist: () => {
+            throw cause
+          },
+        })
+      } catch (error) {
+        dispose()
+        expect(error).toBeInstanceOf(ChildStoreError)
+        expect(error).toHaveProperty("cause", cause)
+        expect((error as ChildStoreError).message).toContain("Failed to create persisted cache")
+        expect((error as ChildStoreError).message).toContain("cache=vcs")
+        expect((error as ChildStoreError).message).toContain("storage=pawwork.workspace.-tmp-project.abc123.dat")
+        expect((error as ChildStoreError).message).toContain("key=workspace:vcs")
+        expect((error as ChildStoreError).context).toEqual({
+          kind: "vcs",
+          directory: "/tmp/project",
+          storage: "pawwork.workspace.-tmp-project.abc123.dat",
+          key: "workspace:vcs",
+        })
+        return
+      }
+
+      dispose()
+      throw new Error("expected persisted cache setup to throw")
+    })
+  })
+
+  test("preserves cause when owner error boundary handles persisted setup failure", () => {
+    const cause = new TypeError("storage init failed")
+    const handled: unknown[] = []
+
+    catchError(
+      () => {
+        createRoot((dispose) => {
+          const owner = getOwner()
+          if (!owner) throw new Error("owner required")
+
+          try {
+            createPersistedChildCache({
+              owner,
+              directory: "/tmp/project",
+              kind: "vcs",
+              messageKey: "error.childStore.persistedCacheCreateFailed",
+              target: {
+                storage: "pawwork.workspace.-tmp-project.abc123.dat",
+                key: "workspace:vcs",
+              },
+              store: createStore({ value: undefined as string | undefined }),
+              translate,
+              persist: () => {
+                throw cause
+              },
+            })
+          } catch (error) {
+            dispose()
+            expect(error).toBeInstanceOf(ChildStoreError)
+            expect(error).toHaveProperty("cause", cause)
+            return
+          }
+
+          dispose()
+          throw new Error("expected persisted cache setup to throw")
+        })
+      },
+      (error) => {
+        handled.push(error)
+      },
+    )
+
+    expect(handled).toEqual([cause])
+  })
+})
--- a/packages/app/src/context/global-sync/child-store-error.ts
+++ b/packages/app/src/context/global-sync/child-store-error.ts
@@ -1,0 +1,121 @@
+import { runWithOwner, type Owner } from "solid-js"
+import type { SetStoreFunction, Store } from "solid-js/store"
+import { persisted } from "@/utils/persist"
+
+export type ChildStoreCacheKind = "workspace" | "vcs" | "project" | "icon"
+
+export type ChildStoreErrorContext = {
+  kind: ChildStoreCacheKind
+  directory: unknown
+  storage?: string
+  key?: string
+}
+
+export type ChildStorePersistTarget = {
+  storage?: string
+  key: string
+  legacy?: string[]
+  migrate?: (value: unknown) => unknown
+}
+
+type PersistedResult<T> = ReturnType<typeof persisted<T>>
+
+export type ChildStorePersistedFactory = <T>(
+  target: ChildStorePersistTarget,
+  store: [Store<T>, SetStoreFunction<T>],
+) => PersistedResult<T>
+
+export class ChildStoreError extends Error {
+  readonly context: ChildStoreErrorContext
+
+  constructor(message: string, context: ChildStoreErrorContext, options?: { cause?: unknown }) {
+    super(message, options)
+    this.name = "ChildStoreError"
+    this.context = context
+  }
+}
+
+function describeDirectory(directory: unknown) {
+  if (typeof directory === "string") return directory.length > 0 ? JSON.stringify(directory) : "empty string"
+  if (directory === undefined) return "undefined"
+  if (directory === null) return "null"
+  try {
+    const json = JSON.stringify(directory)
+    return `${typeof directory}: ${json ?? String(directory)}`
+  } catch {
+    return `${typeof directory}: <unserializable>`
+  }
+}
+
+function contextDetails(context: ChildStoreErrorContext) {
+  return [
+    `cache=${context.kind}`,
+    context.storage ? `storage=${context.storage}` : undefined,
+    context.key ? `key=${context.key}` : undefined,
+    `directory=${describeDirectory(context.directory)}`,
+  ]
+    .filter((part): part is string => !!part)
+    .join(", ")
+}
+
+export function validateChildStoreDirectory(directory: unknown): asserts directory is string {
+  if (typeof directory === "string" && directory.length > 0) return
+  throw new ChildStoreError(`Invalid workspace directory for child store: ${describeDirectory(directory)}`, {
+    kind: "workspace",
+    directory,
+  })
+}
+
+export function createChildStorePersistedCacheError(input: {
+  translate: (key: string) => string
+  messageKey: string
+  context: ChildStoreErrorContext
+  cause?: unknown
+}) {
+  return new ChildStoreError(`${input.translate(input.messageKey)} (${contextDetails(input.context)})`, input.context, {
+    cause: input.cause,
+  })
+}
+
+export function createPersistedChildCache<T>(input: {
+  owner: Owner
+  directory: string
+  kind: Exclude<ChildStoreCacheKind, "workspace">
+  messageKey: string
+  target: ChildStorePersistTarget
+  store: [Store<T>, SetStoreFunction<T>]
+  translate: (key: string) => string
+  persist?: ChildStorePersistedFactory
+}) {
+  const persist: ChildStorePersistedFactory = input.persist ?? persisted
+  const context: ChildStoreErrorContext = {
+    kind: input.kind,
+    directory: input.directory,
+    storage: input.target.storage,
+    key: input.target.key,
+  }
+  let cause: unknown
+
+  try {
+    const cache = runWithOwner(input.owner, () => {
+      try {
+        return persist(input.target, input.store)
+      } catch (error) {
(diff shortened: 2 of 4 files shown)


packages/app/src/context/global-sync/child-store-error.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.name = "ChildStoreError"
+    this.context = context
+  }
+}
+
+function describeDirectory(directory: unknown) {
+  if (typeof directory === "string") return directory.length > 0 ? JSON.stringify(directory) : "empty string"
+  if (directory === undefined) return "undefined"
+  if (directory === null) return "null"
+  const json = JSON.stringify(directory)
```
Astro-Han: **P2** — `JSON.stringify(directory)` throws on `BigInt` and on circular structures. If one of those reaches `ensureChild` (unlikely, but possible via typed-as-`unknown` ingress), the diagnostic reporter itself will throw and replace the real cause with a `TypeError: Do not know how to serialize a BigInt` — exactly the masking behavior this PR was written to prevent. Wrap in a `try { JSON.stringify(...) } catch { return \`${typeof directory}: <unserializable>\` }`.

packages/app/src/context/global-sync/child-store-error.ts
```diff
@@ -0,0 +1,116 @@
...
+  let cause: unknown
+
+  try {
+    const cache = runWithOwner(input.owner, () => {
+      try {
+        return persist(input.target, input.store)
+      } catch (error) {
+        cause = error
+        throw error
+      }
```
Astro-Han, line 107: **P3** — This inner `try { cause = error; throw error }` is load-bearing. Solid's `runWithOwner` catches via `handleError` and returns `undefined` when an error boundary handles it, so without capturing `cause` here first, the outer `catch` never fires and the real cause is lost. Worth a one-line comment (e.g. `// runWithOwner swallows into handleError; snapshot cause first`) so a future reader doesn't "simplify" this away.

packages/app/src/context/global-sync/child-store-error.ts
```diff
@@ -0,0 +1,116 @@
...
+    context.storage ? `storage=${context.storage}` : undefined,
+    context.key ? `key=${context.key}` : undefined,
+    `directory=${describeDirectory(context.directory)}`,
+  ]
+    .filter((part): part is string => !!part)
+    .join(", ")
+}
+
+export function validateChildStoreDirectory(directory: string): asserts directory is string {
+  if (typeof directory === "string" && directory.length > 0) return
```
Astro-Han, line 62: **P3** — The declared parameter type `directory: string` is a white lie: callers routinely pass values typed as `string` but originating from SDK/DOM boundaries (e.g. `sdk.directory`, `target.worktree`) that can actually be `undefined`. Since the whole point is to catch that case, prefer `directory: unknown`; the `asserts directory is string` signature already narrows correctly inside the caller.

packages/app/src/context/global-sync/child-store.ts
```diff
@@ -122,34 +128,43 @@ export function createChildStoreManager(input: {
...
-      if (!vcs) throw new Error(input.translate("error.childStore.persistedCacheCreateFailed"))
+      const vcs = createPersistedChildCache({
+        owner: input.owner,
+        directory,
+        kind: "vcs",
+        messageKey: "error.childStore.persistedCacheCreateFailed",
+        target: Persist.workspace(directory, "vcs", ["vcs.v1"]),
+        store: createStore({ value: undefined as VcsInfo | undefined }),
+        translate: input.translate,
+        ...(input.persist ? { persist: input.persist } : {}),
```
Astro-Han: **P3** — `createPersistedChildCache` already defaults `persist` via `input.persist ?? persisted` (child-store-error.ts:86), so the conditional spread collapses to `persist: input.persist`. Same at L154 and L166 — three copies of the same indirection.

packages/app/src/context/global-sync/child-store.ts
```diff
@@ -122,34 +128,43 @@ export function createChildStoreManager(input: {
   }
 
   function ensureChild(directory: string) {
-    if (!directory) console.error("No directory provided")
+    validateChildStoreDirectory(directory)
```
Astro-Han, line 131: **P2** — Behavior change worth flagging in the PR description: the previous `ensureChild` only `console.error`'d on a falsy directory and kept going, which meant `directory = ""` silently created a throwaway cache under `pawwork.workspace.workspace.<sum>.dat`. The validator now throws on `""` as well as `undefined`/`null`. Given ~20 call sites in layout.tsx/sidebar-*.tsx pass `session.directory`, `target.worktree`, etc., any code path that was previously surviving on an empty string will now surface as a visible crash. That matches the issue #156 intent, but it's not a pure diagnostic-only change — worth a one-liner in the PR body so reviewers know to watch post-merge telemetry.

packages/app/src/context/global-sync/child-store-error.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const translate = (key: string) => {
+  if (key === "error.childStore.persistedCacheCreateFailed") return "Failed to create persisted cache"
+  return key
+}
+
+describe("child store diagnostics", () => {
+  test("rejects invalid workspace directories before persisted cache setup", () => {
+    expect(() => validateChildStoreDirectory(undefined as unknown as string)).toThrow(
+      "Invalid workspace directory for child store",
```
Astro-Han: **P2** — Coverage gap: the validator rejects both `undefined` (tested) and empty string (`directory.length > 0`, not tested). An empty-string regression would slip through. Add one more case: `expect(() => validateChildStoreDirectory("")).toThrow(/Invalid workspace directory/)` with the `context.directory` assertion being `""`.

packages/app/src/context/global-sync/child-store-error.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        })
+      } catch (error) {
+        dispose()
+        expect(error).toBeInstanceOf(ChildStoreError)
+        expect(error).toHaveProperty("cause", cause)
+        expect((error as ChildStoreError).message).toContain("Failed to create persisted cache")
+        expect((error as ChildStoreError).message).toContain("cache=vcs")
+        expect((error as ChildStoreError).message).toContain("storage=pawwork.workspace.-tmp-project.abc123.dat")
+        expect((error as ChildStoreError).message).toContain("key=workspace:vcs")
+        expect((error as ChildStoreError).context).toEqual({
```
Astro-Han, line 86: **P2** — The test injects `persist: () => { throw cause }`, but because the owner from `getOwner()` here has no error boundary, Solid's `runWithOwner` re-throws via `handleError` — meaning the outer `catch (error) { cause ??= error }` is what actually populates `cause` in this test. The inner `try { cause = error; throw error }` block (lines 99–102 of child-store-error.ts) is untested. Consider adding a case that wraps the owner with `catchError(...)` so `runWithOwner` returns `undefined` without re-throwing; without it, the inner block could be deleted and this suite would still pass.

packages/app/src/context/global-sync/child-store-error.ts
```diff
@@ -0,0 +1,116 @@
+import { runWithOwner, type Owner } from "solid-js"
```
Astro-Han, line 1: **P3** — File is named `child-store-error.ts` but exports are dominated by a factory (`createPersistedChildCache`) and a validator (`validateChildStoreDirectory`); the actual error class is incidental. `persisted-child-cache.ts` or `child-store-diagnostics.ts` would read closer to what the module does. Not worth a rename in this PR, but keep in mind if the file grows.

Astro-Han (commented):
Nitpick pass. Logic is sound and CI is green; comments below are P2/P3 polish.

packages/app/src/context/global-sync/child-store-error.ts
Astro-Han: Fixed. `describeDirectory` now catches unserializable values and reports `<unserializable>`; added BigInt and circular-object regression coverage.

packages/app/src/context/global-sync/child-store-error.ts
Astro-Han: Fixed. Added a one-line comment explaining that `runWithOwner` can route the error through Solid boundary handling and return `undefined`, so the cause snapshot is intentional.

packages/app/src/context/global-sync/child-store-error.ts
Astro-Han: Fixed. `validateChildStoreDirectory` now accepts `unknown` and narrows with `asserts directory is string`.

packages/app/src/context/global-sync/child-store.ts
Astro-Han: Fixed. Collapsed the three conditional spreads to `persist: input.persist`; `bun --cwd packages/app typecheck` passes.

packages/app/src/context/global-sync/child-store.ts
Astro-Han: Fixed in the PR body. It now calls out that empty workspace directories intentionally surface as explicit child-store errors instead of continuing into the fallback workspace cache path.

packages/app/src/context/global-sync/child-store-error.test.ts
Astro-Han: Fixed. The validator test now covers both `undefined` and `""`, including the `context.directory` value.

packages/app/src/context/global-sync/child-store-error.test.ts
Astro-Han: Fixed. Added a `catchError(...)` regression that exercises the path where `runWithOwner` returns `undefined`; the `ChildStoreError` still keeps the original cause.

packages/app/src/context/global-sync/child-store-error.ts
Astro-Han: No code change. Keeping the file name in this PR to avoid rename churn; the module is still local to child-store diagnostics, and a rename is only worth revisiting if it grows further.

packages/app/src/context/global-sync/child-store.ts
```diff
@@ -122,34 +128,43 @@ export function createChildStoreManager(input: {
...
+        owner: input.owner,
+        directory,
+        kind: "icon",
+        messageKey: "error.childStore.persistedProjectIconCreateFailed",
+        target: Persist.workspace(directory, "icon", ["icon.v1"]),
+        store: createStore({ value: undefined as string | undefined }),
+        translate: input.translate,
+        persist: input.persist,
+      })
       iconCache.set(directory, { store: icon[0], setStore: icon[1], ready: icon[3] })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Classified as P2 for this PR because a later persisted-cache failure could leave a visible cache map entry for a child store that never finished initializing. Cache map publication is now deferred until `vcs`, `project`, and `icon` persisted caches all succeed, and `child-store.test.ts` covers the project-cache failure path with no `children`, `vcsCache`, `metaCache`, or `iconCache` entry left behind.

PR #184, opened 2026-04-23
fix: localize macOS menu role labels

### LINKED ISSUE #177: [Bug] Desktop menu role items remain English in Chinese locale
### What happened?

In the new version, the macOS desktop menu is still partially English when PawWork is running in Chinese. This is a narrower follow-up split from #84. The Chinese locale is active and PawWork-controlled custom menu labels are already localized, but many menu items that PawWork includes through Electron `role` entries still render in English.

Observed English residual items include `About PawWork`, `Hide PawWork`, `Hide Others`, `Show All`, `Quit PawWork`, `Close Window`, `Undo`, `Redo`, `Cut`, `Copy`, `Paste`, `Select All`, `Reload`, `Toggle Developer Tools`, `Actual Size`, `Zoom In`, `Zoom Out`, `Toggle Full Screen`, `Minimize`, `Zoom`, and `Bring All to Front`.

### Steps to reproduce

1. Run PawWork on macOS with Chinese system language or PawWork language set to Chinese.
2. Open the PawWork application menu, File menu, Edit menu, View menu, and Window menu.
3. Compare PawWork custom items with Electron role items.

In Chinese locale, visible desktop menu items that PawWork includes in its menu template should be Chinese unless the item is genuinely inserted by macOS outside PawWork's template.

### Technical notes

Current root-cause evidence: `packages/desktop-electron/src/main/menu-labels.ts` defines labels only for PawWork custom keys. In `packages/desktop-electron/src/main/menu-template.ts`, localized items use explicit `label: t(...)`, but the English residual items are mostly role-only entries such as `role: "about"`, `role: "hide"`, `role: "hideOthers"`, `role: "unhide"`, `role: "quit"`, `role: "close"`, `role: "undo"`, `role: "redo"`, `role: "cut"`, `role: "copy"`, `role: "paste"`, `role: "selectAll"`, `role: "reload"`, `role: "toggleDevTools"`, `role: "resetZoom"`, `role: "zoomIn"`, `role: "zoomOut"`, `role: "togglefullscreen"`, and `role: "windowMenu"`.

The earlier #84 assumption that Electron or macOS role behavior could be left alone is not holding for the shipped menu, because those role entries still render in English while other system-provided items such as dictation and emoji are Chinese.

### Acceptance criteria

- Chinese locale shows Chinese labels for all PawWork-included visible desktop menu items where PawWork controls the template.
- Role-backed items preserve expected native behavior and shortcuts after localization.
- Items genuinely inserted by macOS outside PawWork's template may remain system-controlled.
- Tests cover the Chinese labels for role-backed menu items, not only PawWork custom menu labels.
- Linkage to #84 remains clear, but this issue can be completed independently from updater and feedback/report-flow work.

### Verification notes

Local verification before filing: `bun test src/main/menu.test.ts src/main/menu-labels.test.ts` from `packages/desktop-electron` passes. The gap is that current tests only prove custom PawWork labels are localized; they do not cover role-based menu item labels.

Related: #84.

Astro-Han:
## Summary

Localizes macOS desktop menu role-backed items in Chinese while keeping Electron roles attached for native behavior.

## Why

Closes #177. PawWork custom menu labels were already localized, but role-only Electron menu items still rendered in English in Chinese locale.

## Related Issue

Closes #177

## How To Verify

```bash
cd packages/desktop-electron && bun test src/main/menu.test.ts src/main/menu-labels.test.ts
bun turbo typecheck --filter=@opencode-ai/desktop-electron
bun turbo test:ci --filter=@opencode-ai/desktop-electron
```

## Screenshots or Recordings

Not captured. This change is covered by menu template tests that assert Chinese labels and preserved Electron roles; a manual macOS menu check is still useful before release.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Enhanced menu localization with role-specific labels, adding a localized "Window" menu and support for app-name placeholders in menu item text.

* **Bug Fixes**
  * Role-backed menu items now consistently show localized labels while keeping their original behavior.

* **Tests**
  * Added tests validating localized role-based menu labels (including Chinese) and the window submenu.

### DIFF
--- a/packages/desktop-electron/src/main/menu-labels.ts
+++ b/packages/desktop-electron/src/main/menu-labels.ts
@@ -4,6 +4,7 @@ export type MenuLabelKey =
   | "file"
   | "edit"
   | "view"
+  | "window"
   | "go"
   | "help"
   | "checkForUpdates"
@@ -25,11 +26,35 @@ export type MenuLabelKey =
   | "reportProblem"
   | "openGithubIssue"
 
+export type MenuRoleLabelKey =
+  | "about"
+  | "hide"
+  | "hideOthers"
+  | "unhide"
+  | "quit"
+  | "close"
+  | "undo"
+  | "redo"
+  | "cut"
+  | "copy"
+  | "paste"
+  | "selectAll"
+  | "reload"
+  | "toggleDevTools"
+  | "resetZoom"
+  | "zoomIn"
+  | "zoomOut"
+  | "togglefullscreen"
+  | "minimize"
+  | "zoom"
+  | "front"
+
 const labels: Record<MenuLocale, Record<MenuLabelKey, string>> = {
   en: {
     file: "File",
     edit: "Edit",
     view: "View",
+    window: "Window",
     go: "Go",
     help: "Help",
     checkForUpdates: "Check for Updates...",
@@ -55,6 +80,7 @@ const labels: Record<MenuLocale, Record<MenuLabelKey, string>> = {
     file: "文件",
     edit: "编辑",
     view: "视图",
+    window: "窗口",
     go: "前往",
     help: "帮助",
     checkForUpdates: "检查更新...",
@@ -75,6 +101,57 @@ const labels: Record<MenuLocale, Record<MenuLabelKey, string>> = {
     pawworkOnGithub: "PawWork 在 GitHub",
     reportProblem: "报告问题",
     openGithubIssue: "打开 GitHub Issue",
+  },
+}
+
+// Keep explicit English role labels so role-backed menu templates stay deterministic
+// in unit tests and non-macOS environments instead of depending on Electron runtime defaults.
+const roleLabels: Record<MenuLocale, Record<MenuRoleLabelKey, string>> = {
+  en: {
+    about: "About {appName}",
+    hide: "Hide {appName}",
+    hideOthers: "Hide Others",
+    unhide: "Show All",
+    quit: "Quit {appName}",
+    close: "Close Window",
+    undo: "Undo",
+    redo: "Redo",
+    cut: "Cut",
+    copy: "Copy",
+    paste: "Paste",
+    selectAll: "Select All",
+    reload: "Reload",
+    toggleDevTools: "Toggle Developer Tools",
+    resetZoom: "Actual Size",
+    zoomIn: "Zoom In",
+    zoomOut: "Zoom Out",
+    togglefullscreen: "Toggle Full Screen",
+    minimize: "Minimize",
+    zoom: "Zoom",
+    front: "Bring All to Front",
+  },
+  zh: {
+    about: "关于 {appName}",
+    hide: "隐藏 {appName}",
+    hideOthers: "隐藏其他",
+    unhide: "显示全部",
+    quit: "退出 {appName}",
+    close: "关闭窗口",
+    undo: "撤销",
+    redo: "重做",
+    cut: "剪切",
+    copy: "复制",
+    paste: "粘贴",
+    selectAll: "全选",
+    reload: "重新加载",
+    toggleDevTools: "切换开发者工具",
+    resetZoom: "实际大小",
+    zoomIn: "放大",
+    zoomOut: "缩小",
+    togglefullscreen: "切换全屏",
+    minimize: "最小化",
+    zoom: "缩放",
+    front: "全部移到前面",
   },
 }
 
@@ -126,4 +203,13 @@ export function menuLabel(locale: MenuLocale, key: MenuLabelKey) {
   if (value !== undefined) return value
   if (import.meta.env.DEV) console.warn("[menu] missing desktop label", { locale, key })
   return key
+}
+
+export function menuRoleLabel(locale: MenuLocale, key: MenuRoleLabelKey, appName: string) {
+  if (import.meta.env.DEV && locale !== "en" && roleLabels[locale]?.[key] === undefined) {
+    console.warn("[menu] missing locale role label, falling back to en", { locale, key })
+  }
+  const template = roleLabels[locale]?.[key] ?? roleLabels.en[key] ?? key
+  if (template === key && import.meta.env.DEV) console.warn("[menu] missing desktop role label", { locale, key })
+  return template.replaceAll("{appName}", appName)
 }
--- a/packages/desktop-electron/src/main/menu-template.ts
+++ b/packages/desktop-electron/src/main/menu-template.ts
@@ -1,4 +1,4 @@
-import { menuLabel, type MenuLocale } from "./menu-labels"
+import { menuLabel, menuRoleLabel, type MenuLocale, type MenuRoleLabelKey } from "./menu-labels"
 import { PAWWORK_GITHUB_ISSUE_URL, PAWWORK_GITHUB_URL } from "./support-links"
 
 export type MenuItemTemplate = {
@@ -31,6 +31,7 @@ type BuildMenuOptions = {
 export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[] {
   const { deps, appName, locale, feedbackEnabled } = options
   const t = (key: Parameters<typeof menuLabel>[1]) => menuLabel(locale, key)
+  const roleLabel = (key: MenuRoleLabelKey) => menuRoleLabel(locale, key, appName)
 
   const helpSubmenu: MenuItemTemplate[] = [
     { label: t("pawworkOnGithub"), click: () => deps.openExternal(PAWWORK_GITHUB_URL) },
@@ -47,7 +48,7 @@ export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[]
     {
       label: appName,
       submenu: [
-        { role: "about" },
+        { label: roleLabel("about"), role: "about" },
         {
           label: t("checkForUpdates"),
           click: () => deps.checkForUpdates(),
@@ -61,11 +62,11 @@ export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[]
           click: () => deps.relaunch(),
         },
         { type: "separator" },
-        { role: "hide" },
-        { role: "hideOthers" },
-        { role: "unhide" },
+        { label: roleLabel("hide"), role: "hide" },
+        { label: roleLabel("hideOthers"), role: "hideOthers" },
+        { label: roleLabel("unhide"), role: "unhide" },
         { type: "separator" },
-        { role: "quit" },
+        { label: roleLabel("quit"), role: "quit" },
       ],
     },
     {
@@ -75,19 +76,19 @@ export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[]
         { label: t("openProject"), accelerator: "Cmd+O", click: () => deps.trigger("project.open") },
         { label: t("newWindow"), accelerator: "Cmd+Shift+N", click: () => deps.newWindow() },
         { type: "separator" },
-        { role: "close" },
+        { label: roleLabel("close"), role: "close" },
       ],
     },
     {
       label: t("edit"),
       submenu: [
-        { role: "undo" },
-        { role: "redo" },
+        { label: roleLabel("undo"), role: "undo" },
+        { label: roleLabel("redo"), role: "redo" },
         { type: "separator" },
-        { role: "cut" },
-        { role: "copy" },
-        { role: "paste" },
-        { role: "selectAll" },
+        { label: roleLabel("cut"), role: "cut" },
+        { label: roleLabel("copy"), role: "copy" },
+        { label: roleLabel("paste"), role: "paste" },
+        { label: roleLabel("selectAll"), role: "selectAll" },
       ],
     },
     {
@@ -97,14 +98,14 @@ export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[]
         { label: t("toggleTerminal"), accelerator: "Ctrl+`", click: () => deps.trigger("terminal.toggle") },
         { label: t("toggleFileTree"), click: () => deps.trigger("fileTree.toggle") },
         { type: "separator" },
-        { role: "reload" },
-        { role: "toggleDevTools" },
+        { label: roleLabel("reload"), role: "reload" },
+        { label: roleLabel("toggleDevTools"), role: "toggleDevTools" },
         { type: "separator" },
-        { role: "resetZoom" },
-        { role: "zoomIn" },
-        { role: "zoomOut" },
+        { label: roleLabel("resetZoom"), role: "resetZoom" },
+        { label: roleLabel("zoomIn"), role: "zoomIn" },
+        { label: roleLabel("zoomOut"), role: "zoomOut" },
         { type: "separator" },
-        { role: "togglefullscreen" },
+        { label: roleLabel("togglefullscreen"), role: "togglefullscreen" },
       ],
     },
     {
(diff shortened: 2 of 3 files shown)


packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -120,7 +121,15 @@ export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[]
         { label: t("nextProject"), accelerator: "Cmd+Option+Down", click: () => deps.trigger("project.next") },
       ],
     },
-    { role: "windowMenu" },
+    {
+      label: t("window"),
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Restored `role: "windowMenu"` on the localized Window menu parent and added test coverage so the template keeps that parent role while still listing the localized standard Window items.

packages/desktop-electron/src/main/menu.test.ts
```diff
@@ -36,6 +46,48 @@ describe("desktop menu template", () => {
     expect(labels(template)).toContain("帮助")
   })
```
Astro-Han, line 137: **P1**: This test only validates Chinese (locale: zh) role labels. Missing test coverage for English role labels (locale: en). Consider adding a parallel test case for English to ensure both locales are validated.

packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -120,7 +121,16 @@ export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[]
         { label: t("nextProject"), accelerator: "Cmd+Option+Down", click: () => deps.trigger("project.next") },
       ],
     },
-    { role: "windowMenu" },
+    {
```
Astro-Han, line 124: **P2**: Expanding `{ role: "windowMenu" }` into a manual submenu may lose Electron's default Window menu behavior. On macOS, `windowMenu` role typically includes additional items like "Merge All Windows" when applicable. Verify that this matches the intended macOS Window menu exactly.

packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -31,6 +31,7 @@ type BuildMenuOptions = {
 export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[] {
   const { deps, appName, locale, feedbackEnabled } = options
   const t = (key: Parameters<typeof menuLabel>[1]) => menuLabel(locale, key)
+  const tr = (key: Parameters<typeof menuRoleLabel>[1]) => menuRoleLabel(locale, key, appName)
```
Astro-Han, line 35: **P2**: The helper function `tr` naming is ambiguous compared to `t`. Consider using a more explicit name like `trLabel` or `roleLabel` to clarify that this function handles role-based label translation.

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -127,3 +202,10 @@ export function menuLabel(locale: MenuLocale, key: MenuLabelKey) {
   if (import.meta.env.DEV) console.warn("[menu] missing desktop label", { locale, key })
   return key
 }
+
+export function menuRoleLabel(locale: MenuLocale, key: MenuRoleLabelKey, appName: string) {
+  const value = roleLabels[locale]?.[key] ?? roleLabels.en[key]
+  if (value !== undefined) return value.replaceAll("{appName}", appName)
```
Astro-Han: **P3**: Missing type annotation for `appName` parameter. Add explicit `: string` for better readability and type safety.

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -127,3 +202,10 @@ export function menuLabel(locale: MenuLocale, key: MenuLabelKey) {
...
+  if (import.meta.env.DEV) console.warn("[menu] missing desktop role label", { locale, key })
+  return key
```
Astro-Han: **P3**: Inconsistent fallback behavior with `menuLabel` function. `menuLabel` returns raw key string, while `menuRoleLabel` returns key but the DEV warning path never applies `replaceAll`. Consider aligning the fallback behavior or documenting the difference.

Astro-Han (commented):
## Code Review Summary

**P1** (required): 1 issue
**P2** (suggested): 2 issues
**P3** (optional): 2 issues

Overall the implementation is solid, but a few areas need attention before merge.

Astro-Han:
@coderabbitai review

packages/desktop-electron/src/main/menu-template.ts
Astro-Han: Fixed. Renamed the helper from `tr` to `roleLabel` so the distinction from `t` is explicit at each call site.

packages/desktop-electron/src/main/menu.test.ts
Astro-Han: Fixed. Added a parallel English role-label test that checks the same role-backed menu surfaces as the Chinese case, including the Window menu parent role.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. `menuRoleLabel()` now follows the same fallback shape more closely by resolving through `?? key`, emitting the DEV warning on the raw-key fallback, and applying the final `replaceAll()` on the returned template path.

packages/desktop-electron/src/main/menu-template.ts
Astro-Han: Checked this locally on macOS with the current head. Keeping the parent `role: "windowMenu"` preserves the native Window menu behavior: the running app still shows native-only items beyond our static entries, including `全部最小化`, `全部缩放`, `填充`, `居中`, `移动与调整大小`, `全屏幕平铺`, `从组中移除窗口`, `排在前面`, plus the dynamic `PawWork` window entry. So I am not changing the structure further here.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Current head already has the explicit type annotation: `menuRoleLabel(locale: MenuLocale, key: MenuRoleLabelKey, appName: string)`.

packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -120,7 +121,16 @@ export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[]
         { label: t("nextProject"), accelerator: "Cmd+Option+Down", click: () => deps.trigger("project.next") },
       ],
     },
-    { role: "windowMenu" },
+    {
+      label: t("window"),
+      role: "windowMenu",
```
Astro-Han, line 130: **P1**: Setting `role: "windowMenu"` alongside a custom `submenu` contradicts [Electron's documented behavior](https://www.electronjs.org/docs/latest/api/menu-item#roles):

> When using a role on macOS, the label and accelerator are the only properties that will modify the behavior of the menu item, all other properties will be ignored.

This means `submenu` could be silently ignored on macOS, and the Window menu items would NOT be localized despite the `label` on each item. Even if the current Electron version happens to merge both (as your manual test suggests), this is undocumented behavior that could break on any Electron upgrade.

Two options:
1. Add an inline comment at this block documenting the assumption and the tested Electron version, so a future breakage is traceable.
2. Remove `role: "windowMenu"` from the parent and add the dynamic window list via Electron's `BrowserWindow.getAllWindows()` or rely on macOS appending it natively. This would make the localization explicit and not depend on undocumented behavior.

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -78,6 +104,55 @@ const labels: Record<MenuLocale, Record<MenuLabelKey, string>> = {
   },
 }
 
+const roleLabels: Record<MenuLocale, Record<MenuRoleLabelKey, string>> = {
```
Astro-Han, line 109: **P2**: The entire `roleLabels.en` map duplicates Electron's built-in English defaults. On English locale, `{ label: "Undo", role: "undo" }` is functionally identical to `{ role: "undo" }`. This creates a maintenance coupling: if Electron changes a default label in a future version, PawWork's explicit label would override it and could look inconsistent with other macOS apps.

Consider either:
1. Skipping the `label` assignment when `locale === "en"` (let Electron use its defaults), or
2. Adding a comment explaining the English labels exist only for the fallback path and are intentionally decoupled from Electron defaults.

packages/desktop-electron/src/main/menu.test.ts
```diff
@@ -21,6 +21,102 @@ function submenu(template: MenuItemTemplate[], label: string) {
   return template.find((item) => item.label === label)?.submenu ?? []
 }
 
+function menuItem(template: MenuItemTemplate[], label: string) {
+  return template.find((item) => item.label === label)
+}
+
+function expectRoleLabels(items: MenuItemTemplate[], expected: Record<string, string>) {
+  for (const [role, label] of Object.entries(expected)) {
```
Astro-Han, line 33: **P2**: `expectRoleLabels` only checks that expected items are *contained* in the array (`toContainEqual`), but does not assert that no extra items exist. If an unintended role-backed item is accidentally added to a submenu, this test would not catch it.

Consider adding an assertion on item count or filtering for role-bearing items and comparing the full set:
```ts
const roleItems = items.filter(i => i.role !== undefined)
expect(roleItems).toHaveLength(Object.keys(expected).length)
```

packages/desktop-electron/src/main/menu.test.ts
```diff
@@ -21,6 +21,102 @@ function submenu(template: MenuItemTemplate[], label: string) {
   return template.find((item) => item.label === label)?.submenu ?? []
 }
```
Astro-Han, line 24: **P2**: `submenu()` returns `[]` when no item matches the label. When `expectRoleLabels([], expected)` is called with a non-empty `expected`, the failure message will be something like `Expected [] to contain object { role: "about", label: "About PawWork" }`, which is confusing and doesn't indicate that the *parent menu* was not found.

Consider asserting the menu exists before iterating:
```ts
const items = submenu(template, appName)
expect(items.length, `menu '${appName}' not found`).toBeGreaterThan(0)
```

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -78,6 +104,55 @@ const labels: Record<MenuLocale, Record<MenuLabelKey, string>> = {
...
+    selectAll: "全选",
+    reload: "重新加载",
+    toggleDevTools: "切换开发者工具",
+    resetZoom: "实际大小",
+    zoomIn: "放大",
+    zoomOut: "缩小",
+    togglefullscreen: "切换全屏",
+    minimize: "最小化",
+    zoom: "缩放",
+    front: "全部移到最前面",
```
Astro-Han: **P3**: macOS Sonoma/Sequoia uses "全部移到前面" for "Bring All to Front". The extra "最" (meaning "most") makes this "Bring All to the *Very* Front", which deviates from the platform-standard translation. Minor, but native users may notice the inconsistency.

packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -31,6 +31,7 @@ type BuildMenuOptions = {
 export function buildMenuTemplate(options: BuildMenuOptions): MenuItemTemplate[] {
   const { deps, appName, locale, feedbackEnabled } = options
   const t = (key: Parameters<typeof menuLabel>[1]) => menuLabel(locale, key)
+  const roleLabel = (key: Parameters<typeof menuRoleLabel>[1]) => menuRoleLabel(locale, key, appName)
```
Astro-Han: **P3**: `Parameters<typeof menuRoleLabel>[1]` is an indirect type reference that requires the reader to look up `menuRoleLabel`'s signature to understand what type is accepted. Consider importing and using `MenuRoleLabelKey` directly for readability, consistent with how `MenuLocale` is imported explicitly:
```ts
import { menuLabel, menuRoleLabel, type MenuLocale, type MenuRoleLabelKey } from "./menu-labels"
// ...
const roleLabel = (key: MenuRoleLabelKey) => menuRoleLabel(locale, key, appName)
```

packages/desktop-electron/src/main/menu-labels.ts
```diff
@@ -127,3 +202,9 @@ export function menuLabel(locale: MenuLocale, key: MenuLabelKey) {
   if (import.meta.env.DEV) console.warn("[menu] missing desktop label", { locale, key })
   return key
 }
+
+export function menuRoleLabel(locale: MenuLocale, key: MenuRoleLabelKey, appName: string) {
```
Astro-Han, line 208: **P3**: When a key exists in `roleLabels.en` but is missing from `roleLabels.zh`, the fallback to English is silent. No DEV warning fires because `template !== key` (the English template is found). This makes incomplete zh translations invisible during development. Consistent with `menuLabel`'s existing behavior, but worth noting as a missed opportunity to improve the pattern in new code.

If desired, a locale-coverage warning could be added:
```ts
if (import.meta.env.DEV && locale !== 'en' && roleLabels[locale]?.[key] === undefined) {
  console.warn('[menu] missing locale role label, falling back to en', { locale, key })
}
```

Astro-Han (commented):
## Review: fix: localize macOS menu role labels

Thorough nitpick review. The approach is sound but a few issues worth addressing.

**P1** (1 issue): windowMenu role + custom submenu contradicts Electron docs
**P2** (3 issues): redundant English labels, loose test assertions, silent submenu miss
**P3** (3 issues): localization accuracy, readability, locale-coverage gap

Key concern: the windowMenu block sets both role and submenu. Electron docs say role takes precedence and other properties are ignored, which would silently discard the custom submenu and defeat the localization goal. Needs either a code comment documenting the tested behavior, or a restructure to avoid the conflict.

packages/desktop-electron/src/main/menu-template.ts
Astro-Han: Fixed via option 1. Added an inline comment at the `windowMenu` block documenting the Electron 40.8.0 / macOS 15 assumption and the fallback plan if a future Electron release stops honoring the merged submenu labels.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. Kept the explicit English role labels, but documented why they exist: they keep role-backed menu templates deterministic in unit tests and non-macOS environments instead of depending on Electron runtime defaults.

packages/desktop-electron/src/main/menu.test.ts
Astro-Han: Fixed. `expectRoleLabels()` now filters the role-bearing items first and asserts the exact count before checking the expected role/label pairs.

packages/desktop-electron/src/main/menu.test.ts
Astro-Han: Fixed. `submenu()` now asserts that the parent menu exists before returning its submenu, so a missing top-level menu fails with a direct `menu ... not found` message instead of a downstream empty-array mismatch.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. Added a DEV-only warning when a non-English locale falls back to the English role label map, so incomplete locale coverage is visible during development.

packages/desktop-electron/src/main/menu-labels.ts
Astro-Han: Fixed. Updated the zh translation for `front` from `全部移到最前面` to the platform-standard `全部移到前面`.

packages/desktop-electron/src/main/menu-template.ts
Astro-Han: Fixed. Imported `MenuRoleLabelKey` directly and used it for the `roleLabel()` helper parameter so the accepted type is visible at the call site.

Issue #189, opened 2026-04-23
[Bug] Solid HMR loses context providers after bulk edits to high-fanout modules


Astro-Han:
## Symptoms

When the dev server is running, a large bulk edit to a high-fanout module, for example `packages/ui/src/components/icon.tsx`, can leave some components detached from their context providers during HMR. The browser then reports errors such as:

```text
Error: Layout context must be used within a context provider
    at use (packages/ui/src/context/helper.tsx:37)
    at DialogSelectFile2 (packages/app/src/components/dialog-select-file.tsx:197)
    at HMRComp.createMemo.name (@solid-refresh:25)
```

The same branch previously reproduced a similar `useGlobalSync must be used within GlobalSyncProvider` error. The failing provider changed, but the symptom pattern was the same.

## Reproduction

1. Run `bun dev:desktop`.
2. Make a large bulk edit to `packages/ui/src/components/icon.tsx`, for example replacing 70+ SVG entries in one change.
3. Open a page that triggers a dialog, overlay, or similar re-render.

## Current Workaround

Stop the dev server with `Ctrl+C`, then restart it with `bun dev:desktop`. A cold restart immediately recovers the app.

This appears to affect only the development loop. Production builds and typecheck still pass.

## Suspected Cause

`@solid-refresh` may recreate referenced components after a large module-level constant change, while the parent provider instance is not re-wired correctly. App-root providers under `packages/app/src/context/*` should not be affected by a leaf icon module edit, but they appear to become detached through HMR.

## Expected Behavior

The dev server should not lose app-root context providers after a leaf or high-fanout UI module changes.

If this is a Solid HMR limitation, a development-only full reload or dev overlay warning would be acceptable.

## Priority

P3. This does not affect production, but it can disrupt the development loop during large UI refactors.

Astro-Han:
Latest verification on `6501ab650`: leaving this open.

I did not get enough new evidence to close it safely:
- the failure mode still depends on a large hot-reload edit to a high-fanout module,
- I did not find a deterministic regression harness or targeted E2E for this exact provider-loss path on current `dev`,
- and this pass was not enough to prove the bug is gone.

So I am treating this as unresolved rather than silently closing it. If someone can reproduce it again on current `dev`, please add the exact edited file, console error, and whether it was `dev:web` or `dev:desktop`.

Issue #181, opened 2026-04-23
[Feature] Home visual polish to match mockup


Astro-Han:
## Context

#151 home redesign shipped in #180 covers layout, Skill card icons, composer chrome slimming, and the Skill starter reshape. Several visual details still diverge from the #151 mockup and are intentionally out of scope for #180 to keep the PR reviewable.

Mockup reference: https://github.com/user-attachments/assets/ebf59723-1642-4990-8367-4439005598ab (and the iterated ChatGPT mockups shared during #180 review).

## Gaps

- [ ] **Composer Send button**: mockup uses an orange-filled circle (brand color). Current build uses the default grey-ish IconButton `variant="primary"`. Touches `packages/app/src/components/prompt-input.tsx` Send region; likely needs a brand variant on `@opencode-ai/ui/icon-button` or a local wrapper.
- [ ] **Composer model chip**: mockup shows `Kimi K2.6` as plain text + chevron with no provider icon. Current build prepends the Big Pickle green "Z" provider icon via `ProviderIcon`. Decide whether to drop the provider icon in home mode only or also in session.
- [ ] **Composer variant chip**: mockup uses short label `中` (one character). Current uses `默认` (two chars). Update `session.new.*` or `common.default` i18n, or switch variant default label rule when `homeMode`.
- [ ] **Composer interior spacing**: extra `py-` inside the bottom bar so chips feel less crowded next to Send.
- [ ] **WorkspaceChip dropdown**: current popover (工作目录 list + 添加工作目录) feels boxy; mockup has no reference so needs a separate design pass. Touches `packages/app/src/components/prompt-input/workspace-chip.tsx`.
- [ ] **Sidebar icon set**: the five nav entries (新建对话 / 搜索 / 插件 / 远程控制 / 自动任务) should use a visually unified icon family. Current icons come from `packages/ui/src/components/icon.tsx` but have mixed stroke weights. Either pick a consistent subset or add locally tuned SVGs for sidebar-specific entries.

## Out of scope

- `session.skill` DB/API cleanup (stays for legacy).
- Session-level Skill indicator, title generation changes.
- Right-panel work (tracked in #154).

## Suggested slicing

Open follow-up PRs in this order so each slice has a clear diff:

1. Send brand button + model/variant chip copy + inner spacing (single `prompt-input.tsx` change)
2. WorkspaceChip dropdown polish (single file + workspace-chip-helpers.ts if needed)
[...]

PR #191, opened 2026-04-23
feat: PawWork brand chrome icon set v2 (issue #181 slice 3)

Astro-Han:
## Summary
- Replace the chrome icon set in `packages/ui` with 80 PawWork-drawn SVGs (74 rebranded + 6 new: `chevron-up` / `plugin` / `automation` / `remote-control` / `schedule` / `key`).
- Delete `skill-icons.tsx`; Skill cards + workspace chip now use the shared `<Icon>` component.
- Wire Review / Terminal tab `-active` variants; fix Open-file menu icon and sidebar search icon size mismatch.
- Polish home Skill pills: equal-width 3-col grid, 20×20 icon, native EN labels (Process docs / Analyze data / Start writing).

Closes part of #181 (slice 3: sidebar icon set alignment).

## Test plan
- [x] `bun turbo typecheck --filter=@opencode-ai/ui --filter=@opencode-ai/app` passes
- [x] Manual: Chinese + English Skill pills render equal width with new icons
- [x] Manual: Review / Terminal tabs show `-active` glyph when selected
- [x] Manual: Search icon matches New-session icon size in sidebar
- [x] Manual: Open-file menu item no longer duplicates Open-project's icon
- [ ] Dev HMR context-loss observed during bulk edit; logged separately as #189 (orthogonal)

## Carve-out note
`packages/ui/src/components/icon.tsx` is now **accept-HEAD** on upstream sync. When upstream introduces new icons or renames, add matching PawWork paths locally; do not replace the file wholesale.

## Summary by CodeRabbit

* **New Features**
  * Side panel tabs now show active-state icons to indicate the selected tab.

* **UI Improvements**
  * Skill cards use a responsive grid (single-column on small, three columns on larger screens).
  * Icon usage unified across the app for consistent visuals (workspace, chevron, plus, pin, skill icons).
  * Sidebar pin button styling updated for better contrast.
  * File dropdown icon updated.

* **Text Updates**
  * Session card titles changed to “Process docs” and “Start writing”.


packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -145,12 +131,12 @@ export const PawworkSidebar = (props: {
                 }}
                 classList={{
                   "inline-flex size-6 items-center justify-center rounded transition-colors": true,
-                  "text-accent-brand opacity-100 pointer-events-auto": isPinned(),
+                  "text-text-strong opacity-100 pointer-events-auto": isPinned(),
```
Astro-Han, line 134: [P3] This weakens the pinned-state affordance. The old row used both a filled pin and the accent color when `isPinned()` was true; now pinned rows use the same `pin` glyph family and only `text-text-strong`, which is much easier to miss in a mixed list. Please keep either an active/filled pin variant or the accent token for the pinned state.

packages/app/src/components/session/session-new-view.tsx
```diff
@@ -23,7 +23,7 @@ export function NewSessionView(props: { composer?: (ctx: ComposerCtx) => JSX.Ele
         <h1 class="text-20-medium text-text-strong">{language.t("session.new.title")}</h1>
         <p class="mt-3 text-14-regular text-text-weak">{language.t("session.new.subtitle")}</p>
 
-        <div class="mt-8 flex w-full max-w-[640px] flex-wrap justify-center gap-3">
+        <div class="mt-8 grid w-fit max-w-[640px] grid-cols-3 gap-3">
```
Astro-Han: [P2] This fixed three-column fit-content grid drops the old wrapping behavior. At narrow desktop widths or in the mobile layout, the three buttons keep their content width plus padding and gaps, so the home screen can overflow horizontally instead of wrapping. Please keep the equal-width desktop layout but let smaller widths fall back to one or two columns, or restore wrapping.

Astro-Han:
Placeholder for inline review

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -26,21 +27,6 @@ const FilterIcon = (props: { size?: number }) => {
   )
 }
 
-const PinIcon = (props: { filled?: boolean; size?: number }) => {
```
Astro-Han, line 29: [P1] test comment

Astro-Han (commented):
Icon set v2 review - see inline comments

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -145,12 +131,12 @@ export const PawworkSidebar = (props: {
...
                   "inline-flex size-6 items-center justify-center rounded transition-colors": true,
-                  "text-accent-brand opacity-100 pointer-events-auto": isPinned(),
+                  "text-text-strong opacity-100 pointer-events-auto": isPinned(),
                   "text-text-weak opacity-0 pointer-events-none group-hover/session:opacity-100 group-hover/session:pointer-events-auto group-focus-within/session:opacity-100 group-focus-within/session:pointer-events-auto hover:text-text-base":
                     !isPinned(),
                 }}
               >
-                <PinIcon filled={isPinned()} />
+                <Icon name="pin" size="small" />
               </button>
```
Astro-Han, line 140: [P1] **Pin icon lost filled/outline differentiation.** The old `PinIcon` supported a `filled` prop: pinned sessions showed a solid pin, unpinned showed an outline pin. The new `<Icon name="pin">` is always a filled glyph with no outline variant. Combined with the color swap from `text-accent-brand` to `text-text-strong` (nearby line), the pinned affordance is significantly weakened. A user can no longer distinguish pinned from unpinned by icon shape alone; they must rely entirely on the color class difference, which is also weaker now (`text-text-strong` vs `text-text-weak` are closer in contrast than `text-accent-brand` vs `text-text-weak`).

Suggest: (a) add a `pin-outline` icon (or `pin-active`) and wire `<Icon name={isPinned() ? "pin" : "pin-outline"}>`, or (b) at minimum restore `text-accent-brand` for the pinned class so color still clearly communicates the toggle.

packages/app/src/components/session/pawwork-skill-meta.ts
```diff
@@ -1,28 +1,28 @@
...
   },
   {
     name: "writing-assistant",
     iconName: "pencil-line" as IconProps["name"],
-    homeIcon: "pencil" as SkillIconName,
-    homeIconClass: "text-violet-500",
+    homeIcon: "pencil-line" as IconProps["name"],
+    homeIconClass: "",
+    homeIconStyle: { color: "#8B5FBF" },
     titleKey: "session.new.card.writing.title",
```
Astro-Han, line 40: [P1] **Hardcoded hex color breaks theme system.** `#8B5FBF` is injected via inline `style`, which cannot be overridden by CSS custom properties and does not respond to dark/light mode token switches. It is also invisible to design-token grep. Prefer adding a semantic Tailwind utility class like `text-icon-accent-writing` (or use `text-violet-500` if the shade must stay in the existing palette) and drop `homeIconStyle` entirely. If the exact shade differs from an existing token, define a named design token so the palette stays centralized.

packages/app/src/components/session/session-new-view.tsx
```diff
@@ -33,14 +33,19 @@ export function NewSessionView(props: { composer?: (ctx: ComposerCtx) => JSX.Ele
...
                   }}
                   onClick={() => toggleSkill(card.name)}
                 >
-                  <SkillIcon name={card.homeIcon} class={`h-6 w-6 shrink-0 ${card.homeIconClass}`} />
+                  <Icon
+                    name={card.homeIcon}
+                    size="normal"
+                    class={`shrink-0 ${card.homeIconClass ?? ""}`}
+                    style={"homeIconStyle" in card ? card.homeIconStyle : undefined}
+                  />
```
Astro-Han, line 48: [P1] **Leaking implementation detail through template.** The conditional `"homeIconStyle" in card` check in JSX couples the view layer to the shape of the data model. If a future card adds any property named `homeIconStyle` (even with an `undefined` value), it will silently pass a broken style. Consider: (a) making `homeIconStyle` an optional field on the card type (with `?`), and using `card.homeIconStyle` directly in the Icon `style` prop, or (b) moving the violet color into the `homeIconClass` field and eliminating `homeIconStyle` altogether (as suggested in the P1 on pawwork-skill-meta.ts).

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -145,12 +131,12 @@ export const PawworkSidebar = (props: {
                 }}
                 classList={{
                   "inline-flex size-6 items-center justify-center rounded transition-colors": true,
-                  "text-accent-brand opacity-100 pointer-events-auto": isPinned(),
+                  "text-text-strong opacity-100 pointer-events-auto": isPinned(),
                   "text-text-weak opacity-0 pointer-events-none group-hover/session:opacity-100 group-hover/session:pointer-events-auto group-focus-within/session:opacity-100 group-focus-within/session:pointer-events-auto hover:text-text-base":
```
Astro-Han, line 135: [P2] **Pinned-session color downgrade.** The old `text-accent-brand` provided a distinct brand-color signal for pinned sessions. The new `text-text-strong` uses the same color as regular active text, making pinned sessions indistinguishable from focused/hovered sessions at a glance. This weakens the information hierarchy. If keeping the brand accent is intentional (to reduce visual noise), consider at least using `text-text-base` for unpinned and `text-text-strong` for pinned so there is *some* contrast step.

packages/app/src/pages/session/session-side-panel.tsx
```diff
@@ -82,10 +82,10 @@ function RightPanelShellIcon(props: { icon: ShellTabIcon }) {
         <Icon name="folder" size="small" class="text-text-weaker" />
       </Match>
       <Match when={props.icon.kind === "icon" && props.icon.name === "review"}>
-        <Icon name="review" size="small" class="text-text-weaker" />
+        <Icon name={props.active ? "review-active" : "review"} size="small" class="text-text-weaker" />
       </Match>
```
Astro-Han, line 86: [P2] **Active-variant icon coupling could grow fragile.** The `RightPanelShellIcon` now branches on `props.active` to select `review-active` vs `review` (and `terminal-active` vs `terminal`). This works today because only two icon kinds need active variants, but as more tab icons gain active states, the Switch/Match will grow. Consider making the active suffix a convention: `<Icon name={props.active ? \`\${icon.name}-active\` : icon.name} />` so every future active variant is picked up automatically without touching this component.

packages/app/src/components/session/session-new-view.tsx
```diff
@@ -23,7 +23,7 @@ export function NewSessionView(props: { composer?: (ctx: ComposerCtx) => JSX.Ele
         <h1 class="text-20-medium text-text-strong">{language.t("session.new.title")}</h1>
         <p class="mt-3 text-14-regular text-text-weak">{language.t("session.new.subtitle")}</p>
 
-        <div class="mt-8 flex w-full max-w-[640px] flex-wrap justify-center gap-3">
+        <div class="mt-8 grid w-fit max-w-[640px] grid-cols-3 gap-3">
           <For each={pawworkSkillCards}>
```
Astro-Han, line 27: [P2] **`grid-cols-3 w-fit` may cause unequal pill widths at smaller viewports.** With `w-fit`, the grid container shrinks to content. If the three pill labels have different widths (e.g., after i18n), the columns will be equal but the overall grid width may not fill the available space, leaving asymmetric margins. The old `flex flex-wrap justify-center gap-3` naturally centered pills regardless of content width. Verify that the three-column grid still looks balanced with CJK labels (zh: 处理文档 / 分析数据 / 开始写作) since those labels have different character counts.

packages/ui/src/components/icon.tsx
```diff
@@ -2,106 +2,115 @@ import { splitProps, type ComponentProps } from "solid-js"
...
+  "plugin": `<g transform="translate(3.542 1.000) scale(0.101695)"><path d="M0,0 L7,1 L8,3 L8,36 L2,39 L-3,35 L-3,3 Z " transform="translate(26.000 0.000)" fill="currentColor"/><path d="M0,0 L5,0 L8,3 L8,35 L4,39 L-1,38 L-3,35 L-3,3 Z " transform="translate(96.000 0.000)" fill="currentColor"/><path d="M0,0 L117,0 L122,5 L122,57 L118,71 L110,82 L103,89 L89,96 L85,97 L65,98 L64,130 L60,133 L55,132 L53,129 L53,97 L34,97 L22,93 L11,86 L3,77 L-3,65 L-5,55 L-5,5 Z M7,11 L6,12 L6,37 L7,58 L11,68 L15,74 L23,81 L31,85 L34,86 L84,86 L96,80 L105,71 L110,59 L111,53 L111,12 L110,11 Z " transform="translate(5.000 44.000)" fill="currentColor"/><path d="M0,0 L50,0 L52,3 L52,8 L48,11 L2,11 L-2,9 L-3,4 Z " transform="translate(39.000 74.000)" fill="currentColor"/></g>`,
```
Astro-Han, line 104: [P2] **Two icon entries on the same line.** `pin` and `chevron-up` are declared on the same line (`pin: ..., "chevron-up": ...`). This makes them harder to grep, harder to diff, and breaks the one-entry-per-line pattern every other icon follows. Please split onto separate lines.

packages/ui/src/components/icon.tsx
```diff
@@ -2,106 +2,115 @@ import { splitProps, type ComponentProps } from "solid-js"
...
+  "bar-chart": `<g transform="translate(1.000 1.959) scale(0.091371)"><path d="M0,0 L31,0 L37,3 L40,7 L40,166 L53,166 L56,168 L57,172 L53,176 L-137,176 L-140,173 L-139,168 L-136,166 L-123,166 L-123,112 L-120,107 L-116,105 L-90,105 L-85,108 L-82,112 L-82,166 L-67,166 L-67,54 L-62,48 L-59,47 L-33,47 L-28,50 L-25,55 L-25,166 L-10,166 L-10,9 L-7,3 Z M0,10 L0,166 L30,166 L30,10 Z M-56,56 L-57,57 L-57,166 L-35,166 L-35,57 L-36,56 Z M-113,114 L-114,115 L-114,166 L-92,166 L-92,115 L-93,114 Z " transform="translate(140.000 0.000)" fill="currentColor"/></g>`,
+
+
+
+
```
Astro-Han: [P3] **Four consecutive blank lines between icon entries.** Between `bar-chart` and `pin` there are four empty lines where one (or zero) would suffice. Likely an artifact of bulk editing; please clean up for consistency with the rest of the file.

packages/app/src/i18n/en.ts
```diff
@@ -589,12 +589,12 @@ export const dict = {
   "session.new.title": "Choose what to do",
   "session.new.subtitle": "PawWork helps you process files, analyze information, write content, and tackle everyday tasks.",
   "session.new.reassurance": "Files and conversations stay on your computer",
-  "session.new.card.document.title": "Process documents",
+  "session.new.card.document.title": "Process docs",
   "session.new.card.document.description": "Edit, convert, and extract from Word, Excel, PowerPoint, and PDF files.",
```
Astro-Han, line 620: [P3] **Label shortening may hurt clarity.** "Process documents" -> "Process docs" and "Write faster" -> "Start writing" change the semantics: "Process docs" is more casual but also vaguer (docs could mean documentation, not documents/files), and "Start writing" shifts from a benefit ("faster") to an action ("start"). The PR description says "native EN labels" but the original labels were already native English. If the shortening was needed to fit the equal-width pill grid, consider whether the grid constraint is worth the clarity trade-off.

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -264,7 +250,7 @@ export const PawworkSidebar = (props: {
             size="large"
             variant="ghost"
             icon="magnifying-glass"
-            class="w-full [&>[data-component=icon]]:!size-5"
+            class="w-full"
             onClick={props.onSearch}
```
Astro-Han, line 254: [P3] **Search button icon size change is implicit.** Removing the `[&>[data-component=icon]]:!size-5` override means the magnifying-glass icon will now render at whatever size the `IconButton` component defaults to. The PR description says this fixes a size mismatch with the new-session button, but the reasoning is not self-documenting in the code. Consider adding a brief code comment explaining why the override is no longer needed (e.g., the new icon set is designed for the default size), so a future reader does not re-add it thinking it was accidental.

Astro-Han (commented):
Detailed icon-set review with inline comments, sorted by priority.

Summary:
- 3x P1 (pin icon affordance, hardcoded hex color, leaked impl detail)
- 4x P2 (color downgrade, active-variant coupling, grid i18n, same-line entry)
- 3x P3 (blank lines, label clarity, implicit size change)

packages/app/src/components/session/pawwork-skill-meta.ts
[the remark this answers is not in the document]
Astro-Han: This repo runs Tailwind v4 with `--color-*: initial` in `packages/app/src/index.css`, which drops the default palette. `text-violet-500` resolves to no CSS variable and renders as the inherited foreground, so the accent disappears. Inline hex is the working form until we introduce a named brand token. Added an inline comment in pawwork-skill-meta.ts documenting the constraint so the next reviewer doesn't re-suggest it.

packages/app/src/components/session/session-new-view.tsx
```diff
@@ -33,14 +33,19 @@ export function NewSessionView(props: { composer?: (ctx: ComposerCtx) => JSX.Ele
...
                     "border-border-interactive-base bg-surface-raised-base shadow-sm": isSelected(),
                   }}
                   onClick={() => toggleSkill(card.name)}
                 >
-                  <SkillIcon name={card.homeIcon} class={`h-6 w-6 shrink-0 ${card.homeIconClass}`} />
+                  <Icon
+                    name={card.homeIcon}
+                    size="normal"
+                    class={`shrink-0 ${card.homeIconClass ?? ""}`}
+                    style={"homeIconStyle" in card ? card.homeIconStyle : undefined}
```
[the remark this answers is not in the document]
Astro-Han: Kept `style` for the writing card's accent — the underlying hex can't be expressed as a Tailwind utility here (see the parallel thread on `pawwork-skill-meta.ts`). Reshaped the card type so `homeIconStyle` is an optional `JSX.CSSProperties` field, and the JSX now passes `card.homeIconStyle` directly without the `in` narrowing.

packages/app/src/components/session/pawwork-skill-meta.ts
```diff
@@ -1,28 +1,28 @@
...
     descriptionKey: "session.new.card.analysis.description",
   },
   {
     name: "writing-assistant",
     iconName: "pencil-line" as IconProps["name"],
-    homeIcon: "pencil" as SkillIconName,
-    homeIconClass: "text-violet-500",
+    homeIcon: "pencil-line" as IconProps["name"],
+    homeIconClass: "",
+    homeIconStyle: { color: "#8B5FBF" },
```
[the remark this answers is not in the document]
Astro-Han: Same as the Gemini thread on this file — Tailwind v4 in this repo disables the default palette via `--color-*: initial`, so `text-violet-500` is a no-op. Kept the inline hex, added a comment documenting why palette utilities can't be used here.

packages/app/src/pages/layout/pawwork-sidebar.tsx
```diff
@@ -145,12 +131,12 @@ export const PawworkSidebar = (props: {
...
                 classList={{
                   "inline-flex size-6 items-center justify-center rounded transition-colors": true,
-                  "text-accent-brand opacity-100 pointer-events-auto": isPinned(),
+                  "text-text-strong opacity-100 pointer-events-auto": isPinned(),
                   "text-text-weak opacity-0 pointer-events-none group-hover/session:opacity-100 group-hover/session:pointer-events-auto group-focus-within/session:opacity-100 group-focus-within/session:pointer-events-auto hover:text-text-base":
                     !isPinned(),
                 }}
               >
-                <PinIcon filled={isPinned()} />
+                <Icon name="pin" size="small" />
```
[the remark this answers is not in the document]
Astro-Han: Verified on the current file: `pin` is at `packages/ui/src/components/icon.tsx:102` and `workspace` at `:99`. The registry uses bare keys (`pin:` / `workspace:`), not quoted (`"pin":`), so the `rg "\"${key}\""` search in the analysis script missed them. Other five-hyphen keys like `review-active` matched because TypeScript quotes hyphenated identifiers. Runtime usage also verified — `<Icon name="pin">` and `<Icon name="workspace">` render fine in the dev build.

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: The pinned affordance in the new layout is carried by the visibility gate, not the glyph: unpinned rows have the pin button at `opacity-0 pointer-events-none` and only reveal it on row hover/focus, while pinned rows stay at `opacity-100 pointer-events-auto` permanently (pawwork-sidebar.tsx:134-137). The always-visible vs only-on-hover differential reads at a glance in a scanning pattern, which is stronger than filled-vs-outline would be at 16×16. The color drop from `text-accent-brand` to `text-text-strong` was an intentional design call earlier in this worktree to reduce visual noise in the session list.

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: The toggle is now carried by the visibility gate rather than the glyph shape. At pawwork-sidebar.tsx:134-137, unpinned rows have the pin button at `opacity-0 pointer-events-none` with an on-hover/focus reveal, and pinned rows pin it at `opacity-100 pointer-events-auto`. So: pinned rows show a pin icon all the time; unpinned rows show no pin icon unless the row is hovered. That's a stronger scanning signal than filled-vs-outline at 16×16. The `text-accent-brand` drop was an intentional quieter-chrome choice earlier in this branch — the memory of that conversation: "不要做填充吧，而且现在填充是品牌色".

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: The hue swap isn't the load-bearing signal here. Pinned vs unpinned is carried by the opacity gate: unpinned buttons sit at `opacity-0 pointer-events-none` and reveal on row hover/focus, pinned buttons are pinned at `opacity-100 pointer-events-auto` (pawwork-sidebar.tsx:134-137). `text-text-strong` on pinned matches the adjacent session title's foreground so the pin doesn't draw extra attention once persistently shown — the visual noise reduction was the point of dropping the brand accent.

packages/app/src/components/session/session-new-view.tsx
Astro-Han: Fixed in `a2e46ccaf1`. Introduced a `PawworkSkillCard` type where `homeIconStyle?: JSX.CSSProperties` is an explicit optional field (not a presence-via-key). Dropped the `as const` literal narrowing in favor of the explicit type, so `card.homeIconStyle` is always readable as `{...} | undefined` and the `in` check in the JSX is gone. Option (a) from your list.

packages/app/src/components/session/pawwork-skill-meta.ts
Astro-Han: Keeping the inline hex for the same reason as the Gemini/CodeRabbit threads on this file: Tailwind v4 here configures `--color-*: initial`, which wipes the default palette. `text-violet-500` produces no CSS variable and renders as the inherited foreground (verified during this branch — the icon rendered black until switched to inline style). Added an inline comment documenting why palette utilities don't work here, so future readers don't re-suggest it. When we introduce a brand-color token layer this switches to a named utility.

packages/app/src/components/session/session-new-view.tsx
Astro-Han: Fixed in `a2e46ccaf1`: the pill grid now uses `grid-cols-1 sm:grid-cols-3`, so below the `sm` breakpoint (640px) the pills stack to a single column and don't overflow the viewport. At ≥640px they keep the equal-width three-column layout.

packages/app/src/components/session/session-new-view.tsx
Astro-Han: Same fix in `a2e46ccaf1` — `grid-cols-1 sm:grid-cols-3`, so narrow viewports stack to one column. Verified with Chinese (处理文档 / 分析数据 / 开始写作) and the current English labels (Process docs / Analyze data / Start writing) that the ≥640px grid stays balanced.

packages/app/src/pages/session/session-side-panel.tsx
Astro-Han: Holding off on the convention for now — only `review` and `terminal` have active variants that belong to this component. `status` is rendered out-of-band by `status-popover.tsx` (already active-aware there), and `folder` has no active state in the current UI. Going convention-based would mean stub entries for tabs that don't semantically toggle, which feels like YAGNI at 2 cases. Happy to flip it when a third shell tab grows an active variant.

packages/ui/src/components/icon.tsx
Astro-Han: Fixed in `6a583fbd8f`: `pin` and `chevron-up` now sit on separate lines.

packages/ui/src/components/icon.tsx
Astro-Han: Fixed in `6a583fbd8f`: removed the 5 stray blank lines between `bar-chart` and `pin`.

packages/app/src/i18n/en.ts
Astro-Han: These went through an explicit wording pass. The three pills on the home now carry parallel verb + noun shapes: `Process docs / Analyze data / Start writing` — "Write faster" was verb + adverb which didn't match the other two. Picked `docs` over `documents` to keep the pill tight within the equal-width grid — "docs" reads as "files you can work with" in a product-UI context; the description line ("Edit, convert, and extract from Word, Excel, PowerPoint, and PDF files") disambiguates.

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: The rationale lives in the commit body of `453fe08692` and in the PR description ("the new icon set is drawn at the 16×16 default, so the forced 20×20 was over-compensating for the old icons"). Leaving a negative inline comment ("why this override is not here") tends to rot once the context around it ages out; the commit + PR layer is the durable record.

packages/app/src/pages/layout/pawwork-sidebar.tsx
Astro-Han: Noted — treated as a test comment, resolving.

PR #236, opened 2026-04-25
feat: default LSP off, scope TS-ecosystem to nearest package

### LINKED ISSUE #232: [Bug] TypeScript LSP can consume excessive CPU and memory on large workspaces
### What happened?
When PawWork opens the PawWork repository workspace, the embedded TypeScript language service can consume a full CPU core and several GB of memory. In the local observation on April 25, 2026, the heavy process was not a normal app service, lint task, or Biome process. It was `typescript-language-server` launched by PawWork, with child `tsserver.js` using about 97 percent CPU and around 3.4 GB RSS at peak. After the CPU dropped, the same `tsserver.js` process still kept about 2.6 GB RSS resident.

The process chain was `PawWork` to `typescript-language-server` to `tsserver.js`. The TypeScript server cwd was `/Users/yuhan/workspace/dev/pawwork`, and the command resolved workspace TypeScript from `node_modules/.bun/typescript@5.8.2/node_modules/typescript/lib/tsserver.js`. File evidence showed the server had opened hundreds of files from `packages/opencode`, `packages/app`, `packages/ui`, and `node_modules/.bun`.

### Steps to reproduce
1. Open PawWork v0.2.11.
2. Open a large TypeScript monorepo workspace, using PawWork itself as the observed example.
3. Read or inspect TypeScript source files so PawWork warms LSP state.
4. Check Activity Monitor or run `ps` for `typescript-language-server` and `tsserver.js`.

Opening or reading files in a large workspace should not start a TypeScript language service that analyzes the whole monorepo by default and holds multiple GB of memory. If LSP is available, it should stay bounded to the relevant package, be explicitly opt in for expensive analysis, or have a user visible way to disable TypeScript LSP for the project.

### PawWork version
v0.2.11

### OS version
macOS 26.3.1, build 25D771280a

Only once so far, but the process state was persistent and directly observable.

Observed process evidence: PID 39681 was `node .../typescript-language-server --stdio`, parented by PawWork. PID 39683 was `node .../typescript/lib/tsserver.js`, parented by PID 39681. At peak it used about 97 percent CPU and about 3.4 GB RSS. Later it idled at 0 percent CPU while still retaining about 2.6 GB RSS.

Initial root cause evidence: `packages/opencode/src/lsp/server.ts` enables the builtin TypeScript LSP and chooses the nearest package lock as the root. In this repository that resolves to the repo root because `bun.lock` is at the root. `packages/opencode/src/tool/read.ts` warms LSP after reading ordinary text files. The root `tsconfig.json` has no `include` or `files`, so TypeScript can treat the repository as a large inferred project. This points to TypeScript LSP project scoping and automatic warming, not Biome or lint, as the layer to fix.

Possible follow up directions: make the root `tsconfig.json` a solution config with `files: []` and package references, avoid warming LSP from ordinary reads unless the user or workflow needs language intelligence, or expose a project config such as `lsp.typescript.disabled` as a documented performance escape hatch.

Astro-Han:
## Summary

Make LSP opt-in (default off) with a Settings → General toggle, and prevent the TypeScript LSP from blowing up to monorepo-root scope when active.

- New `Settings.Service` holds `lspEnabled` in an Effect `Ref`; LSP and ToolRegistry layers read it at state-init time so the lsp tool and server registry both flip together.
- Renderer toggle calls `window.api.setLspEnabled` → Electron main IPC handler enters each active `Instance` scope, runs `LSP.shutdownAll` (only on going false), then `LSP.invalidate` and `ToolRegistry.invalidate` so caches and `tsserver` processes are reset per-project.
- New `JavascriptPackageRoot` helper prepends `tsconfig.json` and `package.json` to lockfile root markers, applied to TypeScript / Vue / ESLint / Svelte / Astro / YamlLS.
- `read.ts` no longer triggers `lsp.touchFile` on read (aligns with Codex CLI / Gemini CLI / Claude Code baseline).
- `LSPServer.Info.packages` becomes the single source of truth for npm-installed servers; install failures from `Npm.add` are classified as transient and skip `s.broken` poison so the next edit can retry.
- New `lsp.server.install.failed` bus event surfaces an error toast in the renderer.

## Why

Issue #232: opening PawWork's own monorepo caused `tsserver` to balloon to ~3.4 GB / 97% CPU because the TS server treated the worktree root (lockfile) as the project root. Two failure modes were tangled: (1) LSP cost was unconditional even when users did not need diagnostics, (2) TS root resolution found the lockfile before the nearest `tsconfig.json`. The toggle gives users the off switch; the NearestRoot fix keeps the on path bounded to one package.

## Related Issue

Closes #232

## How To Verify

```bash
bun --cwd packages/opencode run typecheck
bun --cwd packages/opencode test
bun --cwd packages/app run typecheck
bun --cwd packages/app test
bun --cwd packages/desktop-electron run typecheck
bun --cwd packages/desktop-electron test
```

Manual smoke (run on macOS, in `bun run dev:desktop`):

- ✅ #1 LSP off (default): fresh launch → Agent reads `.ts` → `ps aux | grep -iE "language-server|tsserver|typescript"` empty.
- ✅ #2 LSP on, scope contained: toggle on → Agent edits `packages/app/src/app.tsx` → exactly one `typescript-language-server` plus its `tsserver.js` workers, all rooted at `packages/app/`, RSS well under 1 GB.
- ✅ #3 Toggle off mid-session: with #2 running, flip off → all 4 LSP processes gone within ~1 s.
- Skipped (network gating tedious): #4 offline install failure toast, #5 reconnect retry.

## Screenshots or Recordings

Settings → General toggle row added at the bottom of the existing list; copy is `Language Server Protocol (LSP)` / `Detect type errors and symbol references when editing code`. No layout reflow.

## Checklist

## Release Notes

* **New Features**
  * Added a toggleable LSP (Language Server Protocol) setting in General settings to enable/disable language server functionality
  * Added error notifications displaying package name and error details when LSP server installation fails

* **Improvements**
  * LSP feature is now configurable via settings instead of an experimental flag


packages/opencode/src/lsp/index.ts
```diff
@@ -154,52 +158,33 @@ export namespace LSP {
...
-          const cfg = yield* config.get()
+          const lspEnabled = yield* settings.lspEnabled()
 
           const servers: Record<string, LSPServer.Info> = {}
 
-          if (cfg.lsp === false) {
-            log.info("all LSPs are disabled")
+          if (!lspEnabled) {
+            log.info("LSP disabled (Settings.lspEnabled=false)")
           } else {
```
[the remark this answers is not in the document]
Astro-Han: Verified — this is incorrect. `packages/opencode/src/settings/index.ts:15` initializes the Ref with `Ref.make<boolean>(false)`. `Settings.defaultLayer` is the same Layer reference as `Settings.layer`, and the singleton `memoMap` shared by every `makeRuntime` plus `AppRuntime` guarantees one cell across CLI, embedded server, and Electron. CLI/non-Electron code that does not flip the toggle reads `false`, which is the spec'd default. Test `packages/opencode/test/settings.test.ts` asserts this directly.

packages/app/src/context/settings.tsx
```diff
@@ -238,6 +240,13 @@ export const { use: useSettings, provider: SettingsProvider } = createSimpleCont
         setEditToolPartsExpanded(value: boolean) {
           setStore("general", "editToolPartsExpanded", value)
         },
+        lspEnabled: withFallback(() => store.general?.lspEnabled, defaultSettings.general.lspEnabled),
+        setLspEnabled(value: boolean) {
```
Astro-Han, line 255: [P1] The persisted renderer setting is only sent to the main process when the user changes this setter, so an already-enabled user loses LSP after a restart while the switch still renders as on. `Settings.Service` in `packages/opencode/src/settings/index.ts` defaults to false and has no persisted backing, and I only see `window.api.setLspEnabled` called from this setter. On launch, `store.general.lspEnabled` can hydrate to true from `settings.v3`, but the main-process ref remains false until the user toggles again, leaving both the LSP registry and the `lsp` tool disabled behind an enabled UI. Please sync the hydrated value once the persisted settings store is ready, or move this setting into a backend-persisted source of truth.

packages/opencode/src/lsp/server.ts
```diff
@@ -94,10 +105,8 @@ export namespace LSPServer {
...
     id: "typescript",
-    root: NearestRoot(
-      ["package-lock.json", "bun.lockb", "bun.lock", "pnpm-lock.yaml", "yarn.lock"],
-      ["deno.json", "deno.jsonc"],
-    ),
+    packages: ["typescript-language-server"],
+    root: NearestRoot(JavascriptPackageRoot(), ["deno.json", "deno.jsonc"]),
     extensions: [".ts", ".tsx", ".js", ".jsx", ".mjs", ".cjs", ".mts", ".cts"],
     async spawn(root) {
       const tsserver = Module.resolve("typescript/lib/tsserver.js", Instance.directory)
```
Astro-Han: [P2] This still resolves the TypeScript SDK from `Instance.directory`, even though this PR changes the server root to the nearest package. In the target monorepo case, that means `typescript-language-server` can be started with `cwd = packages/app` while being told to use the root workspace's `typescript/lib/tsserver.js`, or fail entirely when only the package has TypeScript installed. That weakens the package-scoping fix because diagnostics can still be driven by the wrong TS SDK. Please resolve `typescript/lib/tsserver.js` from the selected `root` first, with an explicit fallback only if that is intentional.

packages/opencode/src/lsp/index.ts
```diff
@@ -487,6 +477,20 @@ export namespace LSP {
...
       })
 
+      const shutdownAll = Effect.fn("LSP.shutdownAll")(function* () {
+        if (!(yield* InstanceState.has(state))) return
+        const s = yield* InstanceState.get(state)
+        const inFlight = Array.from(s.spawning.values())
+        s.spawning.clear()
+        yield* Effect.promise(() => Promise.allSettled(inFlight))
+        yield* Effect.promise(() => Promise.all(s.clients.map((c) => c.shutdown().catch(() => {}))))
+        s.clients.length = 0
```
Astro-Han, line 507: [P3] After `shutdownAll()` clears `s.clients`, no `LSP.Event.Updated` event is published. The renderer only refreshes cached LSP status on `lsp.updated`, or lazily when `lsp_ready` is false; once the status popover has loaded it sets `lsp_ready=true`. So toggling LSP off can kill the processes while the UI continues to show the previous connected servers until another event happens. Please publish `Event.Updated` after the shutdown clears clients, or explicitly reset the renderer-side LSP status when the setting is turned off.

Astro-Han (commented):
Review pass: three concrete inline findings, graded by P-level.

PR #245, opened 2026-04-26
feat(question): tighten dock cap, schema bounds, and tool description

### LINKED ISSUE #188: [Feature] Improve question tool clarity and proactive use
Use PawWork's structured question tool as a low-friction way for the agent to clarify user intent, present real trade-offs, and collect decisions before continuing work. This matters when the next action depends on user preference, scope, risk tolerance, or implementation direction.

### What do you do today?
The question tool exists, but the experience is still too compressed. Longer questions or trade-off explanations can appear as one dense block because the question text is rendered as plain text without preserving line breaks or formatting. Multi-select is supported in the protocol and UI, but it is easy to miss unless the agent explicitly sets the right field and the prompt makes that mode clear. The agent also does not call the tool proactively enough, so ambiguity often falls back to ordinary chat instead of a structured decision point.

The agent should use the question tool more actively when a user decision would materially affect the next step, while still avoiding trivial interruptions. The UI should make structured questions easier to scan: preserve intentional line breaks at minimum, make single-select versus multi-select obvious, and consider a separate optional body or description area for context, trade-offs, and longer explanations instead of forcing everything into the question title. Preset options should remain concise, with descriptions carrying the details, and custom answers should remain available.

Both

### Extra context
Comparable tools already treat structured questions as a dedicated elicitation surface. Claude Code's AskUserQuestion supports concise options, descriptions, multiSelect, Other, and optional previews for richer comparisons. Kimi Code documents single-select, multi-select, custom text input, multi-question tabs, keyboard navigation, and restoring previous selections. Gemini CLI's ask_user supports choice, text, yes/no, multiSelect, placeholders, and renders question text through a Markdown-capable dialog. For PawWork, a pragmatic first slice is probably: preserve line breaks in question text, improve the visible single-select or multi-select hint, and tune agent instructions so it chooses the tool for meaningful ambiguity and approach selection.

### LINKED ISSUE #216: [Feature] Cap question dock height so footer and history stay in viewport

Astro-Han:
## Summary

Closes #216 and #188 with four reversible commits:

- **C1 `feat(question)`** — schema bounds (`question.max(200)`, `header.max(30)`, `option.label.max(50)`, `option.description.max(50)`, `options.min(2).max(4)`, `questions.min(1).max(4)`), `Info.refine(custom !== false || options.length > 0)`, every bound carries an actionable hint message so the LLM gets a structured repair path on rejection. Tool description rewritten with `When / NOT / Examples / Format / Content routing` sections (inspired by Claude Code's `EnterPlanMode`); `Content routing` teaches the model to stream long framing as normal output before invoking the tool.
- **C2 `fix(ui)`** — replace the dynamic `measure()` (which fell back to `100dvh` when sticky-class detection failed, the v0.2.12 footer-off-screen repro) with a static `max-height: min(70dvh, calc(100dvh - 180px))`. The 180px buffer is the PawWork session top chrome stack: 44 titlebar + 48 tabs + 48 sticky session title + 40 safety margin. No JS measurement, no `ResizeObserver`, no CSS variables.
- **C3a `feat(ui)`** — honor `Question.Info.custom` flag: wrap the type-your-own-answer slot in `<Show when={q.custom !== false}>`, update `count() / pickFocus() / selectOption()`, and clear cache for tabs whose schema disallows custom (avoids stale focus on remount).
- **C3b `feat(ui)`** — inline mode chip in header: `{n} of {total} questions · {Pick 1 | Pick multiple}`. Mode renders weight 600 with `flex-shrink:0` so it survives ellipsis on narrow widths. Drop the dedicated `[data-slot="question-hint"]` row. Tighten `en` + `zh` hint copy: `Pick 1` / `Pick multiple` / `选 1 个` / `可选多个`. Other 16 locales follow upstream sync.

## Why

#216 reported the question dock blocking the footer at small viewport heights (the `measure()` fallback made it render at full viewport when sticky-class detection failed). #188 asked for clearer triggers and tighter formatting so the model invokes the tool more confidently and writes shorter, more clickable questions. Both ship together because they share the dock TSX surface and the schema-versus-prompt discipline; splitting into separate PRs would have meant rebasing the dock changes twice.

## Related Issue

Closes #216
Closes #188

## How To Verify

```bash
bun --cwd packages/opencode test test/question/schema.test.ts   # 16 pass / 0 fail
bun --cwd packages/opencode typecheck                            # 0 errors
bun --cwd packages/app typecheck                                 # 0 errors
bun --cwd packages/app test                                      # 577 pass / 0 fail
```

Manual UI smoke (passed locally on macOS arm64 dev:desktop):

- **Dock cap (C2)** — default window: dock fits ~70dvh, footer visible without scroll. ~480px-tall window: footer still visible (`100dvh - 180 = 300px` budget). 1080px tall: dock caps at ~70dvh ≈ 756px, internal scroll for overflow options.
- **Header inline mode (C3b)** — single-choice question: `2 of 4 questions · Pick 1` (zh: `第 2 题 / 共 4 题 · 选 1 个`). Multi-choice: `... · Pick multiple` / `可选多个`. ~480px-wide window: title ellipsizes, mode chip remains fully visible (`flex-shrink:0`), progress dots stay aligned right. No dedicated hint row below the question text.
- **Custom flag (C3a)** — default request keeps the type-your-own-answer slot. Request with `custom: false` renders no textarea; keyboard navigation does not focus an empty trailing slot.
- **Schema enforcement (C1)** — model with the new schema naturally writes 33–178 char questions on first try (verified via session DB inspection); a deliberately-long 250x payload was rejected with `Question is too long (max 200 chars)...` and the model self-corrected. 5-option request rejected with `Each question allows at most 4 options...`.

Performance check from session export: average reasoning per turn 96 tok across 12 follow-up turns; first-call schema orientation 615 tok one-time, paid into the prompt cache. No simplification warranted.

## Screenshots or Recordings

Verified on macOS dev:desktop. Header structure changes captured in spec at the linked issue comment; dock cap behavior is non-visual (footer-visible regression check).

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Moved progress indicator and mode hint into the question header.

* **Bug Fixes**
  * Custom-answer control is only shown, focusable, and editable when supported; opening is blocked when not allowed.
  * Validation tightened for question text, headers, option labels/descriptions, and question counts (1–4 / 2–4 limits).

* **Documentation**
  * Updated question tool guidance with clearer decision criteria and input rules.

* **UI/Style**
  * Removed dynamic prompt resizing, refined header layout, and updated selection hint text.

* **Tests**
  * Added schema and tool tests covering the new validation rules.

* **Chores**
  * Enforced minimum main window dimensions.

### DIFF
--- a/packages/opencode/src/question/index.ts
+++ b/packages/opencode/src/question/index.ts
@@ -14,31 +14,62 @@ export namespace Question {
 
   export const Option = z
     .object({
-      label: z.string().describe("Display text (1-5 words, concise)"),
-      description: z.string().describe("Explanation of choice"),
+      label: z
+        .string()
+        .trim()
+        .min(1, "Option label cannot be empty.")
+        .max(50, "Option label is too long (max 50 chars). Keep labels to 1–5 words; put detail in description.")
+        .describe("Display text (1–5 words, max 50 chars)"),
+      description: z
+        .string()
+        .trim()
+        .min(1, "Option description cannot be empty.")
+        .max(50, "Option description is too long (max 50 chars). Keep it to one line; longer trade-off context belongs in the question or in normal streamed output before the tool call.")
+        .describe("One-line explanation of choice (max 50 chars)"),
     })
     .meta({ ref: "QuestionOption" })
   export type Option = z.infer<typeof Option>
 
   export const Info = z
     .object({
-      question: z.string().describe("Complete question"),
-      header: z.string().describe("Very short label (max 30 chars)"),
-      options: z.array(Option).describe("Available choices"),
+      question: z
+        .string()
+        .trim()
+        .min(1, "Question cannot be empty.")
+        .max(200, "Question is too long (max 200 chars). Stream longer framing or trade-off context as normal assistant output before invoking this tool, then keep the question short.")
+        .describe("Short question (max 200 chars). Stream longer framing as normal output first."),
+      header: z
+        .string()
+        .trim()
+        .min(1, "Header cannot be empty.")
+        .max(30, "Header is too long (max 30 chars). Use a chip-sized label like 'Auth method' or 'Approach'.")
+        .describe("Very short label (max 30 chars)"),
+      options: z
+        .array(Option)
+        .min(2, "Each question needs at least 2 options.")
+        .max(4, "Each question allows at most 4 options. Keep choices distinct and mutually exclusive.")
+        .describe("Available choices (2–4)"),
       multiple: z.boolean().optional().describe("Allow selecting multiple choices"),
       custom: z.boolean().optional().describe("Allow typing a custom answer (default: true)"),
     })
     .meta({ ref: "QuestionInfo" })
   export type Info = z.infer<typeof Info>
 
-  export const Prompt = Info.omit({ custom: true }).meta({ ref: "QuestionPrompt" })
+  // Prompt is the LLM-facing schema. Identical to Info — we expose `custom` so the
+  // tool description's "Set false only when the options are exhaustive" instruction
+  // is reachable from a real tool call. The omit({}) round-trip preserves the
+  // separate ref so docs render two distinct schemas (one internal, one tool input).
+  export const Prompt = Info.omit({}).meta({ ref: "QuestionPrompt" })
   export type Prompt = z.infer<typeof Prompt>
 
   export const Request = z
     .object({
       id: QuestionID.zod,
       sessionID: SessionID.zod,
-      questions: z.array(Info).describe("Questions to ask"),
+      questions: z
+        .array(Info)
+        .min(1, "Provide at least one question.")
+        .max(4, "Ask at most 4 questions per invocation. If you have more, split into multiple tool calls or stream context first."),
       tool: z
         .object({
           messageID: MessageID.zod,
--- a/packages/opencode/src/tool/question.txt
+++ b/packages/opencode/src/tool/question.txt
@@ -1,10 +1,53 @@
-Use this tool when you need to ask the user questions during execution. This allows you to:
-1. Gather user preferences or requirements
-2. Clarify ambiguous instructions
-3. Get decisions on implementation choices as you work
-4. Offer choices to the user about what direction to take.
+Use this tool proactively when you need a user decision that meaningfully shapes the next step. The user clicks an option (or types a custom answer) and you continue with their answer in mind.
 
-Usage notes:
-- When `custom` is enabled (default), a "Type your own answer" option is added automatically; don't include "Other" or catch-all options
-- Answers are returned as arrays of labels; set `multiple: true` to allow selecting more than one
-- If you recommend a specific option, make that the first option in the list and add "(Recommended)" at the end of the label
+## When to Use
+
+1. **Hard-to-reverse choices**: pushing to dev, force-pushing, deleting branches/files, replacing branding assets, signing/notarizing.
+   - Example: "We're about to push this branch with 3 commits. Squash on merge or merge as-is?"
+2. **Multiple valid technical approaches**: when several solutions are reasonable and user preference matters.
+   - Example: "For the cap, do you want CSS-only with a constant or a dynamic measurement?"
+3. **Ambiguous scope**: when the request could mean different concrete deliverables.
+   - Example: "When you say 'clean up tests', do you mean delete obsolete ones, refactor for readability, or both?"
+4. **Content-style preferences**: tone, length, register, structure for user-facing copy.
+   - Example: "Should the X post lead with the metric or the story?"
+5. **Trade-off where defaults aren't obvious**: when picking the default would silently commit the user to one branch of a real choice.
+   - Example: "Add this gate to all 3 environments now, or only staging this week?"
+
+## When NOT to Use
+
+- Factual lookups you can answer by reading code, docs, or running a command.
+- The user already gave a specific instruction; just execute it.
+- Pure exploration tasks (use the explore agent or codesearch instead).
+- Trivial style choices a sensible default covers.
+- Confirmation theater: don't ask "should I proceed?" after the user already said proceed.
+
+## Examples
+
+GOOD:
+- `question: "Squash on merge or merge as-is?"` with options "Squash" / "Merge as-is".
+- `question: "Use SF Pro or system stack?"` with options "SF Pro" / "system stack".
+
+BAD (stuffs context into the question field; should stream context first then ask short):
+- `question: "We're shipping a feature flag for the new dock. The flag controls whether the new measurement runs only in dev mode initially or also in prod. The trade-off is that dev-only avoids any prod regressions but means we need a follow-up to flip it for the actual rollout. So should we enable it in prod from day 1?"` — 200+ chars; rejected by schema.
+- `question: "Yes or no?"` with no options array — schema requires 2–4 options.
+
+## Format
+
+- `header`: very short label (max 30 chars), shown as a chip.
+- `question`: short prompt (max 200 chars). If you need more context, stream it as normal assistant output BEFORE invoking this tool, then keep the question short.
+- `options`: 2–4 distinct choices; keep them mutually exclusive unless `multiple: true`.
+- `options[].label`: 1–5 words (max 50 chars).
+- `options[].description`: one-line explanation (max 50 chars). Longer per-option detail belongs in the streamed context above.
+- `multiple: true` only when choices aren't mutually exclusive (e.g. feature toggles).
+- `custom`: defaults to true; the dock adds a "Type your own answer" slot. Set false only when the options are exhaustive.
+- `questions`: 1–4 per invocation. If you have more, split.
+
+## Content routing
+
+The tool is for short clickable decisions, not long explanations.
+
+When framing a decision needs paragraphs of context, trade-offs, or background:
+1. Stream the context as normal assistant output FIRST (no tool call yet). The user reads it like any chat message.
+2. THEN invoke `question` with a short prompt that references the context you just streamed.
+
+This keeps the dock compact, every option visible without scroll, and the user's eye landing on the actual choice instead of a wall of text.
--- a/packages/opencode/test/question/schema.test.ts
+++ b/packages/opencode/test/question/schema.test.ts
@@ -1,0 +1,151 @@
+import { test, expect } from "bun:test"
+import { Question } from "../../src/question"
+
+test("Option.label rejects > 50 chars", () => {
+  expect(() => Question.Option.parse({ label: "x".repeat(51), description: "ok" })).toThrow()
+})
+
+test("Option.label accepts exactly 50 chars", () => {
+  expect(() => Question.Option.parse({ label: "x".repeat(50), description: "ok" })).not.toThrow()
+})
+
+test("Option.description rejects > 50 chars", () => {
+  expect(() => Question.Option.parse({ label: "ok", description: "x".repeat(51) })).toThrow()
+})
+
+test("Option.description accepts exactly 50 chars", () => {
+  expect(() => Question.Option.parse({ label: "ok", description: "x".repeat(50) })).not.toThrow()
+})
+
+const validOption = { label: "ok", description: "ok" }
+const validInfo = {
+  question: "ok",
+  header: "ok",
+  options: [validOption, validOption],
+  multiple: false,
+  custom: true,
+}
+
+test("Info.question rejects > 200 chars", () => {
+  expect(() => Question.Info.parse({ ...validInfo, question: "x".repeat(201) })).toThrow()
+})
+
+test("Info.question accepts exactly 200 chars", () => {
+  expect(() => Question.Info.parse({ ...validInfo, question: "x".repeat(200) })).not.toThrow()
+})
+
+test("Info.header rejects > 30 chars", () => {
+  expect(() => Question.Info.parse({ ...validInfo, header: "x".repeat(31) })).toThrow()
+})
+
+test("Info.header accepts exactly 30 chars", () => {
+  expect(() => Question.Info.parse({ ...validInfo, header: "x".repeat(30) })).not.toThrow()
+})
+
+test("Option.label rejects empty / whitespace-only", () => {
+  expect(() => Question.Option.parse({ label: "", description: "ok" })).toThrow()
+  expect(() => Question.Option.parse({ label: "   ", description: "ok" })).toThrow()
+})
+
+test("Info.question rejects empty / whitespace-only", () => {
+  expect(() => Question.Info.parse({ ...validInfo, question: "" })).toThrow()
+  expect(() => Question.Info.parse({ ...validInfo, question: "   " })).toThrow()
+})
+
+test("Info.options rejects fewer than 2", () => {
+  expect(() => Question.Info.parse({ ...validInfo, options: [validOption] })).toThrow()
+})
+
+test("Info.options rejects more than 4", () => {
+  expect(() => Question.Info.parse({ ...validInfo, options: Array(5).fill(validOption) })).toThrow()
+})
+
+test("Info.options accepts exactly 2", () => {
+  expect(() => Question.Info.parse({ ...validInfo, options: [validOption, validOption] })).not.toThrow()
+})
+
+test("Info.options accepts exactly 4", () => {
+  expect(() => Question.Info.parse({ ...validInfo, options: Array(4).fill(validOption) })).not.toThrow()
+})
+
+test("Info rejects custom: false with empty options (caught by options.min(2))", () => {
+  // options.min(2) catches empty arrays before any custom-vs-options refinement could fire,
+  // so the dock is guaranteed to receive at least 2 selectable options regardless of custom.
+  expect(() => Question.Info.parse({ ...validInfo, custom: false, options: [] })).toThrow()
+})
+
+test("Prompt accepts custom flag (LLM can set it for exhaustive options)", () => {
+  // Prompt = Info.omit({}) — `custom` is exposed so the tool description's
+  // "Set false only when the options are exhaustive" instruction is reachable.
+  const withCustom = {
+    question: "ok",
+    header: "ok",
+    options: [validOption, validOption],
+    multiple: false,
+    custom: false,
+  }
+  expect(() => Question.Prompt.parse(withCustom)).not.toThrow()
+  // And without custom (defaults to true at the dock layer).
+  const withoutCustom = {
+    question: "ok",
+    header: "ok",
+    options: [validOption, validOption],
+    multiple: false,
+  }
+  expect(() => Question.Prompt.parse(withoutCustom)).not.toThrow()
+})
+
+test("Info.question rejection message includes the actionable hint", () => {
+  // Each .max() carries a hint so the LLM can repair on rejection. Guard a representative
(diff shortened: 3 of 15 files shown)


packages/opencode/src/question/index.ts
```diff
@@ -14,31 +14,59 @@ export namespace Question {
...
 
   export const Request = z
     .object({
       id: QuestionID.zod,
       sessionID: SessionID.zod,
-      questions: z.array(Info).describe("Questions to ask"),
+      questions: z
+        .array(Info)
+        .min(1, "Provide at least one question.")
+        .max(4, "Ask at most 4 questions per invocation. If you have more, split into multiple tool calls or stream context first."),
```
Astro-Han, line 72: P1: This limit is not on the model-facing tool schema. I checked `packages/opencode/src/tool/question.ts`; the actual parameter is still `questions: z.array(Question.Prompt)`, and `Question.Service.ask` copies `params.questions` into the request without parsing `Question.Request`. So a tool call with 5 questions still bypasses this bound, even though `question.txt` says 1 to 4. Please put `.min(1).max(4)` on the tool parameter array, or derive the tool schema from a prompt request schema so the main path is enforced.

packages/opencode/src/question/index.ts
```diff
@@ -14,31 +14,59 @@ export namespace Question {
 
   export const Option = z
     .object({
-      label: z.string().describe("Display text (1-5 words, concise)"),
-      description: z.string().describe("Explanation of choice"),
+      label: z
+        .string()
+        .max(50, "Option label is too long (max 50 chars). Keep labels to 1–5 words; put detail in description.")
```
Astro-Han, line 21: P2: This changes the exported question schema, but the tracked SDK/OpenAPI artifacts were not regenerated. I checked `packages/sdk/openapi.json` and `packages/sdk/js/src/v2/gen/types.gen.ts` at this head; they still show the old QuestionOption/QuestionInfo descriptions and no min/max metadata. External SDK users and API docs will not see the new bounds until those generated files are updated or the PR states that generation is deferred.

packages/opencode/test/question/schema.test.ts
```diff
@@ -0,0 +1,107 @@
+import { test, expect } from "bun:test"
+import { Question } from "../../src/question"
+
+test("Option.label rejects > 50 chars", () => {
+  expect(() => Question.Option.parse({ label: "x".repeat(51), description: "ok" })).toThrow()
```
Astro-Han, line 5: P3: The PR's contract is not just reject bad input; it also promises actionable repair messages for the model. These tests only assert `toThrow()`, so replacing the custom messages with a generic Zod error would still pass. Please assert at least one representative length error and one count error include the intended repair hint.

packages/ui/src/components/message-part.css
```diff
@@ -832,7 +832,9 @@
   flex-direction: column;
   gap: 0;
   min-height: 0;
-  max-height: var(--question-prompt-max-height, 100dvh);
+  /* 180px = 44 (titlebar) + 48 (tabs) + 48 (sticky session title) + 40 (safety margin).
+     If session top chrome changes, retune this constant. */
+  max-height: min(70dvh, calc(100dvh - 180px));
```
Astro-Han, line 837: P3: This removes the previous `Math.max(240, ...)` floor, and there is no Electron `minHeight` on the main window. If a user shrinks the window below the 480px smoke case, the whole dock budget becomes `100dvh - 180px`, for example 140px at a 320px window, including header and footer. Please either add a lower bound/min window height, or document and smoke-test the smallest supported height.

Astro-Han (commented):
Found one P1 enforcement gap plus a few smaller contract and UI edge cases. Details are inline.

packages/opencode/src/question/index.ts
Astro-Han: Fixed in 3d451fdd. Moved `.min(1).max(4)` onto the actual entry-point schema (`packages/opencode/src/tool/question.ts` `parameters`), since `Question.Service.ask` takes `Info[]` directly and never `parse(Request)`. The `Request` bound stays for the schema-test invariant guard, but the runtime gate is now on `parameters`. Verified locally with `bun --cwd packages/opencode test test/tool/question.test.ts`.

packages/opencode/src/question/index.ts
Astro-Han: Regenerated in 80ac51d0. Ran `packages/sdk/js/script/build.ts` so `packages/sdk/openapi.json` and `packages/sdk/js/src/v2/gen/{types,sdk}.gen.ts` now reflect the new `maxLength` / `minItems` / `maxItems` on `QuestionOption` / `QuestionInfo` / `QuestionPrompt`. Diff also picks up a few unrelated pre-existing drifts (session.export, session sort=created, copy edits) that had accumulated since the last regen — flagged in the commit body.

packages/opencode/test/question/schema.test.ts
Astro-Han: Fixed in 3d451fdd. Added two hint-content assertions in `schema.test.ts`:

- `Info.question rejection message includes the actionable hint` — asserts `"Question is too long"` + `"max 200 chars"` for a 201-char question.
- `Info.options rejection message includes the actionable hint` — asserts `"at most 4 options"` for a 5-option payload.

Picked one length-error and one count-error so future refactors can't silently drop the hint contract; not asserting all 8 hints individually since these two cover both bound shapes (`.max(N, hint)` on string vs `.max(N, hint)` on array).

packages/ui/src/components/message-part.css
Astro-Han: Fixed in 1a20b5c7 (different commit since it touches `packages/desktop-electron`, not the dock CSS). Set `minWidth: 480` and `minHeight: 480` on the main `BrowserWindow` in `packages/desktop-electron/src/main/windows.ts:69`. 480 matches the smallest height the C2 smoke checklist asserts the footer remains visible at, so any in-bounds window guarantees the `100dvh - 180px` budget stays viable.

packages/opencode/src/question/index.ts
```diff
@@ -14,31 +14,59 @@ export namespace Question {
...
+        path: ["custom"],
+      },
+    )
     .meta({ ref: "QuestionInfo" })
   export type Info = z.infer<typeof Info>
 
+  // Prompt is the LLM-facing schema; `custom` is omitted and the refine is intentionally dropped
+  // (Zod 4 strips refinements on omit, which is the desired behavior here)
   export const Prompt = Info.omit({ custom: true }).meta({ ref: "QuestionPrompt" })
   export type Prompt = z.infer<typeof Prompt>
```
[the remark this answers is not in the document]
Astro-Han: Disagree, pushing back. Empirically tested in this repo's installed Zod (4.1.8): `Info.omit({ custom: true })` after `.refine()` works fine and returns a plain `ZodObject`. The runtime behavior:

```
$ bun -e "
import { Question } from './src/question'
console.log('Prompt is', Question.Prompt.constructor.name)
try { Question.Prompt.parse({question:'x'.repeat(201), header:'h', options:[{label:'a',description:'a'},{label:'b',description:'b'}], multiple:false}); console.log('ACCEPTED — bug') }
catch(e) { console.log('REJECTED ✓') }
"
Prompt is ZodObject
REJECTED ✓
```

The `unit-opencode` job's previous failure was an unrelated test-fixture issue (an existing test with only 1 option tripped the new `options.min(2)` and timed out 30s waiting for a question that never appeared) — fixed in 3d451fdd. Ran 16/16 schema tests + 2/2 tool tests locally and all pass against the current code.

The `InfoBase` extraction would also lose the original Zod-4 idiom, where `.refine()` followed by `.omit()` is the documented pattern for "wrap base with predicate, then derive a stripped-down schema for downstream consumers". And as Gemini's medium-priority comment notes, the refine itself was redundant (`options.min(2)` catches the same case earlier) — so it's now removed in 3d451fdd. Closing this thread as not-an-issue.

packages/opencode/src/question/index.ts
```diff
@@ -14,31 +14,59 @@ export namespace Question {
...
       multiple: z.boolean().optional().describe("Allow selecting multiple choices"),
       custom: z.boolean().optional().describe("Allow typing a custom answer (default: true)"),
     })
+    .refine(
+      (d) => d.custom !== false || d.options.length > 0,
+      {
+        message: "When custom is false, you must provide at least one option (otherwise the user has no way to answer).",
+        path: ["custom"],
+      },
+    )
```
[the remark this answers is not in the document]
Astro-Han: Agreed. Dropped the `Info.refine` in 3d451fdd since `options.min(2)` catches the `custom: false + empty options` case earlier with a clearer message. The renamed test (`Info rejects custom: false with empty options (caught by options.min(2))`) documents the now-explicit hand-off.

packages/opencode/test/question/schema.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  expect(() => Question.Info.parse({ ...validInfo, question: "x".repeat(201) })).toThrow()
+})
+
+test("Info.question accepts exactly 200 chars", () => {
+  expect(() => Question.Info.parse({ ...validInfo, question: "x".repeat(200) })).not.toThrow()
+})
+
+test("Info.header rejects > 30 chars", () => {
+  expect(() => Question.Info.parse({ ...validInfo, header: "x".repeat(31) })).toThrow()
+})
```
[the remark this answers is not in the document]
Astro-Han: Added in 3d451fdd. New test `Info.header accepts exactly 30 chars` placed right after the rejects-31 case for parity with the other three boundary pairs.

packages/opencode/src/tool/question.txt
```diff
@@ -1,10 +1,53 @@
...
+
+## Format
+
+- `header`: very short label (max 30 chars), shown as a chip.
+- `question`: short prompt (max 200 chars). If you need more context, stream it as normal assistant output BEFORE invoking this tool, then keep the question short.
+- `options`: 2–4 distinct, mutually exclusive choices.
+- `options[].label`: 1–5 words (max 50 chars).
+- `options[].description`: one-line explanation (max 50 chars). Longer per-option detail belongs in the streamed context above.
+- `multiple: true` only when choices aren't mutually exclusive (e.g. feature toggles).
+- `custom`: defaults to true; the dock adds a "Type your own answer" slot. Set false only when the options are exhaustive.
```
Astro-Han, line 42: P2: The docs now tell the model to set `custom: false`, but the model-facing schema still uses `Question.Prompt = Info.omit({ custom: true })`. `z.toJSONSchema(item.parameters)` is what gets sent to providers, so `custom` is not advertised as a valid tool argument. That makes the new `custom=false` UI path unreachable from normal `question` tool calls unless we rely on unknown property leakage. Please either include `custom` in `Prompt` or remove this instruction and keep `custom=false` internal only.

packages/opencode/src/question/index.ts
```diff
@@ -14,17 +14,33 @@ export namespace Question {
...
     .object({
-      label: z.string().describe("Display text (1-5 words, concise)"),
-      description: z.string().describe("Explanation of choice"),
+      label: z
+        .string()
+        .max(50, "Option label is too long (max 50 chars). Keep labels to 1–5 words; put detail in description.")
+        .describe("Display text (1–5 words, max 50 chars)"),
+      description: z
+        .string()
+        .max(50, "Option description is too long (max 50 chars). Keep it to one line; longer trade-off context belongs in the question or in normal streamed output before the tool call.")
```
Astro-Han, line 27: P2: This `maxLength` is now part of the exported `QuestionOption` schema, but `Question.Service.ask` does not parse `Info` or `Request` before publishing `question.asked`. Existing internal `PlanExitTool` already passes `"Switch to build agent and start implementing the plan"` as an option description, length 53, so the server can emit an event that violates the regenerated OpenAPI `maxLength: 50` contract. Please validate in `ask` or trim internal callers to the same schema before publishing.

packages/sdk/js/src/v2/gen/sdk.gen.ts
```diff
@@ -756,12 +758,12 @@ export class Session extends HeyApiClient {
...
    */
   public list<ThrowOnError extends boolean = false>(
-    parameters?: {
+    parameters: {
       directory?: string
       workspace?: string
       roots?: boolean
       start?: number
-      cursor?: number | string
+      cursor: number | string
```
Astro-Han: P2: Regeneration made experimental session `cursor` a required SDK field. The source route in `packages/opencode/src/server/instance/experimental.ts` still declares `cursor` as optional and defaults `limit` to 100 for the first page, so `client.experimental.session.list({})` and `list()` should remain valid. As generated, TypeScript consumers cannot call the first page without inventing a cursor. Please fix the source OpenAPI shape around the preprocess optional cursor, or avoid committing this generated regression.

Astro-Han (commented):
Second pass on the fixed head. I found three remaining P2 contract regressions worth addressing before merge.

packages/opencode/src/tool/question.txt
Astro-Han: Fixed in 10b48ebd. Dropped `Info.omit({ custom: true })` and replaced with `Info.omit({})` re-meta'd, so `custom` is now exposed in `QuestionPrompt`'s JSON schema sent to providers. The tool description's "Set false only when options are exhaustive" instruction is reachable now. Empirically:

```
$ bun -e 'import z from "zod"; import { Question } from "./src/question"; console.log(Object.keys(z.toJSONSchema(Question.Prompt).properties))'
[ "question", "header", "options", "multiple", "custom" ]
```

Updated the schema test to assert Prompt parses both `{custom: false}` and the no-custom default.

packages/opencode/src/question/index.ts
Astro-Han: Fixed in 3be1d108. Trimmed `PlanExitTool` option description from "Switch to build agent and start implementing the plan" (53 chars) to "Switch to build agent and implement the plan" (44 chars). Going with the trim-the-caller path rather than adding `parse` in `Question.Service.ask` since the latter adds runtime cost on every internal ask and only one caller currently violates; if more emerge we can promote to schema validation in ask. The "No" option's description is already at exactly 50, no change needed.

packages/sdk/js/src/v2/gen/sdk.gen.ts
Astro-Han: Fixed in a714b7c0. Root cause: `z.preprocess(transform, inner.optional())` does not propagate the inner-schema optionality to the JSON schema output — `z.toJSONSchema` only sees the preprocess wrapper. Adding an outer `.optional()` on the preprocess solves it:

```ts
cursor: z
  .preprocess(
    (value) => (value === "" ? undefined : value),
    z.union([z.coerce.number(), z.string()]).optional(),
  )
  .optional()  // ← new — makes JSON schema honor the optional contract
  .meta({ description: "Cursor for loading the next page" }),
```

Kept the inner `.optional()` so the `"" → undefined` preprocess still parses (verified by `experimental route ignores empty updated cursor` test in global-session-list.test.ts). Regenerated SDK; `client.experimental.session.list({})` and `list()` are valid again (`cursor?: number | string` in sdk.gen.ts).

packages/opencode/src/question/index.ts
```diff
@@ -14,31 +14,54 @@ export namespace Question {
...
+        .describe("One-line explanation of choice (max 50 chars)"),
```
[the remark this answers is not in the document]
Astro-Han: Pushing back as out-of-scope. Three pieces of evidence:

1. **Upstream contract is required**, not optional. `git show dev:packages/opencode/src/question/index.ts` shows `description: z.string()` (no `.optional()`) since the original opencode import. This PR scope is "tighten bounds" (add `.max(50)`), not "loosen field requiredness". Changing from required to optional is a *separate* contract change.

2. **The "optional descriptions goal" is inferred, not stated**. The dock's `<Show when={props.description}>` (`session-question-dock.tsx:51`) is defensive rendering, not a documented design intent — same pattern as many other components guarding for `undefined`. There's no spec or comment saying "description is intentionally optional".

3. **No caller demonstrates need**. `PlanExitTool` always provides description; tool description in `question.txt` says `options[].description: one-line explanation (max 50 chars)` (treats it as expected). For "Yes / No" style choices, the model can pass a short description; the schema's `.max(50)` is the right pressure, not optionality.

If we want to allow label-only options later, that's a new follow-up: schema + tool description + dock copy + tests, plus deciding whether the model is encouraged to omit description vs. forced to provide one. Not folding that into this PR.

Closing as not-an-issue.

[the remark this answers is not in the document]
Astro-Han: Acknowledged — closing.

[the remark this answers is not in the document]
Astro-Han: Fixed in 3db703ca (autosquashed into C1). Added `.trim().min(1, "...")` to `Option.label`, `Option.description`, `Info.question`, and `Info.header`. The `.trim()` strips leading/trailing whitespace as a parse-time mutation; `.min(1)` then catches both `""` and `"   "` with an actionable hint per field. Added 2 representative tests in `schema.test.ts` (`Option.label rejects empty / whitespace-only`, `Info.question rejects empty / whitespace-only`) so the contract has a regression guard. 37 schema tests pass.

packages/opencode/src/tool/question.txt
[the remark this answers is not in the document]
Astro-Han: Fixed in 3db703ca (autosquashed into C1). Reworded line 38 to "options: 2–4 distinct choices; keep them mutually exclusive unless multiple: true." — removes the apparent conflict with line 41's "multiple: true allows non-exclusive" guidance.

PR #251, opened 2026-04-27
fix: prevent session flicker on prompt send

### LINKED ISSUE #248: [Bug] Page flickers when sending a prompt
## What happened?

Every time I send a prompt in PawWork, the page flickers briefly. The flicker appears to be caused by the composer region's ready state being toggled off and back on with a 140ms delay.

## Steps to reproduce

1. Open PawWork
2. Navigate to a session
3. Type a prompt and press Enter / click Send
4. Observe the page flicker

The page should remain visually stable when sending a prompt, without any flicker or flash.

## PawWork version

Current dev branch (latest)

## OS version

macOS (observed on development machine)

Yes, every time

### Root cause analysis

After code review, the flicker appears to come from `SessionComposerRegion`'s ready state delay mechanism in `packages/app/src/pages/session/composer/session-composer-region.tsx` (lines 101-117):

```typescript
createEffect(() => {
  route.sessionKey()
  const ready = props.ready
  const delay = 140

  clear()
  setStore("ready", false)  // Immediately sets ready = false
  if (!ready) return

  frame = requestAnimationFrame(() => {
    timer = window.setTimeout(() => {
      setStore("ready", true)  // 140ms later, sets back to true
    }, delay)
  })
})
```

The problem:
- This effect runs when `sessionKey` or `props.ready` changes
- It **immediately sets `store.ready = false`** every time
- Then waits 140ms + RAF before restoring to `true`

This creates a brief "unready" state that likely causes the visual flicker.

### Why might sending a prompt trigger this?

Although `sessionKey` and `messagesReady()` should remain stable when sending a message, other factors could cause `props.ready` to briefly change. The `props.ready` value is composed of:

```typescript
ready={!store.deferRender && messagesReady()}
```

(from `packages/app/src/pages/session.tsx` line 1971)

Possible triggers:
1. Any re-computation of `sessionKey` (even if value stays the same) could trigger `deferRender` logic
2. SolidJS's effect tracking may capture transient state changes during async operations

### Suggested investigation

1. In browser DevTools, observe whether `props.ready`, `store.deferRender`, or `messagesReady()` change when sending a prompt
2. Add logging to the effect in session-composer-region.tsx to track when it fires
3. Check if `deferRender` is accidentally triggered during message submission

### Potential fix directions

If confirmed, possible fixes:
- Wrap the effect with `on()` to only run when values actually change (not just re-compute)
- Only set `store.ready = false` when `props.ready` transitions from true to false
- Evaluate whether the 140ms delay is necessary (it may be intended for smooth session switching transitions)

Astro-Han:
## Summary

- Keep session artifact history reads on the latest available value during refetch so prompt sends do not bubble to the root Suspense fallback.
- Add an E2E regression test that samples the session flow during prompt send and verifies the message list and composer dock stay mounted.

## Why

Sending a prompt could trigger artifact history refetch in the session page. While that resource was pending, the app-level Suspense fallback briefly replaced the route content with an empty main area, causing the whole session flow page to flicker.

## Related Issue

Fixes #248

## How To Verify

```bash
bun --cwd packages/app playwright test e2e/session/session-send-stability.spec.ts --reporter=line
bun --cwd packages/app playwright test e2e/session/session-send-stability.spec.ts --reporter=line --repeat-each=3
bun --cwd packages/app playwright test e2e/app/session.spec.ts e2e/session/session-send-stability.spec.ts --reporter=line
bun run --cwd packages/app typecheck
```

## Screenshots or Recordings

Manual Electron dev app check completed locally. The reported session-page flicker was no longer visible during prompt send.

## Checklist

- [x] I am targeting dev, and my PR title and commit messages use Conventional Commits in English

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end session stability test that monitors UI stability while sending a prompt, ensuring composer, turn list, and message elements remain mounted and that samples remain consistent.

* **Refactor**
  * Improved artifact history handling to avoid mixing artifacts across sessions and to ensure artifact-derived views reflect the current session.

### DIFF
--- a/packages/app/e2e/session/session-send-stability.spec.ts
+++ b/packages/app/e2e/session/session-send-stability.spec.ts
@@ -1,0 +1,161 @@
+import type { Page } from "@playwright/test"
+import { test, expect } from "../fixtures"
+import { assistantText } from "../actions"
+import {
+  promptSelector,
+  sessionComposerDockSelector,
+  sessionMessageItemSelector,
+  sessionTurnListSelector,
+} from "../selectors"
+
+const MAX_STABILITY_SAMPLES = 256
+
+type StabilitySample = {
+  at: number
+  url: string
+  composerDock: number
+  messageList: number
+  messages: number
+}
+type StabilityProbeResult = {
+  samples: StabilitySample[]
+  unmounts: StabilitySample[]
+  unmountCount: number
+}
+
+async function installSessionStabilityProbe(page: Page) {
+  await page.evaluate(
+    ({ composerDockSelector, maxSamples, messageItemSelector, turnListSelector }) => {
+      const read = () => ({
+        at: performance.now(),
+        url: window.location.href,
+        composerDock: document.querySelectorAll(composerDockSelector).length,
+        messageList: document.querySelectorAll(turnListSelector).length,
+        messages: document.querySelectorAll(messageItemSelector).length,
+      })
+      const changed = (a: ReturnType<typeof read>, b: ReturnType<typeof read>) =>
+        a.url !== b.url ||
+        a.composerDock !== b.composerDock ||
+        a.messageList !== b.messageList ||
+        a.messages !== b.messages
+      const samples = [read()]
+      const unmounts: ReturnType<typeof read>[] = []
+      let unmountCount = 0
+      const pushBounded = (list: ReturnType<typeof read>[], sample: ReturnType<typeof read>) => {
+        if (list.length < maxSamples) list.push(sample)
+      }
+      const record = () => {
+        const next = read()
+        const prev = samples[samples.length - 1]
+        if (!prev || changed(prev, next)) pushBounded(samples, next)
+      }
+      const removedMatches = (node: Node, selector: string) =>
+        node instanceof Element && (node.matches(selector) || !!node.querySelector(selector))
+      const recordRemovedMount = (input: { composerDock: boolean; messageList: boolean }) => {
+        unmountCount += 1
+        const current = read()
+        const sample = {
+          ...current,
+          composerDock: input.composerDock ? 0 : current.composerDock,
+          messageList: input.messageList ? 0 : current.messageList,
+        }
+        pushBounded(samples, sample)
+        pushBounded(unmounts, sample)
+      }
+      const recordMutations = (records: MutationRecord[]) => {
+        for (const record of records) {
+          for (const node of record.removedNodes) {
+            const composerDockRemoved = removedMatches(node, composerDockSelector)
+            const messageListRemoved = removedMatches(node, turnListSelector)
+            if (composerDockRemoved || messageListRemoved) {
+              recordRemovedMount({ composerDock: composerDockRemoved, messageList: messageListRemoved })
+            }
+          }
+        }
+        record()
+      }
+      let frame = requestAnimationFrame(function tick() {
+        record()
+        frame = requestAnimationFrame(tick)
+      })
+      const observer = new MutationObserver(recordMutations)
+      observer.observe(document.body, { childList: true, subtree: true })
+      const win = window as typeof window & {
+        __opencode_e2e?: Record<string, unknown> & {
+          sessionMountProbe?: { stop: () => unknown }
+        }
+      }
+      win.__opencode_e2e = {
+        ...win.__opencode_e2e,
+        sessionMountProbe: {
+          stop() {
+            cancelAnimationFrame(frame)
+            observer.disconnect()
+            record()
+            delete win.__opencode_e2e?.sessionMountProbe
+            return { samples, unmounts, unmountCount }
+          },
+        },
+      }
+    },
+    {
+      composerDockSelector: sessionComposerDockSelector,
+      maxSamples: MAX_STABILITY_SAMPLES,
+      messageItemSelector: sessionMessageItemSelector,
+      turnListSelector: sessionTurnListSelector,
+    },
+  )
+}
+
+async function stopSessionStabilityProbe(page: Page) {
+  const result = await page.evaluate(() => {
+    const win = window as typeof window & {
+      __opencode_e2e?: {
+        sessionMountProbe?: { stop: () => unknown }
+      }
+    }
+    const probe = win.__opencode_e2e?.sessionMountProbe
+    if (!probe) throw new Error("session stability probe was not installed")
+    return probe.stop()
+  })
+  return result as StabilityProbeResult
+}
+
+test("keeps the session flow mounted while sending a prompt", async ({ page, project, assistant }) => {
+  test.setTimeout(120_000)
+
+  await project.open()
+  await assistant.reply("seed reply")
+  const sessionID = await project.prompt("start a session")
+
+  await expect(page.locator(sessionTurnListSelector)).toHaveCount(1)
+  await expect(page.locator(sessionComposerDockSelector)).toHaveCount(1)
+
+  const beforeCalls = await assistant.calls()
+  const token = `I248_OK_${Date.now()}`
+  await assistant.reply(token)
+
+  await installSessionStabilityProbe(page)
+  let result: StabilityProbeResult = { samples: [], unmounts: [], unmountCount: 0 }
+  try {
+    const prompt = page.locator(promptSelector).first()
+    await expect(prompt).toBeVisible()
+    await prompt.click()
+    await page.keyboard.type(`reply with ${token}`)
+    await expect
+      .poll(async () => (await prompt.textContent())?.replace(/\u200B/g, "").trim())
+      .toBe(`reply with ${token}`)
+    await page.keyboard.press("Enter")
+
+    await expect.poll(() => assistant.calls(), { timeout: 30_000 }).toBeGreaterThan(beforeCalls)
+    await expect.poll(() => assistantText(project.sdk, sessionID), { timeout: 30_000 }).toContain(token)
+  } finally {
+    result = await stopSessionStabilityProbe(page)
+  }
+
+  const unmounted = result.samples.filter((sample) => sample.composerDock === 0 || sample.messageList === 0)
+  expect(unmounted).toEqual([])
+  expect(result.unmounts).toEqual([])
+  expect(result.unmountCount).toBe(0)
+  expect(new Set(result.samples.map((sample) => sample.url)).size).toBe(1)
+})
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -58,7 +58,7 @@ import { useSessionLayout } from "@/pages/session/session-layout"
 import { syncSessionModel } from "@/pages/session/session-model-helpers"
 import { isSessionRunning } from "@/pages/session/session-running-state"
 import { SessionSidePanel } from "@/pages/session/session-side-panel"
-import { deriveArtifactFiles, nextFilesPanelAutoOpen } from "@/pages/session/files-tab-state"
+import { deriveArtifactFiles, nextFilesPanelAutoOpen, type SessionArtifactFile } from "@/pages/session/files-tab-state"
 import { TerminalPanel } from "@/pages/session/terminal-panel"
 import { useSessionCommands } from "@/pages/session/use-session-commands"
 import { useSessionHashScroll } from "@/pages/session/use-session-hash-scroll"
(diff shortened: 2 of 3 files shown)


packages/app/src/pages/session.tsx
```diff
@@ -695,9 +695,10 @@ export default function Page() {
   const [artifactHistory, { refetch: refetchArtifactHistory }] = createResource(
     () => params.id,
     (sessionID) => sdk.client.session.artifacts({ sessionID }).then((res) => res.data ?? []).catch(() => []),
+    { initialValue: [] },
   )
   const artifactFiles = createMemo(() => {
-    const history = artifactHistory() ?? []
+    const history = artifactHistory.latest ?? artifactHistory() ?? []
```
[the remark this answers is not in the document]
Astro-Han: Handled with a keyed snapshot instead of the direct suggestion. I verified that replacing `.latest` with `artifactHistory()` reintroduces the session flow unmount in the regression test. The resource now stores `{ sessionID, artifacts }` and `artifactFiles` only uses `.latest` when the snapshot belongs to the current `params.id`, so it avoids Suspense while also avoiding stale artifacts across sessions.

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+type StabilitySample = {
+  at: number
+  url: string
+  composerDock: number
+  messageList: number
+  messages: number
+}
+
+async function installSessionStabilityProbe(page: any) {
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `installSessionStabilityProbe` now accepts Playwright `Page` instead of `any`.

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+            return samples
+          },
+        },
+      })
+    },
+    { composerDockSelector: sessionComposerDockSelector, messageListSelector },
+  )
+}
+
+async function stopSessionStabilityProbe(page: any) {
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `stopSessionStabilityProbe` now accepts Playwright `Page` instead of `any`.

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,98 @@
+import { test, expect } from "../fixtures"
+import { assistantText } from "../actions"
+import { promptSelector, sessionComposerDockSelector } from "../selectors"
+
+const messageListSelector = '[data-slot="session-turn-list"]'
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Renamed the top-level selector constant to `MESSAGE_LIST_SELECTOR` and updated its usages.

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        a.url !== b.url ||
+        a.composerDock !== b.composerDock ||
+        a.messageList !== b.messageList ||
+        a.messages !== b.messages
+      const samples = [read()]
+      const record = () => {
+        const next = read()
+        const prev = samples[samples.length - 1]
+        if (!prev || changed(prev, next)) samples.push(next)
+      }
```
Astro-Han, line 76: **P2** — rAF-only sampling 抓不到 sub-frame 的 unmount/remount。

这条测试的初衷就是抓 Suspense fallback 短暂顶替路由内容，但 `requestAnimationFrame` 一帧只跑一次，如果 fallback 在两次 paint 之间 mount + unmount（同帧 reactive flush 完全有可能），probe 看不到。

建议在 rAF 之外再挂一个 `MutationObserver({ childList: true, subtree: true })` 监听 `document.body`，回调里也调用 `record()`。这样任何 DOM 突变都会被采样，rAF 只作为兜底节奏。否则这个 regression test 的强度其实弱于 issue #248 的描述。

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  await installSessionStabilityProbe(page)
+
+  const prompt = page.locator(promptSelector).first()
+  await expect(prompt).toBeVisible()
+  await prompt.click()
+  await page.keyboard.type(`reply with ${token}`)
+  await expect.poll(async () => (await prompt.textContent())?.replace(/\u200B/g, "").trim()).toBe(`reply with ${token}`)
+  await page.keyboard.press("Enter")
+
```
Astro-Han: **P2** — `installSessionStabilityProbe` 之后任意一条断言抛错（比如 `assistant.calls()` 那个 30s poll 超时），rAF 循环不会停，会一直跑到 page 被销毁。

建议把 install 之后的所有 await 包进 try / finally：

```ts
await installSessionStabilityProbe(page)
let samples: StabilitySample[]
try {
  // ...所有 prompt / poll 断言
} finally {
  samples = await stopSessionStabilityProbe(page)
}
```

好处：失败重跑时 page 不会带着脏的 `__i248SessionStability` 进下一个 test，并且失败信息里也能拿到 partial samples 帮助 debug。

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      const record = () => {
+        const next = read()
+        const prev = samples[samples.length - 1]
+        if (!prev || changed(prev, next)) samples.push(next)
+      }
+      let frame = requestAnimationFrame(function tick() {
+        record()
+        frame = requestAnimationFrame(tick)
+      })
+      Object.assign(window, {
```
Astro-Han: **P3** — `__i248SessionStability` 这个 window 全局名直接引用了 issue 编号。issue 关掉之后，代码里 `i248` 就成了一个没有上下文的死引用，未来 grep 也不知道它在指什么。

建议改成行为命名，比如 `__sessionMountProbe`，或者干脆挂到现有的 `__opencode_e2e` 命名空间下（`__opencode_e2e.sessionMountProbe`），跟仓库里其它 e2e probe 风格一致。issue 链接放在 PR 描述 / test 标题就够了。

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,99 @@
+import type { Page } from "@playwright/test"
+import { test, expect } from "../fixtures"
+import { assistantText } from "../actions"
+import { promptSelector, sessionComposerDockSelector } from "../selectors"
+
```
Astro-Han, line 10: **P3** — `'[data-slot="session-turn-list"]'` 没进 `e2e/selectors.ts`，line 12 的 `[data-message-id]` 字面量同理。

仓库别处的约定是 selector 全都在 `selectors.ts` 里集中（这个文件已经在用 `promptSelector` / `sessionComposerDockSelector`），保证 data-attribute rename 时 grep 一处即可。建议加：

```ts
export const sessionTurnListSelector = '[data-slot="session-turn-list"]'
export const sessionMessageItemSelector = "[data-message-id]"
```

然后 spec 这边 import 进来。

packages/app/src/pages/session.tsx
```diff
@@ -694,11 +694,17 @@ export default function Page() {
   const turnDiffs = createMemo(() => list(lastUserMessage()?.summary?.diffs))
   const [artifactHistory, { refetch: refetchArtifactHistory }] = createResource(
     () => params.id,
-    (sessionID) => sdk.client.session.artifacts({ sessionID }).then((res) => res.data ?? []).catch(() => []),
+    async (sessionID) => ({
+      sessionID,
```
Astro-Han, line 698: **P3** — `{ initialValue: [] }` 里 bare `[]` TS 倾向于推成 `never[]`，于是 resource 的元素类型可能被静默弱化。当前 typecheck 过是因为 `deriveArtifactFiles` 接受 `SessionArtifactFile[]`、union 把 `never[]` 吃掉了，但任何对 `deriveArtifactFiles` 签名的修改都可能让这里失真。

建议显式标注：

```ts
{ initialValue: [] as SessionArtifactFile[] }
// 或
{ initialValue: [] satisfies SessionArtifactFile[] }
```

零成本，未来不用再担心。

packages/app/src/pages/session.tsx
```diff
@@ -694,11 +694,17 @@ export default function Page() {
...
+      artifacts: await sdk.client.session.artifacts({ sessionID }).then((res) => res.data ?? []).catch(() => []),
+    }),
+    { initialValue: { sessionID: "", artifacts: [] } },
```
Astro-Han: **P3** — 设 `initialValue: []` 之后，`artifactHistory.latest` 和 `artifactHistory()` 任何时候都返回 `T[]`、不会 `undefined`。所以 `?? []` 这一段是不可达分支。

要么收成 `const history = artifactHistory.latest`，要么补一行 comment 说为什么留着防御。现在的写法读起来像作者也不确定要用哪一种形式。

（看到你在 Gemini thread 里说会换成 keyed snapshot。如果 follow-up commit 直接把这一段重写了，忽略本条。）

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        composerDock: document.querySelectorAll(composerDockSelector).length,
+        messageList: document.querySelectorAll(messageListSelector).length,
+        messages: document.querySelectorAll("[data-message-id]").length,
+      })
+      const changed = (a: StabilitySample, b: StabilitySample) =>
+        a.url !== b.url ||
+        a.composerDock !== b.composerDock ||
+        a.messageList !== b.messageList ||
+        a.messages !== b.messages
+      const samples = [read()]
```
Astro-Han, line 41: **P3** — `samples` 没有上限。`changed()` 滤波在正常情况下会把数组压得很小，但一旦真的 regress 成持续 flicker，假设 30s prompt 流程里 200 帧每帧都 change，samples 就会爆到 200+，`expect(unmounted).toEqual([])` 输出的 diff 会变得没法读。

加一个软上限就够了：

```ts
if (samples.length >= 256) return
samples.push(next)
```

或者 record 到 256 之后只保留 first 128 + last 128，让失败信息保持可读。

packages/app/e2e/session/session-send-stability.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+async function installSessionStabilityProbe(page: Page) {
+  await page.evaluate(
+    ({ composerDockSelector, messageListSelector }) => {
+      const read = () => ({
+        at: performance.now(),
+        url: window.location.href,
+        composerDock: document.querySelectorAll(composerDockSelector).length,
+        messageList: document.querySelectorAll(messageListSelector).length,
+        messages: document.querySelectorAll("[data-message-id]").length,
```
Astro-Han: **P3** — `StabilitySample` 类型定义在模块顶层（Node 域），却被 `page.evaluate` 的 callback（browser 域）里的 `changed = (a: StabilitySample, b: StabilitySample)` 引用。能跑只是因为 TS 把类型擦掉了，跨 realm 引用是 smell。

两个干净的写法：
1. callback 内部重新声明一个局部类型；
2. 直接去掉 `a`/`b` 的注解，让 TS 从 `read()` 推断（最省事）。

外层的 `StabilitySample`（用于 `stopSessionStabilityProbe` 的返回类型断言）保留即可。

Astro-Han (commented):
鸡蛋里挑骨头 pass (P2/P3 only). 跳过 Gemini / CodeRabbit 已经盯过的 `page: any`、selector constant 命名、`.latest` 的陈旧数据问题。下面 8 条都是没人提过的小毛病，按 P-level 分级。

- P2: 2 条（probe 实际抓 flicker 的能力 + 失败时的 rAF 泄漏）
- P3: 6 条（命名、selectors 集中、类型推断、死代码、cross-realm 类型引用、双 poll 预算）

packages/app/e2e/session/session-send-stability.spec.ts
Astro-Han: Fixed. The probe now uses both requestAnimationFrame sampling and a MutationObserver. The observer inspects removedNodes directly, so a turn list or composer dock remove and reattach inside one mutation delivery is still recorded as an unmount.

packages/app/e2e/session/session-send-stability.spec.ts
Astro-Han: Fixed. Everything after probe installation is wrapped in try/finally, and stop cancels the animation frame, disconnects the observer, deletes the test probe, and returns partial samples for debugging.

packages/app/e2e/session/session-send-stability.spec.ts
Astro-Han: Fixed. The test probe now lives at `__opencode_e2e.sessionMountProbe`, so the window global is behavior-named and aligned with the existing E2E namespace.

packages/app/e2e/session/session-send-stability.spec.ts
Astro-Han: Fixed. Added `sessionTurnListSelector` and `sessionMessageItemSelector` to shared E2E selectors and updated the spec to import them.

packages/app/src/pages/session.tsx
Astro-Han: Fixed. The keyed artifact-history initial value now explicitly types the empty artifact list as `SessionArtifactFile[]`.

packages/app/src/pages/session.tsx
Astro-Han: Covered by the keyed snapshot rewrite. The old `artifactHistory.latest ?? artifactHistory() ?? []` path is gone; the current code reads `artifactHistory.latest` and only uses it when its `sessionID` matches the current route.

packages/app/e2e/session/session-send-stability.spec.ts
Astro-Han: Fixed. Reported samples and unmount samples are capped, while `unmountCount` tracks detection independently so later instability is not hidden by the output cap.

packages/app/e2e/session/session-send-stability.spec.ts
Astro-Han: Fixed. Removed the cross-realm `StabilitySample` reference inside `page.evaluate`; the browser callback now infers from its local `read()` function.

PR #265, opened 2026-04-27
refactor: rename @opencode-ai/shared to @opencode-ai/core (#24309)

Astro-Han:
## Summary

Mechanical rename of the workspace package `@opencode-ai/shared` to `@opencode-ai/core`. `packages/shared` moves to `packages/core`; all consumer imports update from `@opencode-ai/shared/*` to `@opencode-ai/core/*`. No file content changes beyond the rename.

PawWork-local consumers updated: `packages/util`, `packages/ui`, `packages/opencode`, `packages/desktop-electron` runtime-import-guard test.

## Why

First half of upstream's "shared → core" refactor (PR [#24309](https://github.com/anomalyco/opencode/pull/24309)). This PR is intentionally scoped to the rename only so the diff stays mechanical and reviewable. The follow-up commits that move modules (Global, log, flag, cross-spawn, npm) into core land in PR2 of #209.

This is part of the weekly upstream sync per `project_upstream_strategy.md`.

## Related Issue

#209

## How To Verify

```bash
bun install
bun run --cwd packages/core typecheck
bun run --cwd packages/opencode typecheck
bun run --cwd packages/util typecheck
bun run --cwd packages/ui typecheck
bun run --cwd packages/desktop-electron typecheck
```

All five packages typecheck clean locally on this branch.

## Checklist

- [x] I linked the related issue (#209)
- [x] This PR has type, scope, and priority labels (upstream / harness / P3)
- [x] I listed the relevant verification steps
- [ ] No visible UI changes — N/A
- [ ] No desktop/packaging/updater impact — N/A (workspace rename only)
- [x] I considered docs/dependencies — `bun.lock` regenerated; no new deps
- [x] Targeting `dev`, Conventional Commits English title

## Summary by CodeRabbit

* **Chores**
  * Renamed and reorganized an internal utilities package and updated imports/dependencies across the codebase to reference the new namespace. No functional or user-facing behavior changed; this improves internal consistency and module resolution.

Issue #209, opened 2026-04-24
[Feature] Track selective upstream backports from opencode v1.14.17-v1.14.22


Astro-Han:
## What task are you trying to do?

We need a focused review of recent upstream `anomalyco/opencode` changes and decide which upstream PRs are worth selectively backporting into PawWork. The issue header originally said `v1.14.17` through `v1.14.22`, but the actual PawWork sync baseline is `v1.14.20` commit `3175a3c618`, landed through issue #92 and PR #155. The current review window is `v1.14.20..upstream/dev`, 88 commits, with upstream `dev` at `a4bd88ab97` on 2026-04-24.

Use GitHub issue/PR history and `git log` as the source of truth for prior upstream work. Local upstream sync docs are obsolete and should not be treated as current guidance. In particular, do not use `docs/upstream-sync.md` or `docs/upstream_sync.md` as source material.

## What do you do today?

Today the upstream delta is checked manually from release notes, PRs, and tree diffs. This works for a quick scan, but it is easy to mix already absorbed fixes, small safe backports, large schema migrations, UI fork changes, TUI-only work, and upstream history rewrite artifacts. Upstream also rewrote `dev` history in this window, so do not rely only on a PR's GitHub `mergeCommit.oid` for ancestor checks or cherry-pick planning. Verify by PR number and current file content.

## What would a good result look like?

A good result is a short implementation plan that ranks upstream PRs by PawWork value and integration risk, then opens small backport PRs only for the selected items. Each backport should keep one reversible intent, avoid PawWork frontend fork churn, and include focused verification for the touched surface.

Recommended execution shape after independent review:

1. Treat the Effect Schema migration as one coherent slice, starting from PawWork's post-PR #102 tree rather than old `v1.4.11` assumptions. Include the schema chain only if the slice compiles as a unit and updates PawWork-owned `packages/opencode/src/tool/trash.ts` plus `packages/opencode/test/tool/trash.test.ts` to the new tool framework. Keep PawWork tool description text files on the local side.
2. Allow disjoint small PRs to run before or alongside the schema slice when their touched files do not overlap. Good parallel candidates: `anomalyco/opencode#19054`, `#24001` plus `#24016`, and `#23791`.
3. Put schema-adjacent small PRs after the slice unless the implementer proves the conflicts are cheaper the other way. These include `#24063`, `#23797`, `#23770`, and `#23870`.

Initial high-value candidates to re-check and likely backport:

[...]

Issue #27, opened 2026-04-18
[Feature] Sync upstream opencode v1.4.6 → v1.4.11 (graft + multi-PR)


Astro-Han:
## What task are you trying to do?

Keep PawWork's vendored copy of `packages/opencode` and shared infrastructure in lockstep with upstream `anomalyco/opencode` so we do not accumulate architectural drift. The foundation strategy is now stable: opencode remains the foundation until PawWork crosses 10K users. We do not have the team to fork the foundation; staying current is non-negotiable infrastructure work.

This issue tracks the first full sync cycle from our fork point (`v1.4.6` area) to upstream `v1.4.11`, which contains 309 commits including namespace unwrap, zod -> Effect Schema migration, HttpApi route changes, and workspace control-plane refactors.

## What do you do today?

We previously used a pinned vendor fork plus selective cherry-picks. That worked for one cycle but is no longer enough. Skipping the architectural migrations now would make every future cherry-pick pay import-path, schema-shape, and route-contract translation cost.

A second structural problem is history shape. Our initial commit `bde4c3927 initial: based on OpenCode (MIT License)` is an orphan snapshot import, not a normal fork. A naive merge against upstream creates hundreds of false add/add conflicts.

## What would a good result look like?

After this issue closes:

- `packages/opencode` and shared infrastructure reflect upstream `v1.4.11` semantics
- UI rewrite carve-out paths stay on `HEAD` and are still owned by [#26](https://github.com/Astro-Han/pawwork/issues/26)
- PawWork history keeps only PawWork-authored commits, not upstream contributor history
- Future syncs can reuse the same graft + squash workflow with smaller marginal cost
- Phase 1 macOS GUI flows still work end to end

The workflow itself lives in `docs/upstream-sync.md`. This issue tracks live status and lane ownership, not the whole protocol.

## Live status

| # | Branch | Status |
|---|---|---|
| 1 | `sync/setup` | done locally |
| 2 | `sync/opencode-tool` | merged as [#31](https://github.com/Astro-Han/pawwork/pull/31) |
| 3 | `sync/opencode-provider-plugin` | merged as [#33](https://github.com/Astro-Han/pawwork/pull/33) |
| 4 | `sync/opencode-provider-runtime-bridge` | merged as [#34](https://github.com/Astro-Han/pawwork/pull/34) |
| 5 | `sync/opencode-runtime-core` | merged as [#36](https://github.com/Astro-Han/pawwork/pull/36) |
[...]

Issue #26, opened 2026-04-18
[Feature] Land the new PawWork desktop UI in phased PRs


Astro-Han:
### What task are you trying to do?

I want to land the next PawWork desktop UI in the real app.

The target UI for this issue is still:

- A desktop shell with three clear areas: left navigation, center session view, right utility panel.
- A new-session screen with one large input box and three starter cards.
- A left sidebar that makes sessions easy to scan.
- An active session view with a simple header, a centered message timeline, and a composer fixed at the bottom.
- A fuller settings experience instead of hiding it inside a compact dialog.

This is a phased redesign of the existing Solid + Electron app. It is not a prototype import.

### What do you do today?

As of April 19, 2026, Phase 1 items 1 to 3 have already landed on `dev` through [#35](https://github.com/Astro-Han/pawwork/pull/35):

- theme and shell polish
- new-session home and composer
- right utility panel

The remaining gap is now mostly Phase 2 UI work:

- The left side is still organized around projects and worktrees first. Sessions are still harder to scan than they should be.
- Session discovery and organization are not yet close to the target browsing model.
- Settings still lives in a dialog rather than a fuller page-style surface.

So the current app is much closer to the target than before, but the navigation and settings experience still do not present PawWork in the shape we want people to use.

### What would a good result look like?

A good result is a sequence of small PRs that completes the remaining UI work without rewriting the shell.

The next slices under this issue should focus on:

1. Sidebar redesign on top of the current data model
   - keep the existing project/worktree model in code
   - make the visible sidebar much closer to the target layout
2. Session organization
   - pinned sessions
   - recent sessions
   - better session discovery from the command palette
3. Settings redesign
   - move from the current compact dialog shape toward a fuller desktop settings page

### Which audience does this matter to most?

Both

### Extra context

**Current UI design reference for this issue lives at `docs/design/` in the main checkout. It is intentionally local-only and is not tracked in git.** Do not assume worktrees, CI, or fresh clones can see it.

**For implementation PRs, use `docs/design/` as the local visual and interaction reference when it is available, but keep this issue's scope boundaries as the source of truth for what belongs in `#26`.** If a PR depends on details from `docs/design/`, copy the needed constraints into the PR body or an issue comment so reviewers can see them without local-only files.

Relevant design files:

| Remaining slice | Source files | HANDOFF section |
|---|---|---|
| Sidebar redesign | `docs/design/src/sidebar.jsx` | §3 Sidebar |
[...]

PR #266, opened 2026-04-27
refactor: move Global/log/flag/cross-spawn/npm into @opencode-ai/core

Astro-Han:
## Summary

Pulls the upstream "core package consolidation" follow-ups to PR #265 (rename), plus the npm-config snapshot from upstream HEAD:

- **1a734adb4d** core: consolidate shared infrastructure into core
- **705f792e87** core: move Global module to @opencode-ai/core
- **3eee2f6afa** core: move cross-spawn-spawner from opencode to core
- **1e98167b0e** core: move cross-spawn-spawner to root + remove unused types
- **f5dce6d960** core: move npm service to core
- Stage C snapshot: `packages/core/src/npm.ts`, `npm-config.ts`, fixtures, tests pulled from upstream HEAD `17701628bd`

After this PR, `packages/core` owns: `Global`, `Runtime`, log, flag, cross-spawn-spawner, filesystem, util/*, npm, npm-config, installation/version, effect/{logger, memo-map, observability, runtime}.

## Why

Stage C of the foundation sync (#209). Once core owns these modules, the rest of the upstream backlog (Effect Schema migration in Stage D, tool framework in Stage E) can land cleanly. The accidental "Global module hardcoded to opencode" regression that PawWork caught here is fixed by the carve-out below.

## PawWork carve-outs in this PR

- `packages/core/src/global.ts` calls `Runtime.appName()` instead of hardcoded `"opencode"`. Preserves the separate `~/.config/pawwork/` data dir per `feedback_pawwork_positioning.md`. Will need to be reapplied each weekly sync (added to permanent carve-out list).
- `packages/opencode/src/global/index.ts` kept as PawWork's local Runtime-aware Global module. Upstream's auto-rewritten `@opencode-ai/core/global` imports flow through to the carve-out.
- `packages/core/src/effect/logger.ts` keeps PawWork's namespace-style `Log` API (`Log.create` / `Log.Default` / `Log.init`) instead of upstream HEAD's flat-export style. PawWork callers already use the namespace shape.

## Mechanical cleanup

- `packages/opencode/src/installation/meta.ts` removed; callers use `@opencode-ai/core/installation/version` (`InstallationVersion` / `InstallationChannel`).
- `~120` PawWork-side import rewrites (`../util/log` → `@opencode-ai/core/util/log`, `../flag/flag` → `@opencode-ai/core/flag/flag`, `@/effect/cross-spawn-spawner` → `@opencode-ai/core/cross-spawn-spawner`, etc.).
- `@effect/opentelemetry: 4.0.0-beta.46` added to root catalog (matches the rest of the effect ecosystem).

## Related Issue

#209

## How To Verify

```bash
bun install
bun run --cwd packages/core typecheck
bun run --cwd packages/opencode typecheck
bun run --cwd packages/util typecheck
bun run --cwd packages/ui typecheck
bun run --cwd packages/desktop-electron typecheck
```

All 5 packages typecheck clean locally on this branch.

## Stacked on

This PR's base is `claude/upstream-sync-209-pr1-rename` (#265). After #265 merges, GitHub will auto-rebase this PR onto `dev`.

## Checklist

- [x] I linked the related issue (#209)
- [x] This PR has type, scope, and priority labels (upstream / harness / P3)
- [x] I listed the relevant verification steps
- [x] No visible UI changes — N/A
- [x] No desktop/packaging/updater impact — workspace package boundary only
- [x] I considered docs/dependencies — added `@effect/opentelemetry` catalog entry; bun.lock regenerated
- [x] Targeting `dev` (via PR #265 base), Conventional Commits English title


Astro-Han:
Rebased onto `dev` after #265 merged. The 5 review findings (top-level `fs.mkdir` in `global.ts`, `import.meta.resolve` in `npm.ts`, silent error swallow in `npm-config.ts`, `OTEL_RESOURCE_ATTRIBUTES` parsing in `observability.ts`, `package-lock.json` deletion in `npm.ts:222`) are all pre-existing upstream behaviors that this PR moves verbatim from `packages/opencode` (or pulls from upstream HEAD `17701628bd` for the npm snapshot). The only intentional PawWork carve-out in `packages/core` here is `global.ts:9` switching the hardcoded `"opencode"` to `Runtime.appName()` — see PR description "PawWork carve-outs in this PR".

Per AGENTS.md ("commit small / one reversible intent"), behavior changes don't belong in a mechanical move PR. If any of these prove worth fixing for PawWork, they should be tracked as separate scoped issues (e.g. lazy `Path.ensure()` instead of import-time `mkdir`; `which` no longer deleting `package-lock.json`). Resolving the threads as out-of-scope here.

Astro-Han (commented):
## PR #266 Review — Core consolidation follow-ups

Reviewed the full diff (159 files, +975 / -574) focusing on behavioral regressions, boundary conditions, and PawWork carve-out integrity. Findings below, sorted P0→P3.

---

### P0 — Must fix before merge

**1. `global.ts` top-level side effects on import — xdg `!` non-null assertions can silently produce `"undefined/pawwork"` paths**

`packages/core/src/global.ts` evaluates `xdgData!`, `xdgCache!`, `xdgConfig!`, `xdgState!` at module top level. On Linux systems or containers where these XDG env vars are unset, `xdgData` etc. are `undefined`, and `path.join(undefined!, "pawwork")` produces `"undefined/pawwork"` — silently, with no error. The old code had the same `!` assertions but they ran inside an Effect Layer where a failure would propagate. Now the bad path is baked into `Path.data` etc. at import time with no recovery. Both the core `global.ts` and the opencode local `global/index.ts` have this issue.

The top-level `await Promise.all([fs.mkdir(...)])` and `Flock.setGlobal()` also execute on every import. If any consumer imports `Log` from `@opencode-ai/core/util/log` (which chains to `../global`), directory creation and flock wiring happen as a side effect of that import. This is a contract change from the old code where `global.ts` was side-effect-free at import time.

**2. `Npm.add()` skips `loadVirtual()` — corrupted package dirs treated as valid**

Old `packages/opencode/src/npm/index.ts` calls `arborist.loadVirtual()` first to verify tree integrity; only if that fails or the package is absent does it call `reify()`. This means a partially-installed or corrupted `node_modules` gets repaired on the next `add()`.

New `packages/core/src/npm.ts` replaces this with `afs.existsSafe(dir)` — if the directory exists, it returns the cached entry point immediately without verifying the tree. A corrupted install (e.g. interrupted `reify()`, deleted `node_modules` inner files) will be treated as valid, and LSP servers / plugins will fail at spawn time with no self-repair path.

---

### P1 — Should fix before merge (user-facing regression or high risk)

**3. `emitLspInstallFailedIfApplicable()` removed — LSP install failure UI notification is dead**

The old `npm/index.ts` published `LSP.Event.InstallFailed` to the Bus when an LSP server package failed to install. The `packages/opencode/src/lsp/index.ts` still defines `Event.InstallFailed` and still checks `err instanceof Npm.InstallFailedError`, but the new core `npm.ts` never publishes this event — it can't, since core has no knowledge of LSP or Bus. The SDK type `EventLspServerInstallFailed` still exists. The result: LSP install failures will be caught by the cooldown logic in `lsp/index.ts`, but no `InstallFailed` Bus event is ever fired, so any UI that renders a toast or notification for this event is dead code.

**4. `InstallFailedError` schema changed — `{ pkg: string }` → `{ add?: string[], dir: string, cause?: Defect }`**

`packages/opencode/src/lsp/index.ts` line checks `err instanceof Npm.InstallFailedError` — this will still match because the class name is the same. However, any code that reads `err.payload.pkg` (or `err.pkg`) will get `undefined` instead of the package name. The LSP error handler doesn't read the payload today, but the Bus event schema `z.object({ pkg: z.string(), error: z.string() })` expects `pkg` which no longer exists on the error object. If anyone wires up the event again, the Zod validation will fail.

**5. `serviceName: "opencode"` hardcoded in `observability.ts` — PawWork traces report as "opencode"**

`packages/core/src/effect/observability.ts` `resource()` returns `serviceName: "opencode"`. PawWork's OTel traces and logs will be attributed to the `"opencode"` service. This should use `Runtime.appName()` like the Global carve-out does, so PawWork reports as `"pawwork"`. If this is intentional for upstream compatibility it should be documented in the carve-out list.

**6. Dual Global module — cache version check TOCTOU race**

Both `packages/core/src/global.ts` and `packages/opencode/src/global/index.ts` independently run the cache version check (read version → if mismatch, `readdir` + `rm` all cache contents → write new version). When both modules are loaded in the same process (which happens — core is imported by e.g. `@opencode-ai/core/util/log`, and opencode's local `global/index.ts` is imported by ~12 still-unmigrated files), these two checks race. If the cache version is being bumped, one module could delete files the other is reading, or write the version file before the other finishes cleaning. Low frequency (only fires on version bumps) but can corrupt cache state.

---

### P2 — Should fix soon (maintainability / correctness)

**7. Windows Bun workaround removed — `__FAKE_PLATFORM__ = "linux"` gone**

Old `npm/index.ts` sets `process.env.__FAKE_PLATFORM__ = "linux"` before importing Arborist on Windows, because "Bun on Windows does not support the UV_FS_O_FILEMAP flag used by tar." New `npm.ts` does a plain `import("@npmcli/arborist")` without this workaround. If PawWork supports Windows + Bun (the project is a macOS + Windows desktop product per AGENTS.md), npm package installation may fail on Windows.

**8. `NpmConfig.load()` silently swallows all errors → returns `{}`**

`packages/core/src/npm-config.ts` catches all exceptions in `load()` and returns an empty object. This means authentication errors (bad `.npmrc` tokens), misconfigured registries, or corrupt config files are silently ignored. The caller proceeds with default registry + no auth, producing confusing "401 Unauthorized" errors downstream instead of an upfront clear failure.

**9. `Npm.outdated()` removed with no replacement**

The old `Npm.outdated(pkg, cachedVersion)` function checked the npm registry for newer versions. While no current code calls it, it was a public export of the `Npm` namespace. Any future code that needs version checking will need to re-implement this.

**10. Inconsistent import styles for `CrossSpawnSpawner` and `EffectLogger`**

Some files use `import * as CrossSpawnSpawner from "@opencode-ai/core/cross-spawn-spawner"` while others use `import { CrossSpawnSpawner } from "..."`. Both produce the same runtime result because the module uses `export * as CrossSpawnSpawner from "./cross-spawn-spawner"`, but the inconsistency is a maintenance hazard. Same for `EffectLogger` — `import * as EffectLogger` vs `import { EffectLogger }`. Recommend standardizing on `{ CrossSpawnSpawner }` / `{ EffectLogger }` since those are the explicit named namespace exports.

**11. `install()` writable check silently masks permission errors**

New `npm.ts` `install()` checks `canWrite` and returns early if false. Old code attempted the operation regardless. If a directory exists but is read-only (common in CI, Docker, or restricted environments), the install silently does nothing instead of surfacing the permission problem. The caller gets no error and no indication that dependencies weren't installed.

---

### P3 — Nice to have / low risk

**12. `types.d.ts` removal drops `Bun` global type declaration**

`packages/core/src/types.d.ts` had `declare var Bun: ... | undefined`. If `@types/bun` isn't configured for the core package, other files referencing the `Bun` global may get implicit `any`. The new `npm.ts` uses `typeof Bun !== "undefined"` which works at runtime, but type safety is reduced.

**13. `Npm.which()` behavioral change — returns `undefined` instead of falling back to first binary**

Old `which()` with no `bin` parameter: if the requested binary isn't found, falls through to `add()` and returns whatever `pick()` resolves. New `which()` with `bin` parameter: if the specific `bin` isn't in `.bin/`, returns `Option.none()`. This is actually better behavior (explicit failure vs silent fallback), but is a behavioral change that could surprise callers.

**14. Half-migrated Global imports — 12 opencode source files still import from local `../global`**

Files like `src/storage/db.ts`, `src/snapshot/index.ts`, `src/lsp/server.ts`, etc. still import `Global` from the local opencode module rather than `@opencode-ai/core/global`. This means both Global modules are loaded in every process, persisting the dual-module risks (#6, #10). Completing the migration and deleting the local `global/index.ts` (or making it re-export from core) would eliminate the entire class of dual-Global issues.

**15. `mergeWithExisting: false` in observability logs is a behavioral change from old `loggerMergeWithExisting: true`**

Old `packages/opencode/src/effect/observability.ts` used `loggerMergeWithExisting: true`. New code uses `mergeWithExisting: false`. The new code explicitly composes `EffectLogger.logger + OtlpLogger.make()` so both are present, but if any downstream code added custom loggers to the Effect runtime independently, they'd be dropped when OTel is enabled. The non-OTel path is unaffected.

**16. `setGlobalContextManager` is a global singleton mutation in `traces()`**

`packages/core/src/effect/observability.ts` calls `context.setGlobalContextManager(mgr)`. This cannot be undone and will throw or silently overwrite if called more than once. In test environments that import this module multiple times or across test isolation boundaries, this could cause failures.

Astro-Han (commented):
## PR #266 Review — Supplemental findings (round 2)

Deeper inspection of barrel files, dead code, cross-module dependency chains, and boundary conditions missed in the first pass.

---

### P1 — Should fix before merge

**17. `resource()` hardcodes `ensureProcessMetadata("main")` — worker processes get wrong OTel process role**

`packages/core/src/effect/observability.ts:281` calls `ensureProcessMetadata("main")` every time `resource()` is evaluated. If a worker process sets `OPENCODE_PROCESS_ROLE = "worker"` via `ensureProcessRole("worker")` BEFORE the observability layer is built for the first time, the `??=` operator will leave `"worker"` intact and the role is correct. But if the observability layer is built first (e.g., during app startup), `??=` will set the env var to `"main"` permanently, and the worker's OTel data will be incorrectly labeled as `opencode.process_role: main`.

The fix: either don't call `ensureProcessMetadata` in `resource()` (use the env var directly with a fallback), or pass a role parameter through the layer construction chain.

---

### P2 — Should fix soon

**18. `packages/opencode/src/effect/oltp.ts` is dead code — duplicate `Observability` namespace ghost**

After `packages/opencode/src/effect/observability.ts` was deleted (replaced by core), the sibling file `oltp.ts` still exists in opencode with its own `export namespace Observability`. No code imports from it anymore — both the barrel (`packages/opencode/src/effect/index.ts`) and `run-service.ts` now import from `@opencode-ai/core/effect/observability`. If any future code accidentally imports `Observability` from `@/effect/oltp`, it would get a stale, different implementation with no trace support and the old `loggerMergeWithExisting: true` behavior.

Recommendation: delete `packages/opencode/src/effect/oltp.ts`.

**19. `sanitize()` is a no-op on non-Windows — no path traversal guard for `..` in package names**

`packages/core/src/npm.ts:sanitize()` returns the input unchanged on macOS/Linux. The `directory()` function does `path.join(global.cache, "packages", sanitize(pkg))`. On non-Windows, a crafted package name containing path traversal sequences (e.g., `../../etc` passed as the pkg argument) would escape the cache directory. While `npa()` and Arborist validate package names against npm registry rules, the internal `directory()` function doesn't enforce this. A second defense-in-depth guard (e.g., `path.basename(pkg) === pkg` check) would be safer.

**20. `packages/core/src/util/log.ts` triggers core `global.ts` top-level side effects on import**

The core `log.ts` does `import * as Global from "../global"` at module scope. Since `core/global.ts` has top-level `await` (directory creation + cache version check), importing `Log` from `@opencode-ai/core/util/log` now creates directories as a side effect. This is a contract change from the old `packages/opencode/src/util/log.ts` which imported `Global` from the old opencode global that also had top-level directory creation — so this was already a side-effecting import. However, the new import chain is longer (Log → core/global → Flock → Hash, including the `Flock.setGlobal()` call), and the xdg `!` non-null assertions in core/global can now fail silently if `xdgData` is undefined.

**21. `packages/opencode/src/effect/index.ts` barrel and `oltp.ts` export conflicting `Observability` namespaces**

```ts
// barrel: packages/opencode/src/effect/index.ts
export { Observability } from "@opencode-ai/core/effect/observability"  // has traces + new setup

// dead file: packages/opencode/src/effect/oltp.ts
export namespace Observability { ... }  // old Otlp.layerJson, no traces
```

TypeScript won't complain about this since nothing imports from `oltp.ts`. But if the barrel is the only import path, it'll always resolve to core. Risk: a search-and-replace or IDE auto-import could inadvertently wire up the dead `oltp.ts` Observability.

---

### P3 — Nice to have

**22. `loggerExportInterval: Duration.seconds(1)` removed — log batching behavior may change**

The old `packages/opencode/src/effect/observability.ts` used `Otlp.layerJson({ loggerExportInterval: Duration.seconds(1), ... })`. The new core `observability.ts` uses `OtlpLogger.make({ url, resource, headers })` without an explicit export interval. The default interval for `OtlpLogger` may differ (likely 5s or 30s depending on the Effect version). This could change OTel log delivery latency for PawWork telemetry.

**23. `resource()` not memoized — called every time `logs()` is invoked**

`logs()` creates a new `Logger.layer` with `OtlpLogger.make({ url, resource: resource(), headers })`. Each call to `resource()` re-computes the attributes object (including `ensureProcessMetadata`, `OTEL_RESOURCE_ATTRIBUTES` parsing, and the spread). While `ensureProcessMetadata` is idempotent (`??=`), the `OTEL_RESOURCE_ATTRIBUTES` parsing and object creation happen fresh each time. This is low-impact (happens once per runtime construction) but needless work.

**24. `packages/core/package.json` has `"bin": { "opencode": "./bin/opencode" }` — wrong for a library package**

The core package exports `"bin": { "opencode": "./bin/opencode" }`. Since `@opencode-ai/core` is a dependency library (not a directly-installed binary), this bin entry has no effect unless someone `npm install -g @opencode-ai/core` or another package hoists it. Harmless but misleading — if this bin entry is intended for the opencode package, it should be in `packages/opencode/package.json` instead.

**25. `ensureProcessMetadata("main")` hardcoded in `observability.ts` but `opencode-process.ts` API expects a `fallback` parameter**

`ensureProcessMetadata(fallback)` accepts `"main" | "worker"`. The observability code hardcodes `"main"`. If the observability layer is used in a worker process where `OPENCODE_PROCESS_ROLE` is already `"worker"`, the `??=` operator preserves the existing value and the hardcoded `"main"` is just a fallback. However, the fallback parameter name implies it should match the caller's context — if a worker builds the layer, the semantic fallback should be `"worker"`, not `"main"`. Currently this is fine because the observability layer is only built in the main process, but it's fragile documentation.

**26. `@effect/opentelemetry` catalog entry added to root but `@opentelemetry/context-async-hooks` is a direct dependency in core/package.json**

The root `package.json` adds `"@effect/opentelemetry": "4.0.0-beta.46"` to the catalog. The core `package.json` lists `@opentelemetry/context-async-hooks: 2.6.1` as a direct dependency (not catalog). This is fine since it's a transitive peer, but if this dependency is moved to the catalog in the future, the version pin needs to stay synchronized.

**27. Self-referential `export * as Global from "./global"` in `core/global.ts` could confuse TypeScript autocompletion**

The pattern creates `Global.Global.Global...` in IntelliSense. For developers working in the core package, autocomplete will show the recursive self-reference as a valid path. Purely cosmetic but worth standardizing — either use `export namespace Global { ... }` (old style) or rename the re-export label.

packages/core/src/effect/observability.ts
```diff
@@ -0,0 +1,107 @@
...
+    },
+  }
+}
+
+function logs() {
+  return Logger.layer(
+    [
+      EffectLogger.logger,
+      OtlpLogger.make({
+        url: `${base}/v1/logs`,
```
Astro-Han, line 62: **P2**: `OtlpLogger.make({})` without explicit `loggerExportInterval`. The deleted `packages/opencode/src/effect/observability.ts` used `Otlp.layerJson({ loggerExportInterval: Duration.seconds(1), ... })`. The default interval for `OtlpLogger` may differ (likely 5-30s), which could change OTel log delivery latency.

packages/core/src/effect/observability.ts
```diff
@@ -0,0 +1,107 @@
...
+        acc[key] = value.join("=")
+        return acc
+      },
+      {} as Record<string, string>,
+    )
+  : undefined
+
+export function resource(): { serviceName: string; serviceVersion: string; attributes: Record<string, string> } {
+  const processMetadata = ensureProcessMetadata("main")
+  const attributes: Record<string, string> = (() => {
```
Astro-Han, line 27: **P1**: `ensureProcessMetadata("main")` hardcodes the process role. If a worker process builds this observability layer before `ensureProcessRole("worker")` sets `OPENCODE_PROCESS_ROLE`, the `??=` locks the env var to "main" permanently. Worker OTel data will be mislabeled. Fix: read env var directly with a fallback, or thread the role through the layer chain.

packages/core/src/npm.ts
```diff
@@ -0,0 +1,267 @@
...
+      return resolveEntryPoint(first.name, first.path)
+    }, Effect.scoped)
+
+    const install: Interface["install"] = Effect.fn("Npm.install")(function* (dir, input) {
+      const canWrite = yield* afs.access(dir, { writable: true }).pipe(
+        Effect.as(true),
+        Effect.orElseSucceed(() => false),
+      )
+      if (!canWrite) return
+
```
Astro-Han, line 166: **P2**: `canWrite` early-return silently masks permission errors. Old code attempted the operation regardless and would surface the error. In read-only environments (CI, Docker) this silently does nothing instead of surfacing the problem. Consider logging a warning at minimum when returning early.

packages/core/src/npm.ts
```diff
@@ -0,0 +1,267 @@
...
+import { Effect, Schema, Context, Layer, Option, FileSystem } from "effect"
+import { NodeFileSystem } from "@effect/platform-node"
+import { AppFileSystem } from "./filesystem"
+import { Global } from "./global"
+import { EffectFlock } from "./util/effect-flock"
+import { makeRuntime } from "./effect/runtime"
+import { NpmConfig } from "./npm-config"
+
+export class InstallFailedError extends Schema.TaggedErrorClass<InstallFailedError>()("NpmInstallFailedError", {
+  add: Schema.Array(Schema.String).pipe(Schema.optional),
```
Astro-Han, line 13: **P1**: `InstallFailedError` schema changed from `{ pkg: string }` to `{ add?: string[], dir: string, cause?: Defect }`. `packages/opencode/src/lsp/index.ts` still checks `err instanceof Npm.InstallFailedError` which will match, but `err.payload.pkg` is now `undefined`. The LSP Bus event schema `z.object({ pkg: z.string() })` would fail Zod validation if the event were ever published.

packages/core/src/npm.ts
```diff
@@ -0,0 +1,267 @@
...
+        Effect.withSpan("Npm.reify", {
+          attributes: input,
+        }),
+      )
+
+    const add = Effect.fn("Npm.add")(function* (pkg: string) {
+      const dir = directory(pkg)
+      const name = (() => {
+        try {
+          return npa(pkg).name ?? pkg
```
Astro-Han: **P2**: `loadVirtual()` optimization removed from `add()`. Old code called `arborist.loadVirtual()` to verify tree integrity before using cached packages; corrupted installs would be repaired by calling `reify()`. New code only checks `afs.existsSafe(dir)` — a corrupted or partially-installed `node_modules` directory is treated as valid. Consider calling `loadVirtual()` before the early return at line 118 to verify tree integrity.

packages/core/src/effect/observability.ts
```diff
@@ -0,0 +1,107 @@
...
+        }),
+      )
+    } catch {
+      return {}
+    }
+  })()
+
+  return {
+    serviceName: "opencode",
+    serviceVersion: InstallationVersion,
```
Astro-Han, line 45: **P1**: `serviceName: "opencode"` hardcoded. PawWork OTel traces/logs will be attributed to "opencode" instead of "pawwork". Use `Runtime.appName()` like the Global carve-out does, or add a comment documenting this as intentional for upstream compatibility.

packages/core/src/effect/observability.ts
```diff
@@ -0,0 +1,107 @@
...
+    [
+      EffectLogger.logger,
+      OtlpLogger.make({
+        url: `${base}/v1/logs`,
+        resource: resource(),
+        headers,
+      }),
+    ],
+    { mergeWithExisting: false },
+  ).pipe(Layer.provide(OtlpSerialization.layerJson), Layer.provide(FetchHttpClient.layer))
```
Astro-Han, line 68: **P2**: `mergeWithExisting: false` is a behavioral change from the old `loggerMergeWithExisting: true`. The old code merged OTel loggers into the existing Effect logger stack. The new code replaces all loggers. While `EffectLogger.logger` is explicitly included in the array, any custom loggers added independently to the runtime would be dropped when OTel is enabled. The non-OTel path (`!base`) is unaffected.

packages/core/src/npm.ts
```diff
@@ -0,0 +1,267 @@
...
+  ) => Effect.Effect<void, EffectFlock.LockError | InstallFailedError>
+  readonly which: (pkg: string, bin?: string) => Effect.Effect<Option.Option<string>>
+}
+
+export class Service extends Context.Service<Service, Interface>()("@opencode/Npm") {}
+
+const illegal = process.platform === "win32" ? new Set(["<", ">", ":", '"', "|", "?", "*"]) : undefined
+
+export function sanitize(pkg: string) {
+  if (!illegal) return pkg
```
Astro-Han, line 42: **P2**: `sanitize()` is a no-op on non-Windows (line below: `if (!illegal) return pkg`). On macOS/Linux, a package name containing `..` passes through unsanitized to `path.join(global.cache, "packages", sanitize(pkg))`. While `npa()` and Arborist validate npm package names, this internal helper should add a defense-in-depth guard (e.g. `path.basename(pkg) === pkg`).

packages/core/src/npm-config.ts
```diff
@@ -0,0 +1,40 @@
...
+// @ts-expect-error npm does not publish types for this internal config API.
+import Config from "@npmcli/config"
+// @ts-expect-error npm does not publish types for this internal config API.
+import { definitions, flatten, nerfDarts, shorthands } from "@npmcli/config/lib/definitions/index.js"
+import { Effect } from "effect"
+
+const npmPath = fileURLToPath(new URL("..", import.meta.url))
+
+export const load = (dir: string) =>
+  Effect.tryPromise({
```
Astro-Han, line 13: **P2**: `NpmConfig.load()` silently swallows all errors (catches and returns `{}`). Authentication errors (bad `.npmrc` tokens), misconfigured registries, or corrupt config files are silently ignored. The caller proceeds with default registry and no auth, producing confusing downstream errors like "401 Unauthorized" rather than a clear upfront failure. Consider logging a warning when the config fails to load.

packages/core/src/npm.ts
```diff
@@ -0,0 +1,267 @@
...
+    const afs = yield* AppFileSystem.Service
+    const global = yield* Global.Service
+    const fs = yield* FileSystem.FileSystem
+    const flock = yield* EffectFlock.Service
+    const directory = (pkg: string) => path.join(global.cache, "packages", sanitize(pkg))
+    const reify = (input: { dir: string; add?: string[] }) =>
+      Effect.gen(function* () {
+        yield* flock.acquire(`npm-install:${input.dir}`)
+        const { Arborist } = yield* Effect.promise(() => import("@npmcli/arborist"))
+        const add = input.add ?? []
```
Astro-Han, line 91: **P2**: Windows Bun workaround removed. Old `packages/opencode/src/npm/index.ts` set `process.env.__FAKE_PLATFORM__ = "linux"` before importing Arborist because "Bun on Windows does not support the UV_FS_O_FILEMAP flag used by tar". The new code does a plain `import("@npmcli/arborist")` without this workaround. If PawWork targets Windows + Bun (per SPEC.md), package installation may fail on that platform.

Astro-Han (commented):
## PR #266 Inline Review — P0–P3 summary

10 inline comments posted across 3 files covering critical issues. Full summary below.

### Inline comments posted

**`packages/core/src/effect/observability.ts`** (6 comments)
- L26: **P1** `ensureProcessMetadata("main")` hardcoded — worker process role mislabel
- L44: **P1** `serviceName: "opencode"` hardcoded — PawWork traces misattributed  
- L61: **P2** `loggerExportInterval` removed — log batching behavior change
- L67: **P2** `mergeWithExisting: false` — behavioral change from old code

**`packages/core/src/npm.ts`** (6 comments)
- L14: **P1** `InstallFailedError` schema change breaks Bus event validation
- L43: **P2** `sanitize()` no-op on non-Windows — no path traversal guard
- L82: **P2** `__FAKE_PLATFORM__` workaround removed — Windows+Bun may fail
- L117: **P2** `loadVirtual()` removed — corrupted packages not self-repaired
- L139: **P2** `canWrite` early return silently masks permission errors

**`packages/core/src/npm-config.ts`** (1 comment)
- L13: **P2** `NpmConfig.load()` silently swallows all errors

### Additional findings not covered by inline comments

- **P0**: `global.ts` xdg `!` non-null assertions produce `"undefined/pawwork"` paths when XDG vars are unset
- **P0**: Top-level `await` in `global.ts` triggers FS side effects on import — changes module border contract  
- **P1**: `emitLspInstallFailedIfApplicable()` removed — LSP install failure UI notification is dead code
- **P1**: Dual Global module cache version check TOCTOU race when both core and opencode globals load
- **P2**: `packages/opencode/src/effect/oltp.ts` is dead code with conflicting `Observability` namespace
- **P3**: Inconsistent `* as` vs `{ }` import styles for CrossSpawnSpawner/EffectLogger
- **P3**: Core `package.json` carries `"bin": { "opencode": ... }` — wrong package for a binary

### Quick action items

| Priority | Action |
|---|---|
| P0 | Guard `xdgData` etc. with fallback before `!` assertion in global.ts |
| P0 | Restore `loadVirtual()` tree integrity check in `Npm.add()` |
| P1 | Patch `serviceName` → `Runtime.appName()` for PawWork |
| P1 | Either restore LSP install failed event or document the removal |
| P1 | Fix `InstallFailedError` schema alignment with LSP Bus event |
| P2 | Restore `__FAKE_PLATFORM__` on Windows if Bun is supported |
| P2 | Delete `oltp.ts` dead code |

packages/core/src/global.ts
```diff
@@ -1,43 +1,66 @@
...
+
+export class Service extends Context.Service<Service, Interface>()("@opencode/Global") {}
+
+export interface Interface {
+  readonly home: string
+  readonly data: string
+  readonly cache: string
+  readonly config: string
+  readonly state: string
+  readonly bin: string
```
Astro-Han, line 47: **P0**: `xdgData!`, `xdgCache!`, `xdgConfig!`, `xdgState!` non-null assertions at module top-level. On Linux systems or containers where these XDG env vars are unset, each is `undefined`, and `path.join(undefined!, "pawwork")` silently produces `"undefined/pawwork"` with no error. The old code had the same `!` assertions inside an Effect Layer where a failure could propagate. Now the bad paths are baked into `Path.data` etc. at import time with no recovery. Guard with a fallback before the assertion (e.g., `xdgData ?? os.tmpdir()`).

packages/core/src/npm.ts
```diff
@@ -0,0 +1,267 @@
...
+      const name = (() => {
+        try {
+          return npa(pkg).name ?? pkg
+        } catch {
+          return pkg
+        }
+      })()
+
+      if (yield* afs.existsSafe(dir)) {
+        return resolveEntryPoint(name, path.join(dir, "node_modules", name))
```
Astro-Han: **P0**: `existsSafe(dir)` replaces `arborist.loadVirtual()` — corrupted packages not self-repaired. Old code called `loadVirtual()` first to verify the npm tree integrity; only if that failed or the package was absent did it call `reify()`. This meant a partially-installed or corrupted `node_modules` got repaired on the next `add()` call. The new code checks only directory existence — a corrupted install (e.g. interrupted `reify()`, deleted inner files) is treated as valid, and LSP servers / plugins will fail at spawn time with no self-repair path. Restore the `loadVirtual()` call before this early return.

Astro-Han:
## Disposition of GLM-5.1 review (rounds 1 + 2)

Pushed three follow-up commits to the branch:

- `4181b92a` `fix(npm): preserve PawWork carve-outs in core npm move` — restores `arborist.loadVirtual()` integrity check in `Npm.add()` and the Windows `__FAKE_PLATFORM__ = "linux"` workaround that the upstream HEAD `npm.ts` snapshot dropped.
- `221911060` `fix(lsp): re-emit InstallFailed Bus event around core Npm calls` — wraps `Npm.add` with an opencode-side `npmWhich` helper that publishes `LSP.Event.InstallFailed` before falling back to `undefined`. Updates the Bus event payload, openapi snapshot, SDK type, and toast listener to the new `{add, dir, error}` shape.
- `22140c174` `chore(core): drop dead remnants from the npm move` — deletes `packages/opencode/src/effect/oltp.ts` (stale `Observability` namespace), removes the `bin: ./bin/opencode` entry from `packages/core/package.json` (file doesn't exist; old `shared/package.json` had no bin), and drops the now-unused `npm-package-arg` dependency from core.

### Items addressed (6/27)

| # | Status |
|---|---|
| #2 / #PR-line-117 / #PR-line-124 — `loadVirtual()` integrity check | Fixed in `4181b92a` |
| #3 / #4 — LSP `InstallFailed` dead listener + schema drift | Fixed in `221911060` |
| #7 — Windows `__FAKE_PLATFORM__` workaround | Fixed in `4181b92a` |
| #18 / #21 — `oltp.ts` dead code with conflicting `Observability` namespace | Fixed in `22140c174` |
| #24 — wrong `bin` entry in `core/package.json` | Fixed in `22140c174` |

### Items pushed back as factually incorrect

- **#1** `xdgData!` produces `"undefined/pawwork"` silently — `path.join(undefined, "x")` throws `TypeError: ERR_INVALID_ARG_TYPE` in Bun and Node, not silent corruption. The old `packages/opencode/src/global/index.ts` evaluated `xdgData!` at module top in identical structure, not inside an Effect Layer.
- **#5** `serviceName: "opencode"` is a regression — old `packages/opencode/src/effect/observability.ts` also hardcoded `serviceName: "opencode"` at the same line. Pre-existing issue, not introduced by the move.
- **#6** dual Global module TOCTOU race — new `core/global.ts` does not contain a cache-version-check sequence (only paths + `mkdir` + flock setup). The version check lives only in `packages/opencode/src/global/index.ts`. No race.
- **#12** `Bun` global type lost from removed `types.d.ts` — `packages/core/package.json` lists `@types/bun: catalog:` and `tsconfig.json` extends `@tsconfig/bun/tsconfig.json`. Bun globals are typed.
- **#13** `which()` behavior change — flagged as "actually better" by the same review. Accepting upstream's improvement.

### Items pushed back as out-of-scope mechanical move

These reflect upstream HEAD behaviors that ship verbatim with the move. They may warrant their own focused PRs but don't belong in a "move modules into core" commit.

- **#8** `NpmConfig.load()` silent error swallow
- **#10** `* as` vs `{ }` import style for `CrossSpawnSpawner` / `EffectLogger` (cosmetic; both forms work via `export * as`)
- **#11** `install()` `canWrite` early return masks permission errors
- **#14** ~17 opencode files still import `Global` from local `../global` (PR description documents the local Global is intentionally retained as PawWork's Runtime-aware module)
- **#15 / #22** `mergeWithExisting: false` and `loggerExportInterval` removal
- **#16** `setGlobalContextManager` singleton mutation (test-environment theory; Bun runs each test file in its own process)
- **#17 / #25** `ensureProcessMetadata("main")` hardcoded — `ensureProcessMetadata` is a net-new upstream addition; the old observability had no equivalent call. Not a regression.
- **#19** `sanitize()` no path traversal guard — pre-existing upstream behavior; only triggerable via self-authored plugin/provider config (low value attack vector). Worth a follow-up issue if defense-in-depth is desired.
- **#20** `core/util/log.ts` triggers `core/global.ts` side effects on import — same module-top side-effect pattern the old `opencode/util/log.ts` already had via opencode's `Global` import.
- **#23** `resource()` not memoized — runs once per runtime construction; micro-optimization.
- **#26** `@opentelemetry/context-async-hooks` direct dep vs catalog — works as a transitive peer.
- **#27** `export * as Global from "./global"` self-referential pattern — established upstream convention (see `log.ts`, `npm.ts`).

### Verification

```
bun turbo typecheck   # 8 packages, all pass
bun --cwd packages/core test test/npm.test.ts            # 3 pass
bun --cwd packages/opencode test test/lsp/server.test.ts # 2 pass
```

CI will revalidate the full matrix.

packages/core/src/effect/observability.ts
Astro-Han: Out-of-scope for this mechanical move PR. Upstream HEAD's OtlpLogger.make signature dropped loggerExportInterval; restoring it would be a behavior tweak, not a move-completion. If log batching latency proves user-visible, worth a separate PR.

packages/core/src/effect/observability.ts
Astro-Han: Verified against origin/dev: old packages/opencode/src/effect/observability.ts had no ensureProcessMetadata call at all. This is a net-new upstream addition, not a regression introduced by the move. The 'main' fallback is only consulted when OPENCODE_PROCESS_ROLE is unset.

packages/core/src/npm.ts
Astro-Han: Upstream HEAD behavior. The early return is meant to make read-only project dirs (think /workspace mounts) non-fatal. If silent skip turns out to mask real install failures in PawWork, worth a separate PR — out-of-scope here.

packages/core/src/npm.ts
Astro-Han: Addressed in 221911060: opencode-side npmWhich helper now reads err.add/err.dir and republishes LSP.Event.InstallFailed with the new {add, dir, error} payload. openapi snapshot, SDK type, and toast listener updated to match.

packages/core/src/effect/observability.ts
Astro-Han: Verified against origin/dev: old packages/opencode/src/effect/observability.ts hardcoded serviceName: 'opencode' on the same line. Pre-existing PawWork attribution issue, not a regression introduced by the move. Worth its own carve-out PR if Runtime.appName() attribution matters.

packages/core/src/npm.ts
Astro-Han: Addressed in 4181b92a: existsSafe(dir) shortcut now delegates to a loadVirtual integrity check; corrupted/partial installs fall through to a fresh reify.

packages/core/src/effect/observability.ts
Astro-Han: Out-of-scope for this mechanical move PR. The flip from loggerMergeWithExisting:true → mergeWithExisting:false is the upstream HEAD behavior; no current PawWork consumer adds custom loggers to the Effect runtime independently, so the runtime impact is null. Tracking via separate issue if a use case appears.

packages/core/src/npm.ts
Astro-Han: Pre-existing upstream behavior; macOS/Linux sanitize() is identity by design (the illegal-char set only exists for Windows file naming). Only triggerable via self-authored plugin/provider config — low value attack vector. Worth a follow-up issue if defense-in-depth is desired.

packages/core/src/npm-config.ts
Astro-Han: Out-of-scope mechanical move; this is the upstream HEAD npm-config.ts implementation. Same finding from the gemini review pass earlier. If 401-from-bad-.npmrc-config diagnostics matter to PawWork, worth a separate PR.

packages/core/src/npm.ts
Astro-Han: Addressed in 4181b92a: a loadArborist() helper sets __FAKE_PLATFORM__ = 'linux' on Windows before the first @npmcli/arborist import (which transitively loads tar). Both reify() and the new loadVirtual() integrity check route through this helper.

packages/core/src/global.ts
Astro-Han: Verified two sub-claims as factually incorrect: (a) old packages/opencode/src/global/index.ts evaluated xdgData!/xdgCache!/xdgConfig!/xdgState! at module top in identical structure — not inside an Effect Layer; (b) path.join(undefined, 'x') throws TypeError: ERR_INVALID_ARG_TYPE in both Bun and Node, not silent 'undefined/pawwork'. xdg-basedir@5.1.0 only returns undefined when os.homedir() is empty, which is extremely rare on macOS/Windows. Side-effect-on-import (mkdir + Flock.setGlobal) is the established upstream pattern that the old global already had — out-of-scope to refactor here.

packages/core/src/npm.ts
Astro-Han: Same fix as the loadVirtual thread above — addressed in 4181b92a.

packages/core/package.json
```diff
@@ -1,28 +1,45 @@
...
   "devDependencies": {
     "@tsconfig/bun": "catalog:",
-    "@types/bun": "catalog:"
+    "@types/bun": "catalog:",
+    "@types/cross-spawn": "catalog:",
+    "@types/npmcli__arborist": "6.3.3",
+    "@types/semver": "catalog:"
   },
   "dependencies": {
+    "@effect/opentelemetry": "catalog:",
```
Astro-Han, line 25: [P1] This new dependency path is missing a runtime peer. `observability.ts` imports `@effect/opentelemetry/NodeSdk`, and the desktop smoke build now fails because that module imports `@opentelemetry/sdk-trace-node`, while core only declares `@opentelemetry/sdk-trace-base`. Please add the missing dependency or avoid the NodeSdk import path.

packages/core/src/util/log.ts
```diff
@@ -1,7 +1,7 @@
 import path from "path"
 import fs from "fs/promises"
 import { createWriteStream } from "fs"
-import { Global } from "../global"
+import * as Global from "../global"
```
Astro-Han, line 4: [P1] Moving this module breaks an existing import that was not updated. `packages/app/script/e2e-local.ts` still imports `../../opencode/src/util/log`, and the e2e smoke job now fails with `Cannot find module '../../opencode/src/util/log'`. Please update that script to import the moved core module.

packages/core/src/effect/observability.ts
```diff
@@ -0,0 +1,107 @@
...
+          return [decodeURIComponent(entry.slice(0, index)), decodeURIComponent(entry.slice(index + 1))]
+        }),
+      )
+    } catch {
+      return {}
+    }
+  })()
+
+  return {
+    serviceName: "opencode",
```
Astro-Han: [P2] This keeps PawWork telemetry attributed to `opencode`. The PR already carves out `global.ts` to use `Runtime.appName()` so PawWork keeps its own namespace; the telemetry resource should follow the same boundary, or the exception should be documented as intentional.

Astro-Han (commented):
Review comments focused on current CI blocking regressions and PawWork boundary behavior.

packages/core/package.json
Astro-Han: Fixed in 640143c8: added @opentelemetry/sdk-trace-node 2.6.1 to packages/core/package.json. Verified locally: bun --cwd packages/opencode run build:embedded-server now resolves NodeSdk and exits with 'Build complete'.

packages/core/src/util/log.ts
Astro-Han: Fixed in d0fbcf30: lazy import in packages/app/script/e2e-local.ts now points at @opencode-ai/core/util/log instead of the moved-away ../../opencode/src/util/log.

packages/core/src/effect/observability.ts
Astro-Han: Reversed my earlier push back — the global.ts carve-out already breaks the 'old code hardcoded opencode' precedent in this PR, so observability should follow the same boundary. Fixed in 640143c8: serviceName now uses Runtime.appName(), so PawWork OTel traces report as 'pawwork' consistently with the data dir naming.

Astro-Han (commented):
## Review Summary

This PR moves Global, log, flag, cross-spawn-spawner, npm, and observability into `@opencode-ai/core`. The mechanical import rewrites look correct, and the carve-outs for PawWork branding are well-documented. Below are inline-style comments organized by severity.

---

### [P0] `packages/core/src/global.ts` — top-level `await` in module scope

The new `global.ts` performs top-level `await Promise.all([fs.mkdir(...)])` and calls `Flock.setGlobal({ state })` at import time. This is a significant behavior change from the previous `Layer.effect`-based lazy initialization.

- **Boundary case**: In test environments where `OPENCODE_TEST_HOME` is set *after* the module is first imported (e.g., via `vitest`/`bun test` preload patterns), the paths will already be computed against the real home directory. The previous `Layer.effect` deferred this until `Effect.runSync` or similar.
- **Regression risk**: Any test that imports `@opencode-ai/core/global` transitively before setting `OPENCODE_TEST_HOME` will pollute the real `~/.config/pawwork/` or `~/.cache/pawwork/` directories.
- **Suggested fix**: Move the eager `mkdir` and `Flock.setGlobal` into a `Global.init()` function that callers invoke explicitly, or keep them inside the `Layer.effect` so initialization is lazy and test-controllable.

---

### [P0] `packages/core/src/npm.ts` — `Npm.which` swallows all errors silently

```ts
export async function which(...args) {
  const resolved = await runPromise((svc) => svc.which(...args))
  return Option.getOrUndefined(resolved)
}
```

The upstream `which()` implementation returns `Option.none()` on *any* failure (including `InstallFailedError`, `EffectFlock.LockError`, file system errors, etc.). The PR description acknowledges this and adds `npmWhich()` in `lsp/server.ts` to re-emit `InstallFailedError`.

- **Boundary case**: Non-LSP callers of `Npm.which` (e.g., `Formatter` in `packages/opencode/src/format/formatter.ts`, `Provider` in `provider/provider.ts`) now lose all failure signals. Previously, `Npm.which` in the old `packages/opencode/src/npm/index.ts` would throw on `add()` failure, letting callers handle it.
- **Regression risk**: Formatter or provider resolution may silently fail without any log or user feedback. The `npmWhich` wrapper only covers LSP servers.
- **Suggested fix**: Either restore error propagation in `Npm.which`, or audit all non-LSP callers to ensure silent failure is acceptable.

---

### [P1] `packages/core/src/npm.ts` — `loadArborist` mutates `process.env` globally

```ts
if (process.platform === "win32") {
  process.env.__FAKE_PLATFORM__ = "linux"
}
```

This is set at module load time inside `loadArborist()`, which is called lazily. However, once set, it persists for the entire process.

- **Boundary case**: If PawWork later needs to run *actual* tar operations on Windows (e.g., packaging, extracting user archives), `__FAKE_PLATFORM__=linux` may cause incorrect behavior in those tar instances.
- **Regression risk**: Global env mutation is hard to trace. If another dependency also uses tar and expects Windows behavior, it will be silently broken.
- **Suggested fix**: Consider scoping this to the Arborist child process only, or documenting it prominently as a known process-wide side effect.

---

### [P1] `packages/core/src/effect/observability.ts` — `serviceName: Runtime.appName()` but `opencode.client` attribute remains

The carve-out changes `serviceName` from hardcoded `"opencode"` to `Runtime.appName()` (which returns `"pawwork"`). However, the OTel resource attributes still include `"opencode.client"`, `"opencode.process_role"`, `"opencode.run_id"`.

- **Boundary case**: If PawWork ever sends telemetry to an upstream opencode collector, the `serviceName` will say "pawwork" but the attributes will still say "opencode". This is inconsistent branding.
- **Regression risk**: Low for functionality, but medium for telemetry attribution confusion.
- **Suggested fix**: Either rename the attribute keys to `pawwork.*` or add a comment explaining why they remain `opencode.*` (e.g., upstream schema compatibility).

---

### [P1] `packages/core/src/npm.ts` — `InstallFailedError` schema change breaks external consumers

The error shape changed from `{ pkg: string }` to `{ add: string[], dir: string, cause: Schema.Defect }`. The PR updates the SDK types and OpenAPI snapshot, which is good.

- **Boundary case**: Any external SDK consumer (third-party integrations, scripts) that parsed `properties.pkg` will break. The OpenAPI spec change is a breaking API change.
- **Regression risk**: If there are downstream consumers outside this repo (e.g., cloud control plane, mobile app), they need coordinated updates.
- **Suggested fix**: Add a note to the PR description or a migration doc about the breaking SDK change. Consider whether `pkg` should be kept as a deprecated alias for backward compatibility.

---

### [P2] `packages/core/src/global.ts` — `xdg-basedir` modules may return `undefined`

```ts
const data = path.join(xdgData!, app)
```

The code uses non-null assertion (`!`) on `xdgData`, `xdgCache`, etc. In exotic environments (Windows without WSL, containers without XDG vars, certain CI images), these can return `undefined`.

- **Boundary case**: On Windows or minimal Docker containers, `xdg-basedir` may return `undefined` for some paths, causing `path.join(undefined, "pawwork")` to throw.
- **Regression risk**: Crash on startup in unsupported environments. The old code had the same issue, so this is not a new regression, but the move to core makes it harder to patch locally.
- **Suggested fix**: Add fallback values (e.g., `path.join(os.homedir(), ".local", "share")`) or at least a runtime check with a clear error message.

---

### [P2] `packages/core/src/npm.ts` — `resolveEntryPoint` uses `import.meta.resolve` with Bun-only guard

```ts
const resolved = typeof Bun !== "undefined" ? import.meta.resolve(name, dir) : import.meta.resolve(dir)
```

- **Boundary case**: In Node.js (non-Bun) environments, `import.meta.resolve(dir)` resolves the directory itself, not the package `name`. This is likely incorrect — it should resolve `name` with `dir` as the parent.
- **Regression risk**: If PawWork ever runs on Node.js (e.g., for debugging, testing, or future portability), `resolveEntryPoint` will return wrong paths.
- **Suggested fix**: Use `require.resolve(name, { paths: [dir] })` for Node.js, or document that this module is Bun-only.

---

### [P2] `packages/core/src/npm-config.ts` — imports from `@npmcli/config/lib/definitions/index.js`

```ts
import { definitions, flatten, nerfDarts, shorthands } from "@npmcli/config/lib/definitions/index.js"
```

This imports from a deep internal path that npm does not guarantee as stable API.

- **Boundary case**: A future npm patch release could move or rename this file, breaking the build.
- **Regression risk**: Build breakage on `bun install` when npm updates. The `// @ts-expect-error` comment acknowledges this, but it doesn't prevent runtime breakage.
- **Suggested fix**: Pin `@npmcli/config` to an exact version in `package.json` (already done: `10.8.1`), and add a CI check or dependabot rule to block automatic minor/patch bumps for this package.

---

### [P2] `packages/core/src/util/opencode-process.ts` — env var names still say `OPENCODE`

```ts
export const OPENCODE_RUN_ID = "OPENCODE_RUN_ID"
export const OPENCODE_PROCESS_ROLE = "OPENCODE_PROCESS_ROLE"
```

These are now in `@opencode-ai/core` but the env var names are still `OPENCODE_*`. This is inconsistent with the PawWork carve-out philosophy.

- **Boundary case**: If a user runs both opencode and pawwork on the same machine, these env vars will collide.
- **Regression risk**: Very low, but it’s a branding inconsistency that will need to be carved out in every future sync.
- **Suggested fix**: Add these to the permanent carve-out list and rename to `PAWWORK_RUN_ID` / `PAWWORK_PROCESS_ROLE`, or document why they must stay `OPENCODE_*`.

---

### [P2] `packages/core/test/npm.test.ts` — test uses `Bun.write` and `fs.stat` but no cleanup on failure

```ts
await Bun.write(path.join(tmp.path, ".npmrc"), "omit=dev\n")
```

The `tmpdir` helper uses `Symbol.asyncDispose`, which is good. But if `Bun.write` or `Npm.install` throws, the `using` block should still dispose. However, `Bun.write` is not wrapped in `try/finally`.

- **Boundary case**: Not a real issue because `using` handles disposal, but `Bun.write` may leave partial files if interrupted.
- **Regression risk**: Minimal — test-only.
- **Suggested fix**: None critical, but consider using `fs.writeFile` for consistency with the rest of the test suite (the cross-spawn test already switched from `Bun.write` to `fs.writeFile`).

---

### [P3] `packages/core/src/npm.ts` — `Npm.install` early-return on `!canWrite` without logging

```ts
if (!canWrite) return
```

If the target directory is not writable, `Npm.install` silently returns without installing anything and without logging.

- **Boundary case**: Read-only project directories (e.g., mounted volumes, protected workspaces).
- **Regression risk**: Users may wonder why dependencies are missing with no error.
- **Suggested fix**: Add a `log.warn` or similar before the early return.

---

### [P3] `packages/core/package.json` — `overrides: { "drizzle-orm": "catalog:" }` unexplained

The new `overrides` field in `packages/core/package.json` pins `drizzle-orm` to the catalog version. `core` does not appear to depend on `drizzle-orm` directly.

- **Boundary case**: This may be a transitive override to resolve a version conflict elsewhere in the workspace.
- **Regression risk**: If the override is unnecessary, it adds maintenance burden. If it is necessary, it should be documented.
- **Suggested fix**: Add a comment in the PR or a `package.json` comment explaining why this override exists.

---

### [P3] `packages/core/src/effect/observability.ts` — `AsyncLocalStorageContextManager` registered globally without cleanup

```ts
const mgr = new AsyncLocalStorageContextManager()
mgr.enable()
context.setGlobalContextManager(mgr)
```

This is called inside the `traces()` async function, which is invoked once when the observability layer is built. There is no `mgr.disable()` or cleanup on process exit.

- **Boundary case**: In test suites that create and tear down multiple runtimes, the global context manager may accumulate state or interfere with other tests.
- **Regression risk**: Test flakiness in observability-enabled test runs.
- **Suggested fix**: Add a cleanup hook or document that observability layer should not be rebuilt in tests.

---

### [P3] `packages/app/src/context/global-sync.tsx` — `add?.[0]` fallback may show empty string

```ts
pkg: properties?.add?.[0] ?? "",
```

If `add` is an empty array, `add?.[0]` is `undefined`, so the toast description will show an empty package name.

- **Boundary case**: `InstallFailedError` constructed with `add: []` (theoretically possible via the schema since `add` is optional).
- **Regression risk**: User sees "Failed to install " with no package name.
- **Suggested fix**: Use `properties?.add?.[0] ?? properties?.dir ?? "unknown"` for a better fallback.

---

### [P3] `packages/core/src/npm.ts` — `reify` uses `Effect.withSpan` but `add` is not sanitized in span attributes

```ts
Effect.withSpan("Npm.reify", { attributes: input })
```

If `input.add` contains a scoped package with a registry token (e.g., `@acme/pkg@npm:registry.example.com/pkg?authToken=secret`), the token will be emitted into the OTel span.

- **Boundary case**: Private registry URLs with embedded credentials.
- **Regression risk**: Credential leakage in telemetry.
- **Suggested fix**: Sanitize or redact `add` entries before putting them in span attributes.

---

### [P3] `packages/opencode/src/lsp/server.ts` — `npmWhich` catches `Npm.InstallFailedError` but re-throws other errors

```ts
if (err instanceof Npm.InstallFailedError) { ... }
throw err
```

If `Npm.add` throws a non-`InstallFailedError` (e.g., `EffectFlock.LockError`, unexpected runtime error), it bubbles up and may crash the LSP server initialization.

- **Boundary case**: File system lock contention or disk-full errors.
- **Regression risk**: LSP server crash instead of graceful degradation.
- **Suggested fix**: Consider catching all errors in `npmWhich`, logging them, and returning `undefined` to match the "which returns undefined on failure" contract.

Astro-Han (commented):
## Review Summary

This PR moves Global, log, flag, cross-spawn-spawner, npm, and observability into `@opencode-ai/core`. The mechanical import rewrites look correct, and the carve-outs for PawWork branding are well-documented. Below are inline-style comments organized by severity.

---

### [P0] `packages/core/src/global.ts` — top-level `await` in module scope

The new `global.ts` performs top-level `await Promise.all([fs.mkdir(...)])` and calls `Flock.setGlobal({ state })` at import time. This is a significant behavior change from the previous `Layer.effect`-based lazy initialization.

- **Boundary case**: In test environments where `OPENCODE_TEST_HOME` is set *after* the module is first imported (e.g., via `vitest`/`bun test` preload patterns), the paths will already be computed against the real home directory. The previous `Layer.effect` deferred this until `Effect.runSync` or similar.
- **Regression risk**: Any test that imports `@opencode-ai/core/global` transitively before setting `OPENCODE_TEST_HOME` will pollute the real `~/.config/pawwork/` or `~/.cache/pawwork/` directories.
- **Suggested fix**: Move the eager `mkdir` and `Flock.setGlobal` into a `Global.init()` function that callers invoke explicitly, or keep them inside the `Layer.effect` so initialization is lazy and test-controllable.

---

### [P0] `packages/core/src/npm.ts` — `Npm.which` swallows all errors silently

```ts
export async function which(...args) {
  const resolved = await runPromise((svc) => svc.which(...args))
  return Option.getOrUndefined(resolved)
}
```

The upstream `which()` implementation returns `Option.none()` on *any* failure (including `InstallFailedError`, `EffectFlock.LockError`, file system errors, etc.). The PR description acknowledges this and adds `npmWhich()` in `lsp/server.ts` to re-emit `InstallFailedError`.

- **Boundary case**: Non-LSP callers of `Npm.which` (e.g., `Formatter` in `packages/opencode/src/format/formatter.ts`, `Provider` in `provider/provider.ts`) now lose all failure signals. Previously, `Npm.which` in the old `packages/opencode/src/npm/index.ts` would throw on `add()` failure, letting callers handle it.
- **Regression risk**: Formatter or provider resolution may silently fail without any log or user feedback. The `npmWhich` wrapper only covers LSP servers.
- **Suggested fix**: Either restore error propagation in `Npm.which`, or audit all non-LSP callers to ensure silent failure is acceptable.

---

### [P1] `packages/core/src/npm.ts` — `loadArborist` mutates `process.env` globally

```ts
if (process.platform === "win32") {
  process.env.__FAKE_PLATFORM__ = "linux"
}
```

This is set at module load time inside `loadArborist()`, which is called lazily. However, once set, it persists for the entire process.

- **Boundary case**: If PawWork later needs to run *actual* tar operations on Windows (e.g., packaging, extracting user archives), `__FAKE_PLATFORM__=linux` may cause incorrect behavior in those tar instances.
- **Regression risk**: Global env mutation is hard to trace. If another dependency also uses tar and expects Windows behavior, it will be silently broken.
- **Suggested fix**: Consider scoping this to the Arborist child process only, or documenting it prominently as a known process-wide side effect.

---

### [P1] `packages/core/src/effect/observability.ts` — `serviceName: Runtime.appName()` but `opencode.client` attribute remains

The carve-out changes `serviceName` from hardcoded `"opencode"` to `Runtime.appName()` (which returns `"pawwork"`). However, the OTel resource attributes still include `"opencode.client"`, `"opencode.process_role"`, `"opencode.run_id"`.

- **Boundary case**: If PawWork ever sends telemetry to an upstream opencode collector, the `serviceName` will say "pawwork" but the attributes will still say "opencode". This is inconsistent branding.
- **Regression risk**: Low for functionality, but medium for telemetry attribution confusion.
- **Suggested fix**: Either rename the attribute keys to `pawwork.*` or add a comment explaining why they remain `opencode.*` (e.g., upstream schema compatibility).

---

### [P1] `packages/core/src/npm.ts` — `InstallFailedError` schema change breaks external consumers

The error shape changed from `{ pkg: string }` to `{ add: string[], dir: string, cause: Schema.Defect }`. The PR updates the SDK types and OpenAPI snapshot, which is good.

- **Boundary case**: Any external SDK consumer (third-party integrations, scripts) that parsed `properties.pkg` will break. The OpenAPI spec change is a breaking API change.
- **Regression risk**: If there are downstream consumers outside this repo (e.g., cloud control plane, mobile app), they need coordinated updates.
- **Suggested fix**: Add a note to the PR description or a migration doc about the breaking SDK change. Consider whether `pkg` should be kept as a deprecated alias for backward compatibility.

---

### [P2] `packages/core/src/global.ts` — `xdg-basedir` modules may return `undefined`

```ts
const data = path.join(xdgData!, app)
```

The code uses non-null assertion (`!`) on `xdgData`, `xdgCache`, etc. In exotic environments (Windows without WSL, containers without XDG vars, certain CI images), these can return `undefined`.

- **Boundary case**: On Windows or minimal Docker containers, `xdg-basedir` may return `undefined` for some paths, causing `path.join(undefined, "pawwork")` to throw.
- **Regression risk**: Crash on startup in unsupported environments. The old code had the same issue, so this is not a new regression, but the move to core makes it harder to patch locally.
- **Suggested fix**: Add fallback values (e.g., `path.join(os.homedir(), ".local", "share")`) or at least a runtime check with a clear error message.

---

### [P2] `packages/core/src/npm.ts` — `resolveEntryPoint` uses `import.meta.resolve` with Bun-only guard

```ts
const resolved = typeof Bun !== "undefined" ? import.meta.resolve(name, dir) : import.meta.resolve(dir)
```

- **Boundary case**: In Node.js (non-Bun) environments, `import.meta.resolve(dir)` resolves the directory itself, not the package `name`. This is likely incorrect — it should resolve `name` with `dir` as the parent.
- **Regression risk**: If PawWork ever runs on Node.js (e.g., for debugging, testing, or future portability), `resolveEntryPoint` will return wrong paths.
- **Suggested fix**: Use `require.resolve(name, { paths: [dir] })` for Node.js, or document that this module is Bun-only.

---

### [P2] `packages/core/src/npm-config.ts` — imports from `@npmcli/config/lib/definitions/index.js`

```ts
import { definitions, flatten, nerfDarts, shorthands } from "@npmcli/config/lib/definitions/index.js"
```

This imports from a deep internal path that npm does not guarantee as stable API.

- **Boundary case**: A future npm patch release could move or rename this file, breaking the build.
- **Regression risk**: Build breakage on `bun install` when npm updates. The `// @ts-expect-error` comment acknowledges this, but it doesn't prevent runtime breakage.
- **Suggested fix**: Pin `@npmcli/config` to an exact version in `package.json` (already done: `10.8.1`), and add a CI check or dependabot rule to block automatic minor/patch bumps for this package.

---

### [P2] `packages/core/src/util/opencode-process.ts` — env var names still say `OPENCODE`

```ts
export const OPENCODE_RUN_ID = "OPENCODE_RUN_ID"
export const OPENCODE_PROCESS_ROLE = "OPENCODE_PROCESS_ROLE"
```

These are now in `@opencode-ai/core` but the env var names are still `OPENCODE_*`. This is inconsistent with the PawWork carve-out philosophy.

- **Boundary case**: If a user runs both opencode and pawwork on the same machine, these env vars will collide.
- **Regression risk**: Very low, but it's a branding inconsistency that will need to be carved out in every future sync.
- **Suggested fix**: Add these to the permanent carve-out list and rename to `PAWWORK_RUN_ID` / `PAWWORK_PROCESS_ROLE`, or document why they must stay `OPENCODE_*`.

---

### [P2] `packages/core/test/npm.test.ts` — test uses `Bun.write` and `fs.stat` but no cleanup on failure

```ts
await Bun.write(path.join(tmp.path, ".npmrc"), "omit=dev\n")
```

The `tmpdir` helper uses `Symbol.asyncDispose`, which is good. But if `Bun.write` or `Npm.install` throws, the `using` block should still dispose. However, `Bun.write` is not wrapped in `try/finally`.

- **Boundary case**: Not a real issue because `using` handles disposal, but `Bun.write` may leave partial files if interrupted.
- **Regression risk**: Minimal — test-only.
- **Suggested fix**: None critical, but consider using `fs.writeFile` for consistency with the rest of the test suite (the cross-spawn test already switched from `Bun.write` to `fs.writeFile`).

---

### [P3] `packages/core/src/npm.ts` — `Npm.install` early-return on `!canWrite` without logging

```ts
if (!canWrite) return
```

If the target directory is not writable, `Npm.install` silently returns without installing anything and without logging.

- **Boundary case**: Read-only project directories (e.g., mounted volumes, protected workspaces).
- **Regression risk**: Users may wonder why dependencies are missing with no error.
- **Suggested fix**: Add a `log.warn` or similar before the early return.

---

### [P3] `packages/core/package.json` — `overrides: { "drizzle-orm": "catalog:" }` unexplained

The new `overrides` field in `packages/core/package.json` pins `drizzle-orm` to the catalog version. `core` does not appear to depend on `drizzle-orm` directly.

- **Boundary case**: This may be a transitive override to resolve a version conflict elsewhere in the workspace.
- **Regression risk**: If the override is unnecessary, it adds maintenance burden. If it is necessary, it should be documented.
- **Suggested fix**: Add a comment in the PR or a `package.json` comment explaining why this override exists.

---

### [P3] `packages/core/src/effect/observability.ts` — `AsyncLocalStorageContextManager` registered globally without cleanup

```ts
const mgr = new AsyncLocalStorageContextManager()
mgr.enable()
context.setGlobalContextManager(mgr)
```

This is called inside the `traces()` async function, which is invoked once when the observability layer is built. There is no `mgr.disable()` or cleanup on process exit.

- **Boundary case**: In test suites that create and tear down multiple runtimes, the global context manager may accumulate state or interfere with other tests.
- **Regression risk**: Test flakiness in observability-enabled test runs.
- **Suggested fix**: Add a cleanup hook or document that observability layer should not be rebuilt in tests.

---

### [P3] `packages/app/src/context/global-sync.tsx` — `add?.[0]` fallback may show empty string

```ts
pkg: properties?.add?.[0] ?? "",
```

If `add` is an empty array, `add?.[0]` is `undefined`, so the toast description will show an empty package name.

- **Boundary case**: `InstallFailedError` constructed with `add: []` (theoretically possible via the schema since `add` is optional).
- **Regression risk**: User sees "Failed to install " with no package name.
- **Suggested fix**: Use `properties?.add?.[0] ?? properties?.dir ?? "unknown"` for a better fallback.

---

### [P3] `packages/core/src/npm.ts` — `reify` uses `Effect.withSpan` but `add` is not sanitized in span attributes

```ts
Effect.withSpan("Npm.reify", { attributes: input })
```

If `input.add` contains a scoped package with a registry token (e.g., `@acme/pkg@npm:registry.example.com/pkg?authToken=secret`), the token will be emitted into the OTel span.

- **Boundary case**: Private registry URLs with embedded credentials.
- **Regression risk**: Credential leakage in telemetry.
- **Suggested fix**: Sanitize or redact `add` entries before putting them in span attributes.

---

### [P3] `packages/opencode/src/lsp/server.ts` — `npmWhich` catches `Npm.InstallFailedError` but re-throws other errors

```ts
if (err instanceof Npm.InstallFailedError) { ... }
throw err
```

If `Npm.add` throws a non-`InstallFailedError` (e.g., `EffectFlock.LockError`, unexpected runtime error), it bubbles up and may crash the LSP server initialization.

- **Boundary case**: File system lock contention or disk-full errors.
- **Regression risk**: LSP server crash instead of graceful degradation.
- **Suggested fix**: Consider catching all errors in `npmWhich`, logging them, and returning `undefined` to match the "which returns undefined on failure" contract.

Astro-Han:
Cross-checked all 17 findings against pre-PR2 baseline (`aed455918:packages/opencode/src/global/index.ts` etc.) and current PR state. Pushing back on all of them — most rest on a wrong premise about pre-move behavior, a few are upstream patterns, and the rest are YAGNI for this mechanical-move PR.

| # | Verdict | Reason |
|---|---|---|
| P0-1 top-level await in global.ts | Push back | Pre-existing upstream pattern. `git show aed455918:packages/opencode/src/global/index.ts` already has `await Promise.all([fs.mkdir(...)])` at module scope. PR2 moved the file unchanged. The "previous `Layer.effect`-based lazy initialization" premise is wrong. PawWork's test setup uses `OPENCODE_TEST_HOME` which is set before process start (memory: `reference_bun_homedir_lock.md`). |
| P0-2 `Npm.which` swallows errors | Push back | Verified non-LSP callers: only `packages/opencode/src/format/formatter.ts` (3 callsites: prettier/oxfmt/@biomejs/biome). All use `const bin = await Npm.which(...); if (bin) return [bin, ...]` — formatter quietly disabled when bin missing, which is correct UX (best-effort feature). No `Npm.which` in `provider/provider.ts`. LSP keeps its wrapper. No regression. |
| P1-1 `__FAKE_PLATFORM__` global env mutation | Push back | Documented carve-out for `tar`'s `UV_FS_O_FILEMAP` issue on Bun+Windows. tar reads `__FAKE_PLATFORM__` once at module init, so persistence is required, not accidental. The comment in `loadArborist()` explains exactly this. |
| P1-2 `serviceName` vs `opencode.client` attribute inconsistency | Push back | Attribute keys are upstream's OTel schema, not introduced by this PR. The serviceName carve-out is targeted; broader attribute-key rename is a separate cleanup that needs its own decision (telemetry collector compatibility, dashboard queries). Not part of mechanical-move scope. |
| P1-3 InstallFailedError SDK breaking change | Already done | SDK is updated in this PR — `packages/sdk/openapi.json` and `packages/sdk/js/src/v2/gen/types.gen.ts` both modified. No external SDK consumers outside this repo. No additional code change needed. |
| P2-1 xdg-basedir `!` non-null assertion | Push back | Pre-existing in `aed455918:packages/opencode/src/global/index.ts:9-12`. Mechanical move preserved upstream pattern. Out of scope for PR2. |
| P2-2 `import.meta.resolve` Bun-only | Push back | Upstream pattern: PawWork ships as a Bun-compiled binary (`packages/opencode/script/build.ts`), so the Bun-only assumption is intentional. No Node.js runtime target. |
| P2-3 deep import from `@npmcli/config/lib/definitions` | Push back | Upstream pattern. Pin already in place (`10.8.1` exact). `// @ts-expect-error` comment documents the awareness. Adding a CI check for transitive npm bumps is out-of-scope tooling work. |
| P2-4 `OPENCODE_RUN_ID` env var names | Push back | Internal runtime metadata for OTel process correlation, not user-facing CLI env. No collision risk with the opencode binary because each process owns its own env. Renaming would affect telemetry attribution continuity across upstream syncs. |
| P2-5 test cleanup with `Bun.write` | Push back | Test-only. `using` handles disposal. Switching to `fs.writeFile` is a stylistic preference, not a fix. |
| P3-1 `!canWrite` silent return | Push back | Upstream pattern. Read-only project dirs is an intentional opt-out for Npm.install (cache reads still work). Adding warn log is upstream territory. |
| P3-2 `drizzle-orm` override | Push back | Upstream's, not ours. `git show 826a5f74b:packages/core/package.json` shows the override predates this PR. Not introduced here. |
| P3-3 `AsyncLocalStorageContextManager` no cleanup | Push back | PawWork carve-out (`packages/core/src/effect/observability.ts`) — global registration is intentional because the AI SDK's `tracer.startActiveSpan()` uses `context.active()` from the global API. The comment block above the registration explains this. Not test-relevant since observability is gated by `OTEL_EXPORTER_OTLP_ENDPOINT`. |
| P3-4 `add?.[0] ?? ""` fallback | Push back | YAGNI. `npmWhich` always passes `pkg` to `Array.from(err.add ?? [pkg])`, so `add[0]` is guaranteed non-empty. The `?? ""` is already defense-in-depth for an impossible path. |
| P3-5 `Npm.reify` span attributes leak secrets | Push back | Theoretical. `add` entries here are package names from `package.json` deps + LSP server bundles, not user-provided registry URLs with auth tokens. Authentication tokens live in `.npmrc` (handled by `NpmConfig.load`), not in package specifiers. |
| P3-6 `npmWhich` rethrows non-`InstallFailedError` | Push back | Intentional. Catching everything would swallow programmer errors and OOM/disk-full as "feature unavailable". `EffectFlock.LockError` is a recoverable known type — if it becomes a real LSP failure pattern, we'd add it explicitly to the catch, not blanket-catch. |

If any of these reads as defensiveness rather than reasoning, name the specific finding and I'll re-verify.

Astro-Han:
Re-审一遍后调整：P3-4 和 P3-6 纳入修复，已 squash 进 `53e5bb61d`（LSP wrapper commit）。其余 15 条仍 push back。

**已修：**

- **P3-6 npmWhich 兜底 LockError**：`Npm.add` 的错误通道含 `EffectFlock.LockError`（`LockTimeoutError | LockCompromisedError`），都是已知可恢复型。原来 blanket throw 会让 LSP bootstrap 因为 lock 竞争直接崩溃。改成显式 `instanceof EffectFlock.LockTimeoutError | LockCompromisedError → log.warn + return undefined`，与 which() "feature unavailable" 契约一致；其他未知错误仍然抛上去保留 debugging 能力。
- **P3-4 toast pkg fallback**：改成 `?? properties?.dir ?? "unknown"`，避免极端情况下 toast 显示残缺串。是 cosmetic 但 1 行成本。

**重新评估后仍 push back 的：**

- **P1-2 attribute key 一致性**：本来想顺手改，verify 后发现：`opencode.client` attribute 读的是 `Flag.OPENCODE_CLIENT` env var。改 attribute 不改 env 是误导，改 env 又会牵动 CLI flag、用户文档、`packages/core/test/effect/observability.test.ts` 等。完整 carve-out 范围远大于 attribute key 本身，需要独立 issue 决策（telemetry 收集器兼容、dashboard 查询语句迁移等）。serviceName 这一层先单独立住。

其余 14 条理由同上次评论。

PR #271, opened 2026-04-27
feat: enable default websearch

### LINKED ISSUE #132: [Feature] Define default web access and source safety strategy
We want PawWork agents to search the web when current or external information matters, without turning web access into a skill or making users manually enable a mode.

## What do you do today?
Web search and fetch exist as tools, but the harness direction does not yet define when models should search, how they should treat web content, or how to avoid prompt injection and low-quality sources.

Web search and fetch are default available tools. The harness guidance tells models to search when information may be stale, external, niche, price-related, news-related, legal, financial, medical, product-related, or otherwise likely to change. Retrieved pages are treated as untrusted input. The model prefers official, primary, recent, and source-attributed evidence, and it avoids following instructions found inside fetched pages.

Both

## Extra context
Users often complain that agents search too little, not too much. The strategy should encourage appropriate search while keeping source quality and prompt injection risks clear.

## Acceptance criteria
- Web search and fetch are part of the default tool surface, not a skill.
- The prompt or tool guidance defines when the model should search.
- Fetched page content is explicitly treated as untrusted input.
- The guidance warns against following instructions from web pages.
- Source priority favors official, primary, recent, and attributable sources.
- User-facing answers include citations or source links when web evidence materially affects the conclusion.

Astro-Han:
## Summary

- Expose WebSearch by default and keep WebFetch allowed by default while CodeSearch remains gated.
- Add saved Exa key support with precedence: saved key, EXA_API_KEY, bundled anonymous Exa MCP endpoint.
- Add General settings controls for Web Search, Exa key save/remove, and recovery toasts for quota or invalid-key failures.

## Why

PawWork should provide out-of-the-box web search. Users can start without configuring an API key, and only need to add EXA_API_KEY or a saved Exa key when the bundled quota is exhausted or their saved key needs attention.

## Related Issue

Closes #132

## How To Verify

```bash
bun install --frozen-lockfile
cd packages/opencode && bun test test/tool/registry.test.ts test/agent/agent.test.ts test/tool/mcp-exa.test.ts test/tool/websearch-auth.test.ts test/tool/websearch.test.ts
cd packages/opencode && bun run typecheck
cd packages/app && bun test --preload ./happydom.ts ./src/pages/session/websearch-toasts.test.ts ./src/pages/session/settings-websearch-source.test.ts
cd packages/app && bun run typecheck
cd packages/desktop-electron && bun test src/main/websearch-ipc-source.test.ts
cd packages/desktop-electron && bun run typecheck
cd packages/app && bun run test:e2e --grep "PawWork settings opens as a full-pane surface"
```

All commands above passed in this worktree.

## Screenshots or Recordings

No screenshot attached. The visible change is a General settings row and Exa API key control; the Settings surface was checked with the existing Playwright smoke e2e, and the new controls are covered by source contract tests.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Web Search: toggle in General settings, Manage Key dialog to add/remove Exa API keys, localized status chips, and guidance treating search results as untrusted.
  * Tools now appear or hide based on the Web Search setting.

* **Notifications**
  * Recovery toasts for quota-exceeded and invalid-key with quick Open Settings action; session-scoped deduplication.

* **Translations**
  * Complete English and Chinese copy for the Web Search UX.

* **Tests**
  * Expanded coverage for auth, tooling exposure, toasts, credential handling, and settings persistence.


packages/desktop-electron/src/main/ipc.ts
```diff
@@ -206,6 +206,43 @@ export function registerIpcHandlers(deps: Deps) {
...
+      ),
+    )
+    for (const [index, result] of results.entries()) {
+      if (result.status === "rejected") {
+        console.warn("websearch-set-enabled failed for instance", {
+          directory: directories[index],
+          error: result.reason,
+        })
+      }
+    }
```
Astro-Han, line 229: **P1.** Silent IPC failure swallows the rollback. `Promise.allSettled` plus a `console.warn` means this handler resolves successfully even when *every* per-instance `Instance.provide` rejects. The renderer's optimistic update in `packages/app/src/context/settings.tsx` only rolls back via `.catch(...)`, so the toggle reads as applied while no live project actually re-evaluated `tools()`. Either reject the IPC when all (or a configurable majority of) per-instance invalidations fail, or return a structured `{ok, failed}` payload and have the renderer roll back when `ok === 0`. The exact pattern exists in `lsp-set-enabled`, which is why doubling down deserves a fix here rather than copying it.

packages/opencode/src/settings/index.ts
```diff
@@ -5,17 +5,22 @@ export namespace Settings {
...
   }
 
   export class Service extends Context.Service<Service, Interface>()("@opencode/Settings") {}
 
   export const layer = Layer.effect(
     Service,
     Effect.gen(function* () {
-      const ref = yield* Ref.make<boolean>(false)
+      const lspEnabled = yield* Ref.make<boolean>(false)
+      const webSearchEnabled = yield* Ref.make<boolean>(true)
```
Astro-Han: **P1.** Toggle state is in-memory only. On Electron restart this defaults to `true` regardless of the user's last choice. The renderer's `createEffect` re-pushes the persisted store once `ready()` flips, but anything that reads `Settings.webSearchEnabled()` before that push (a session bootstrapped during initial agent boot, headless CLI flows, harness tests) sees `true`. Either persist this through the same Auth/JsonMigration mechanism that other settings use, or gate `WebSearchTool` resolution behind 'renderer push received' so the runtime never disagrees with the UI.

packages/opencode/src/tool/websearch-auth.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const status: Interface["status"] = Effect.fn("WebSearchAuth.status")(function* () {
+        return statusFrom(yield* getSaved())
+      })
+
+      const saveKey: Interface["saveKey"] = Effect.fn("WebSearchAuth.saveKey")(function* (key: string) {
+        const trimmed = key.trim()
+        if (!trimmed) return yield* Effect.fail(new Error("Exa API key is required"))
+        const credential: Credential = { source: "saved", key: trimmed }
+        yield* McpExa.validateCredential(http, credential, "8 seconds")
```
Astro-Han: **P2.** `validateCredential` bills the user's quota on every Save. This sends a real `web_search_exa` request (`query: "PawWork"`) against the very key the user just typed. A mistype plus retry already burns 2 quota slots before the feature is used in anger. If Exa exposes a cheaper validation/whoami endpoint, prefer that. Otherwise, skip eager validation and let the first real search surface `invalid_key` via the existing `markNeedsAttention` recovery path; that's already wired up and avoids the double-spend.

Astro-Han: **P2.** Network failure during validation blocks an otherwise valid key. A flaky link or a brief Exa outage propagates here as a generic `Error`, so the user cannot save a syntactically valid key at all. Distinguish `failure.kind === "network"` and either save with `metadata.status = "unverified"` (so subsequent searches still mark `needs_attention` if they fail), or surface a clearer toast that lets the user retry without retyping.

packages/opencode/src/tool/websearch-auth.ts
```diff
@@ -0,0 +1,122 @@
...
+      })
+
+      const status: Interface["status"] = Effect.fn("WebSearchAuth.status")(function* () {
+        return statusFrom(yield* getSaved())
+      })
+
+      const saveKey: Interface["saveKey"] = Effect.fn("WebSearchAuth.saveKey")(function* (key: string) {
+        const trimmed = key.trim()
+        if (!trimmed) return yield* Effect.fail(new Error("Exa API key is required"))
+        const credential: Credential = { source: "saved", key: trimmed }
```
Astro-Han: **P3.** Variable shadowing. Inner `const credential: Credential` shadows the outer `credential` Effect declared on line 55. It type-checks, but reading the function it's easy to think you're calling the cached credential builder. Rename the inner one (e.g. `submitted` or `nextCredential`) to keep intent clear.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -1,12 +1,84 @@
...
+
+export function endpoint(credential: Credential = credentialFromEnv()) {
+  if (credential.source === "anonymous") return BASE_URL
+  const url = new URL(BASE_URL)
+  url.searchParams.set("exaApiKey", credential.key)
+  return url.toString()
+}
+
+function statusFromBody(body: string) {
+  const match = body.match(/\((\d{3})\)|\bstatus\s*[:=]?\s*(\d{3})\b/i)
```
Astro-Han: **P2.** Regex too eager. `/\((\d{3})\)|\bstatus\s*[:=]?\s*(\d{3})\b/i` matches any parenthesized 3-digit number in the body. Real SSE error payloads occasionally include phrases like `(404 errors)` or version strings (`v3.200`), and would mis-classify into `quota_exceeded` / `invalid_key`. Anchor the alternation to error-shaped contexts (e.g. `/\berror\s*\(\s*(\d{3})\s*\)/i`) and only return a value when the surrounding text actually looks like an HTTP status.

packages/app/src/pages/session/message-timeline.tsx
```diff
@@ -247,6 +250,28 @@ export function MessageTimeline(props: {
     if (!id) return emptyMessages
     return sync.data.message[id] ?? emptyMessages
   })
+  const webSearchToastSeen = new Map<string, number>()
```
Astro-Han: **P2.** `webSearchToastSeen` grows unbounded. Entries are inserted on every classified failure but never evicted, even after `WEBSEARCH_TOAST_TTL_MS` has passed. A long session with repeated quota errors accumulates dead call-IDs that will never be revisited, so this is a per-tab leak proportional to error count. Sweep expired entries when adding a new one, or store `(id, expiresAt)` and prune at the top of the effect.

packages/app/src/pages/session/message-timeline.tsx
```diff
@@ -247,6 +250,28 @@ export function MessageTimeline(props: {
...
+          variant: "error",
+          actions: [
+            {
+              label: language.t(toast.actionKey),
+              onClick: () => shellSurface.openSettings(),
+            },
+          ],
+        })
+      }
+    }
```
Astro-Han, line 299: **P2.** O(N x M) iteration per sync update. This effect re-walks every message and every part on each `sync.data.part` change just to find the freshly arrived error part. For a 200-message session that updates dozens of times per minute, most of the work is wasted (dedup-discarded) traversal. Subscribe to a narrower channel that emits failed parts, or track a watermark `(messageID, partID)` so each sync tick only touches new parts.

packages/app/src/components/settings-general.tsx
```diff
@@ -184,6 +188,82 @@ export const SettingsGeneral: Component = () => {
...
   }
 
+  const [webSearchStatusResource, webSearchStatusActions] = createResource(() => window.api?.webSearchStatus?.())
+  const webSearchStatus = createMemo(
+    () =>
+      webSearchStatusResource.latest ?? {
+        source: "anonymous" as const,
+        configured: false,
+        needsAttention: false,
+      },
```
Astro-Han: **P3.** Wrong-status flash on load. `webSearchStatusResource.latest ?? { source: "anonymous", ... }` makes the description read 'Using PawWork's bundled Web Search quota' for the first paint even when a saved key exists, until the resource resolves. Either wrap the status line in `<Show when={webSearchStatusResource.latest}>` or render a small placeholder so the user never sees a transiently wrong source.

packages/app/src/components/settings-general.tsx
```diff
@@ -263,6 +343,69 @@ export const SettingsGeneral: Component = () => {
...
+              label={language.t("settings.general.webSearch.apiKey.title")}
+              hideLabel
+              type="password"
+              placeholder={language.t("settings.general.webSearch.apiKey.placeholder")}
+              autocomplete="off"
+              autocorrect="off"
+              autocapitalize="off"
+              spellcheck={false}
+              validationState={store.exaKeyError ? "invalid" : undefined}
+              error={store.exaKeyError}
```
Astro-Han: **P3.** Stale inline error. `store.exaKeyError` is set on Save failure but is only cleared on the next Save click; while the user types a new key the field stays visibly invalid. Add an `onInput` handler that clears `exaKeyError` (and probably re-enables a `Save` signal) so the form behaves like a normally validated input.

Astro-Han (commented):
Review of #271 (default WebSearch). Surface lines up with the LSP toggle pattern; copy, i18n, and sanitization tests look good. Inline notes below, graded P1 to P3. The two P1s: silent partial failure in `websearch-set-enabled` (renderer rollback never fires) and `Settings.webSearchEnabled` only living in an in-memory `Ref` (lost on restart and racy with first-paint reads).

packages/app/src/components/settings-general.tsx
```diff
@@ -263,6 +343,69 @@ export const SettingsGeneral: Component = () => {
...
+                  disabled={store.removingExaKey}
+                  onClick={removeExaApiKey}
+                >
+                  {language.t("settings.general.webSearch.apiKey.remove")}
+                </Button>
+              </Show>
+              <Button
+                size="small"
+                variant="secondary"
+                disabled={store.savingExaKey || !window.api?.saveExaApiKey}
```
Astro-Han: **P3.** Save button enabled with empty input. `disabled={store.savingExaKey || !window.api?.saveExaApiKey}` lets the user click Save while the field is empty; the click then raises the inline `required` error. Either disable while the trimmed input value is empty (track it via a small signal so reactivity hooks in) or treat the empty case as a silent no-op.

packages/desktop-electron/src/preload/types.ts
```diff
@@ -22,6 +22,12 @@ export type WindowConfig = {
 
 export type LinuxDisplayBackend = "wayland" | "auto"
 
+export type WebSearchStatus = {
+  source: "saved" | "env" | "anonymous"
+  configured: boolean
+  needsAttention: boolean
```
Astro-Han, line 28: **P3.** `WebSearchStatus` defined four times. Same shape now lives here, in `app.tsx`, in `env.d.ts`, and as `WebSearchAuth.Status`. Adding a field later is a four-place edit waiting to drift. Pick one source (probably `WebSearchAuth.Status` since it is the originator) and re-export through the preload/types boundary so the renderer pulls a single definition.

packages/desktop-electron/src/main/ipc.ts
```diff
@@ -206,6 +206,43 @@ export function registerIpcHandlers(deps: Deps) {
...
 
+  ipcMain.handle("websearch-set-enabled", async (_event: IpcMainInvokeEvent, value: boolean) => {
+    const { Settings, ToolRegistry, Instance } = await import("virtual:opencode-server")
+    await Settings.setWebSearchEnabled(value)
+    const directories = Instance.directories()
+    const results = await Promise.allSettled(
+      directories.map((directory) =>
+        Instance.provide({
+          directory,
+          fn: () => ToolRegistry.invalidate(),
```
Astro-Han, line 218: **P1** (resilience gap): `websearch-set-enabled` uses `Promise.allSettled` but only logs failures with `console.warn`. The IPC call returns `void` regardless, so the renderer never knows if some project instances failed to invalidate their tool registries. A stale instance with websearch still enabled could execute searches after the toggle is off. Consider returning an aggregated status or at minimum a success/failure boolean so the renderer can surface a warning toast.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -1,12 +1,84 @@
...
+  const value = match?.[1] ?? match?.[2]
+  return value ? Number(value) : undefined
+}
```
Astro-Han, line 53: **P1** (misclassification risk): `statusFromBody` regex `body.match(/\((\d{3})\)|\bstatus\s*[:=]?\s*(\d{3})\b/i)` is broad enough to match status codes in search result text content (e.g., "document status 404 was returned"). If `input.status` is absent (undefined), a false match could route an error to the wrong `classifyFailure` branch. Tighten the regex to match only at the start of a line or after MCP protocol markers like `"error"`, or use `input.status` as the primary signal with the body regex as a fallback only for `isError` SSE responses.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -18,11 +90,16 @@ const McpResult = Schema.Struct({
 
 const decode = Schema.decodeUnknownEffect(Schema.fromJsonString(McpResult))
 
-const parseSse = Effect.fn("McpExa.parseSse")(function* (body: string) {
+const parseSse = Effect.fn("McpExa.parseSse")(function* (body: string, source: Credential["source"]) {
   for (const line of body.split("\n")) {
     if (!line.startsWith("data: ")) continue
     const data = yield* decode(line.substring(6))
-    if (data.result.content[0]?.text) return data.result.content[0].text
```
Astro-Han, line 25: **P1** (silent pass-through): `parseSse` returns `undefined` when the SSE body contains no `data:` lines. This value flows back through `call()` as a successful result. In `validateCredential`, a 200 response with an empty or non-SSE body would pass validation and save the key without any verification. Consider returning `yield* Effect.fail(...)` at the end of `parseSse` if no data lines were processed, with a generic "no response from Exa" error.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -1,12 +1,84 @@
...
+  const match = body.match(/\((\d{3})\)|\bstatus\s*[:=]?\s*(\d{3})\b/i)
+  const value = match?.[1] ?? match?.[2]
+  return value ? Number(value) : undefined
+}
+
+function classifyFailure(input: { status?: number; body: string; source: Credential["source"] }): Failure {
+  const status = input.status ?? statusFromBody(input.body)
+  const text = input.body.toLowerCase()
+  if (status === 402 || status === 429 || /quota|rate.?limit|too many requests|usage limit/.test(text)) {
+    return { kind: "quota_exceeded", source: input.source, status }
```
Astro-Han, line 59: **P2** (classification granularity): `classifyFailure` checks `status === 402` for quota. HTTP 402 is Payment Required (rarely used in practice; standard rate limiting is 429). Verify that Exa actually returns 402 for quota exhaustion. If they only return 429, the 402 check is dead code. If they return 402 for billing failures, it should be a separate `billing` kind so the recovery message doesn't misleadingly say "search quota was reached" when the issue is payment.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -1,12 +1,84 @@
...
+    return { kind: "invalid_key", source: input.source, status }
+  }
+  return { kind: "unknown", source: input.source, status }
+}
+
+export function messageForFailure(failure: Failure) {
+  if (failure.kind === "invalid_key") {
+    if (failure.source === "saved") return "The saved Exa API key is invalid. Update or remove it in Settings."
+    if (failure.source === "env") return "The EXA_API_KEY environment variable is invalid. Update it and retry."
+    return "Exa rejected the anonymous Web Search request."
```
Astro-Han, line 71: **P2** (localization): `messageForFailure` returns hardcoded English strings displayed directly to users (e.g., "The saved Exa API key is invalid..."). The rest of the UI uses i18n keys. For consistency, these messages should use localization keys too, or at minimum document that backend-originated error strings are intentionally not localized.

packages/app/src/pages/session/websearch-toasts.ts
```diff
@@ -0,0 +1,67 @@
...
+  input: { now: number; seen: Map<string, number> },
+): WebSearchRecoveryToast | undefined {
+  const failure = failureFrom(part)
+  if (!failure) return
+  const callID = (part as { callID?: unknown }).callID
+  const id = typeof callID === "string" && callID ? callID : part.id
+  const previous = input.seen.get(id)
+  if (previous !== undefined && input.now - previous < WEBSEARCH_TOAST_TTL_MS) return
+
+  const kind = failure.kind
```
Astro-Han: **P2** (type safety): `const callID = (part as { callID?: unknown }).callID` casts the Part type to access `callID`, which isn't part of the public Part interface. If the SDK's Part type ever drops `callID` from tool parts, this silently falls back to `part.id` for deduplication. Add `callID` to the `failureFrom` guard or import the tool-part-specific type if available.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -68,11 +146,46 @@ export const call = <F extends Schema.Struct.Fields>(
...
+      return yield* Effect.fail(new McpExaError(failure, messageForFailure(failure)))
+    }
+    return yield* parseSse(body, credential.source)
   })
+
+export function validateCredential(http: HttpClient.HttpClient, credential: Credential, timeout: Duration.Input) {
+  return call(
+    http,
+    "web_search_exa",
+    SearchArgs,
```
Astro-Han: **P3**: `validateCredential` hardcodes `query: "PawWork"` as the test search. If the product name changes, this validation probe becomes outdated and could confuse Exa analytics. Use a neutral query like `"test"` or `"connection test"` instead of the product name.

packages/opencode/src/tool/websearch-auth.ts
```diff
@@ -0,0 +1,122 @@
...
+import { makeRuntime } from "../effect/run-service"
+import * as McpExa from "./mcp-exa"
+
+export namespace WebSearchAuth {
+  export const AUTH_KEY = "pawwork:websearch:exa"
+
+  export type Credential = McpExa.Credential
+  export type Status = {
+    source: Credential["source"]
+    configured: boolean
```
Astro-Han, line 16: **P3**: `AUTH_KEY = "pawwork:websearch:exa"` — product name in credential namespace. This is fine as it's an internal storage key, but all such internal namespacing in the codebase should follow the same convention so a future rename is a single find-and-replace.

Astro-Han (commented):
Thorough review of the websearch default-enable feature. No confirmed P0. Several P1 findings worth attention before merge.

**Security review**: key never leaks to renderer (verified by tests). `exaApiKey` in query string is pre-existing and follows the Exa MCP API contract. `type="password"` on the API key input field. Good.

**Test coverage**: strong. Unit tests for McpExa (credential routing, isError, unknown), WebSearchAuth (validate-then-save, key precedence, removal), websearch tool (failure metadata), toasts (dedupe, TTL, network exclusion), registry visibility toggle, IPC contract test, and source-level assertion tests for settings UI surface. Integration tests wire up all layers including prompt-effect and snapshot-tool-race.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -68,11 +146,46 @@ export const call = <F extends Schema.Struct.Fields>(
...
+      Effect.timeoutOrElse({
+        duration: timeout,
+        orElse: () =>
+          Effect.fail(
+            new McpExaError(
+              { kind: "network", source: credential.source },
+              `${tool} request timed out`,
+            ),
+          ),
+      }),
```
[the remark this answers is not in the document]
Astro-Han: Handled in the `McpExa.call` refactor. The timeout now wraps `http.execute` plus `response.text`, and the `mapError` branch preserves existing `McpExaError` instances instead of nesting them.

packages/desktop-electron/src/main/ipc.ts
Astro-Han: Fixed. `websearch-set-enabled` now rejects when any live instance fails to invalidate, so the renderer catch path rolls back the optimistic toggle instead of silently showing stale state.

packages/opencode/src/settings/index.ts
Astro-Han: Fixed. Runtime settings now load from `Storage` before the service is exposed, and `setWebSearchEnabled` persists through the same service path. Added settings tests for persisted startup state and concurrent toggle writes.

packages/opencode/src/tool/websearch-auth.ts
Astro-Han: Fixed. Saving an Exa API key no longer calls `web_search_exa`, so Save does not spend quota. The first real search still marks `invalid_key` / quota state through the existing recovery path.

Astro-Han: Fixed by removing eager validation. Network failures during validation can no longer block saving a syntactically present key, because Save is now a local credential update.

packages/opencode/src/tool/websearch-auth.ts
Astro-Han: Fixed. The validation probe and inner `credential` variable were removed, so this shadowing path no longer exists.

packages/opencode/src/tool/mcp-exa.ts
Astro-Han: Fixed. `statusFromBody` now only accepts error-shaped status text and has regression coverage for incidental text like `document status 404 was returned`.

packages/app/src/pages/session/message-timeline.tsx
Astro-Han: Fixed. Toast dedupe no longer uses expiring timestamps. Each surfaced failure id is tracked once per session view, and the scan also tracks pending websearch parts so normal running-to-error transitions are still detected.

packages/app/src/pages/session/message-timeline.tsx
Astro-Han: Fixed. The effect now keeps a per-message part cursor and only revisits pending websearch parts that can still transition to error, instead of rewalking every historical part on every sync tick.

packages/app/src/components/settings-general.tsx
Astro-Han: Fixed. The settings row no longer falls back to anonymous status before the resource resolves; it shows a small `Checking` / `检查中` chip instead.

packages/app/src/components/settings-general.tsx
Astro-Han: Fixed in the dialog rewrite. Key entry now lives in `DialogConnectWebSearch`, and the input clears validation state on change.

packages/app/src/components/settings-general.tsx
Astro-Han: Fixed in the dialog rewrite. Save buttons are disabled for empty trimmed input, with a handler guard as a fallback.

packages/desktop-electron/src/preload/types.ts
Astro-Han: Leaving this as a boundary trade-off rather than adding cross-package coupling. The duplicated shape exists at Electron preload, renderer window augmentation, and server virtual-module boundaries; the source-contract tests cover the current fields. A shared exported type can wait until the settings/API type boundary is refactored more broadly.

packages/app/src/pages/session/websearch-toasts.ts
```diff
@@ -0,0 +1,67 @@
...
+      : kind === "invalid_key"
+        ? ({
+            id,
+            titleKey: "toast.websearch.invalidKey.title",
+            descriptionKey: "toast.websearch.invalidKey.description",
+            actionKey: "toast.websearch.action.openSettings",
+          } satisfies WebSearchRecoveryToast)
+        : undefined
+  if (!toast) return
+  input.seen.set(id, input.now)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Dedupe is now a permanent per-failure marker for the current session view, so old failed parts do not re-arm on later rerenders.

packages/opencode/test/agent/agent.test.ts
```diff
@@ -436,6 +436,17 @@ test("webfetch is allowed by default", async () => {
...
 
+test("websearch is allowed by default", async () => {
+  await using tmp = await tmpdir()
+  await Instance.provide({
+    directory: tmp.path,
+    fn: async () => {
+      const build = await Agent.get("build")
+      expect(evalPerm(build, "websearch")).toBe("allow")
+    },
+  })
```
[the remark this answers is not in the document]
Astro-Han: Not changed. This file already uses `Instance.provide` broadly around these agent-permission checks, and this added assertion follows the local pattern without introducing new fixture complexity.

packages/opencode/test/tool/mcp-exa.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    })
+  })
+
+  test("unknown failure copy does not ask users to configure a key", async () => {
+    const McpExa = await import("../../src/tool/mcp-exa")
+
+    expect(McpExa.messageForFailure({ kind: "unknown", source: "anonymous", status: 500 })).not.toMatch(
+      /key|settings|configure/i,
+    )
+  })
```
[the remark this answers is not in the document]
Astro-Han: Not changed. The `mcp-exa` tests use explicit `HttpClient` fakes per case, which is the behavior under test here. Switching the harness would be style churn without changing the risk being covered.

packages/opencode/test/tool/websearch.test.ts
```diff
@@ -0,0 +1,78 @@
...
+            kind: "quota_exceeded",
+            source: "anonymous",
+            status: 429,
+          },
+        },
+      },
+    })
+    expect(JSON.stringify(metadata)).not.toContain("exaApiKey")
+  })
+})
```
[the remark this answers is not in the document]
Astro-Han: Not changed. These websearch tests need explicit fake service layers around auth and HTTP recovery metadata. The current helper keeps that dependency injection visible and scoped to the tests.

packages/desktop-electron/src/main/ipc.ts
Astro-Han: Fixed. The IPC handler now throws after logging per-instance failures, and the app-side mirror prevents rollback loops while restoring the UI toggle.

packages/opencode/src/tool/mcp-exa.ts
Astro-Han: Fixed. Status parsing is now limited to error-shaped status text, with regression coverage proving ordinary content text containing `status 404` stays `unknown`.

packages/opencode/src/tool/mcp-exa.ts
Astro-Han: Fixed. Empty or non-SSE 200 responses now fail with an `unknown` `McpExaError` instead of returning `undefined` as a successful result.

packages/opencode/src/tool/mcp-exa.ts
Astro-Han: Handled as a product recovery choice. Bare 402 is classified as `quota_exceeded` so anonymous built-in quota exhaustion still gets stable UI recovery; 429 and quota/rate-limit body text are also covered. Added HTTP 402 regression coverage.

packages/opencode/src/tool/mcp-exa.ts
Astro-Han: Leaving backend-originated tool errors in English intentionally. These messages are model-facing/tool-result text; user-facing settings, chips, dialog copy, and recovery toasts use i18n keys.

packages/app/src/pages/session/websearch-toasts.ts
Astro-Han: Fixed. `callID` is now read inside the guarded failure extraction path from a checked record, and falls back to `part.id` only after the part is known to be a websearch tool error.

packages/opencode/src/tool/mcp-exa.ts
Astro-Han: Fixed by removing eager validation entirely. There is no validation search query now, so this hardcoded product-name probe no longer exists.

packages/opencode/src/tool/websearch-auth.ts
Astro-Han: No code change. This is an internal credential namespace and matches the current product storage boundary; changing it now would create migration risk without user-visible benefit.

packages/app/src/components/dialog-connect-websearch.tsx
```diff
@@ -0,0 +1,281 @@
...
+                    {language.t("dialog.websearch.action.remove")}
+                  </Button>
+                  <Button
+                    size="large"
+                    variant="primary"
+                    disabled={saving() || apiKeyInput().trim() === ""}
+                    onClick={handleSave}
+                  >
+                    {language.t("dialog.websearch.action.update")}
+                  </Button>
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Save and remove now cross-disable each other, and the handlers no-op if either operation is already in flight.

packages/app/src/components/dialog-connect-websearch.tsx
```diff
@@ -0,0 +1,281 @@
...
+                  <Button
+                    size="large"
+                    variant="primary"
+                    disabled={saving() || apiKeyInput().trim() === ""}
+                    onClick={handleSave}
+                  >
+                    {language.t("dialog.websearch.action.saveShort")}
+                  </Button>
+                </div>
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The needs-attention branch uses the same cross-disable logic and handler guards as the healthy saved-key branch.

packages/opencode/src/tool/websearch-auth.ts
```diff
@@ -0,0 +1,139 @@
...
+    needsAttention: boolean
+    quotaExceeded: boolean
+  }
+
+  export interface Interface {
+    readonly credential: () => Effect.Effect<Credential, Auth.AuthError>
+    readonly status: () => Effect.Effect<Status, Auth.AuthError>
+    readonly saveKey: (key: string) => Effect.Effect<Status, Auth.AuthError | Error | McpExa.McpExaError>
+    readonly removeKey: () => Effect.Effect<Status, Auth.AuthError>
+    readonly markNeedsAttention: (failure: McpExa.Failure) => Effect.Effect<void, Auth.AuthError>
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The empty-key branch now uses `WebSearchAuthMissingKeyError` via `Schema.TaggedErrorClass` instead of leaking a bare `Error`.

packages/opencode/test/tool/websearch-auth.test.ts
```diff
@@ -0,0 +1,177 @@
+import { afterEach, describe, expect, test } from "bun:test"
+import { Effect, Layer } from "effect"
+import { HttpClient, HttpClientResponse } from "effect/unstable/http"
```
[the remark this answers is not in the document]
Astro-Han: Not changed. The auth tests intentionally use a small local runner because each case injects a custom in-memory Auth layer. Refactoring to the shared harness would be style-only churn for this PR.

packages/app/src/context/settings.tsx
```diff
@@ -189,6 +191,20 @@ export const { use: useSettings, provider: SettingsProvider } = createSimpleCont
...
+      const value = store.general?.webSearchEnabled ?? defaultSettings.general.webSearchEnabled
+      if (rollingBackWebSearchEnabled) {
+        rollingBackWebSearchEnabled = false
+        return
+      }
+      void window.api?.setWebSearchEnabled?.(value)?.catch(() => {
+        rollingBackWebSearchEnabled = true
+        setStore("general", "webSearchEnabled", !value)
+      })
+    })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The rollback handlers now ignore stale IPC failures when the user has already toggled again, and the same guard is applied to both LSP and Web Search mirror effects.

packages/app/src/pages/session/websearch-toasts.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    })
+
+    expect(toast).toEqual({
+      id: "call-1",
+      titleKey: "toast.websearch.quota.title",
+      descriptionKey: "toast.websearch.quota.description",
+      actionKey: "toast.websearch.action.openSettings",
+    })
+    expect(JSON.stringify(toast)).not.toContain("key")
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The quota toast test now includes a synthetic secret in failure metadata and asserts neither the metadata key nor the secret leaks into the toast payload.

packages/app/src/pages/session/websearch-toasts.ts
```diff
@@ -0,0 +1,72 @@
...
+          actionKey: "toast.websearch.action.openSettings",
+        } satisfies WebSearchRecoveryToast)
+      : kind === "invalid_key"
+        ? ({
+            id: context.id,
+            titleKey: "toast.websearch.invalidKey.title",
+            descriptionKey: "toast.websearch.invalidKey.description",
+            actionKey: "toast.websearch.action.openSettings",
+          } satisfies WebSearchRecoveryToast)
+        : undefined
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Saved-key invalid failures still open settings, but env-backed invalid keys no longer show the saved-key recovery toast because the dialog cannot repair an environment variable.

packages/opencode/test/tool/registry.test.ts
```diff
@@ -551,4 +552,65 @@ describe("tool.registry", () => {
...
+          agent: { name: "build", mode: "primary", permission: [], options: {} },
+        })
+        const ids = tools.map((tool) => tool.id)
+
+        expect(ids).toContain("websearch")
+        expect(ids).toContain("webfetch")
+        expect(ids).not.toContain("codesearch")
+      },
+    })
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The registry test now captures and restores the previous Web Search setting in a finally block for isolation.

packages/desktop-electron/src/main/ipc.ts
```diff
@@ -206,6 +206,47 @@ export function registerIpcHandlers(deps: Deps) {
...
+          directory: directories[index],
+          error: result.reason,
+        })
+      }
+    }
+    const failures = results.filter((result) => result.status === "rejected")
+    if (failures.length > 0) {
+      throw new Error(`Failed to refresh Web Search tools in ${failures.length} project instance(s)`)
+    }
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. The IPC handler now captures the previous Web Search setting and restores it before throwing if any live instance invalidation fails, so persisted/runtime state does not stay on the failed value.

packages/opencode/src/settings/index.ts
```diff
@@ -1,29 +1,56 @@
...
+      const lspEnabled = yield* Ref.make<boolean>(stored.lspEnabled ?? false)
+      const webSearchEnabled = yield* Ref.make<boolean>(stored.webSearchEnabled ?? true)
+      const persistLock = yield* Semaphore.make(1)
+      const persist = (patch: Stored, apply: Effect.Effect<void>) =>
+        persistLock.withPermit(
+          Effect.gen(function* () {
+            const current = yield* storage
+              .read<Stored>(STORAGE_KEY)
+              .pipe(Effect.catch(() => Effect.succeed({} as Stored)))
+            yield* storage.write(STORAGE_KEY, { ...current, ...patch })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Runtime settings now only fall back to an empty document for `Storage.NotFoundError`; non-missing read failures are surfaced instead of being overwritten on the next setter. Added regression coverage.

packages/opencode/src/tool/mcp-exa.ts
```diff
@@ -18,13 +95,26 @@ const McpResult = Schema.Struct({
...
+    const text = data.result.content[0]?.text
+    if (data.result.isError) {
+      const failure = classifyFailure({ body: text ?? "", source })
+      return yield* Effect.fail(new McpExaError(failure, messageForFailure(failure)))
+    }
+    if (text) return text
   }
-  return undefined
+  const message = sawData ? "Web Search returned an empty Exa response." : "Web Search did not receive an Exa response."
+  return yield* Effect.fail(new McpExaError({ kind: "unknown", source }, message))
```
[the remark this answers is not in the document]
Astro-Han: Fixed. SSE parsing now buffers `data:` lines until the event terminator and decodes the joined payload, while preserving typed `McpExaError` wrapping for malformed events. Added multiline SSE coverage.

PR #284, opened 2026-04-28
fix: guard session message cache reads

Astro-Han:
## Summary

Guard session message cache reads behind a shared safe reader, and reuse it in the session page, session context tab, and session command navigation.

## Why

PawWork 0.2.14 produced a renderer crash on a session route with `Cannot read properties of undefined (reading 'filter')`. The packaged stack maps to `messages().filter(...)` in the session page. If background prefetch or cache eviction leaves a session message cache value temporarily missing or malformed, the renderer should treat it as an empty message list instead of crashing.

## Related Issue

No GitHub issue yet. This PR is based on local problem report `pwr_moibnr62_2cfkgq`.

## How To Verify

List the commands run for this PR:

```bash
bun install --frozen-lockfile
bun test --preload ./happydom.ts src/pages/session/session-messages.test.ts src/pages/session/use-session-commands.test.ts src/pages/session/session-running-state.test.ts src/components/session/session-context-metrics.test.ts
bun run test:unit
bun run typecheck
```

## Screenshots or Recordings

Not applicable. This is a renderer crash guard with no intended visible UI change.

## Checklist

### DIFF
--- a/packages/app/src/components/session/session-context-tab.tsx
+++ b/packages/app/src/components/session/session-context-tab.tsx
@@ -10,10 +10,11 @@ import { StickyAccordionHeader } from "@opencode-ai/ui/sticky-accordion-header"
 import { File } from "@opencode-ai/ui/file"
 import { Markdown } from "@opencode-ai/ui/markdown"
 import { ScrollView } from "@opencode-ai/ui/scroll-view"
-import type { Message, Part, UserMessage } from "@opencode-ai/sdk/v2/client"
+import type { Message, Part } from "@opencode-ai/sdk/v2/client"
 import { useLanguage } from "@/context/language"
 import { useProviders } from "@/hooks/use-providers"
 import { useSessionLayout } from "@/pages/session/session-layout"
+import { emptyMessages, emptyUserMessages, readSessionMessages, readUserMessages } from "@/pages/session/session-messages"
 import { getSessionContextMetrics } from "./session-context-metrics"
 import { estimateSessionContextBreakdown, type SessionContextBreakdownKey } from "./session-context-breakdown"
 import { createSessionContextFormatter } from "./session-context-format"
@@ -87,9 +88,6 @@ function RawMessage(props: {
   )
 }
 
-const emptyMessages: Message[] = []
-const emptyUserMessages: UserMessage[] = []
-
 export function SessionContextTab() {
   const sync = useSync()
   const language = useLanguage()
@@ -101,15 +99,14 @@ export function SessionContextTab() {
   const messages = createMemo(
     () => {
       const id = params.id
-      if (!id) return emptyMessages
-      return (sync.data.message[id] ?? []) as Message[]
+      return readSessionMessages(id ? sync.data.message[id] : undefined)
     },
     emptyMessages,
     { equals: same },
   )
 
   const userMessages = createMemo(
-    () => messages().filter((m) => m.role === "user") as UserMessage[],
+    () => readUserMessages(messages()),
     emptyUserMessages,
     { equals: same },
   )
--- a/packages/app/src/pages/session/session-messages.test.ts
+++ b/packages/app/src/pages/session/session-messages.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, test } from "bun:test"
+import type { Message } from "@opencode-ai/sdk/v2/client"
+import { readSessionMessages, readUserMessages } from "./session-messages"
+
+const message = (id: string, role: Message["role"]): Message =>
+  ({
+    id,
+    sessionID: "ses_1",
+    role,
+    time: { created: 1 },
+  }) as Message
+
+describe("session message readers", () => {
+  test("returns a stable empty list for missing or invalid session cache values", () => {
+    const empty = readSessionMessages(undefined)
+
+    expect(empty).toHaveLength(0)
+    expect(Object.isFrozen(empty)).toBe(true)
+    expect(readSessionMessages(null)).toBe(empty)
+    expect(readSessionMessages({})).toBe(empty)
+  })
+
+  test("preserves loaded message arrays", () => {
+    const loaded = [message("msg_1", "user"), message("msg_2", "assistant")]
+
+    expect(readSessionMessages(loaded)).toBe(loaded)
+  })
+
+  test("filters user messages from a safe message list", () => {
+    const loaded = [message("msg_1", "assistant"), message("msg_2", "user")]
+
+    expect(readUserMessages(readSessionMessages(loaded)).map((item) => item.id)).toEqual(["msg_2"])
+  })
+
+  test("returns a stable empty user list for missing or invalid inputs", () => {
+    const empty = readUserMessages(undefined)
+
+    expect(empty).toHaveLength(0)
+    expect(Object.isFrozen(empty)).toBe(true)
+    expect(readUserMessages("not an array")).toBe(empty)
+  })
+
+  test("skips malformed entries while filtering user messages", () => {
+    const loaded = [null, {}, message("msg_1", "assistant"), message("msg_2", "user")]
+
+    expect(readUserMessages(loaded).map((item) => item.id)).toEqual(["msg_2"])
+  })
+})
--- a/packages/app/src/pages/session/session-messages.ts
+++ b/packages/app/src/pages/session/session-messages.ts
@@ -1,0 +1,18 @@
+import type { Message, UserMessage } from "@opencode-ai/sdk/v2/client"
+
+export const emptyMessages = Object.freeze([]) as unknown as Message[]
+export const emptyUserMessages = Object.freeze([]) as unknown as UserMessage[]
+
+export function readSessionMessages(value: unknown): Message[] {
+  return Array.isArray(value) ? (value as Message[]) : emptyMessages
+}
+
+function isUserMessage(value: unknown): value is UserMessage {
+  return !!value && typeof value === "object" && "role" in value && value.role === "user"
+}
+
+export function readUserMessages(messages: unknown): UserMessage[] {
+  if (!Array.isArray(messages)) return emptyUserMessages
+  const users = messages.filter(isUserMessage)
+  return users.length > 0 ? users : emptyUserMessages
+}
--- a/packages/app/src/pages/session/use-session-commands.tsx
+++ b/packages/app/src/pages/session/use-session-commands.tsx
@@ -16,6 +16,7 @@ import { showToast } from "@opencode-ai/ui/toast"
 import { findLast } from "@opencode-ai/util/array"
 import { canCloseSessionTab, closeSessionTab } from "@/pages/session/close-session-tab"
 import { createSessionTabs } from "@/pages/session/helpers"
+import { readSessionMessages, readUserMessages } from "@/pages/session/session-messages"
 import { toggleDesktopTerminal } from "@/pages/session/terminal-shell-tab"
 import { extractPromptFromParts } from "@/utils/prompt"
 import { UserMessage } from "@opencode-ai/sdk/v2"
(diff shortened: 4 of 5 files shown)


Astro-Han (commented):
## Review: fix: guard session message cache reads

### P0 — Critical

**None.** The crash guard logic is sound and the fix addresses the reported issue correctly.

### P1 — Important

**1. `readSessionMessages` silently swallows non-array truthy values**

In `session-messages.ts:7`, if `sync.data.message[id]` is ever a non-array object (e.g., a loading placeholder, a corrupted cache entry, or a future API change), `readSessionMessages` returns `emptyMessages` without any trace. This makes debugging future cache corruption extremely difficult. Consider adding a `console.warn` or `console.debug` when `value` is truthy but not an array, so that developers can detect when the cache shape has drifted.

**2. `readUserMessages` uses `emptyUserMessages` for empty results but not for the `users` array itself**

In `session-messages.ts:13`, when `users.length > 0` you return `users`, but `users` is a newly allocated array on every call. The original code in `session-context-tab.tsx` and `session.tsx` used `emptyUserMessages` as the memo initial value and relied on stable references for `createMemo` equality checks. While `readUserMessages` is now called inside `createMemo` with `{ equals: same }`, the returned `users` array is still a new reference on every filter. This is fine for `same` equality (which does deep comparison), but if any caller ever uses reference equality, it will re-trigger unnecessarily. Consider whether `readUserMessages` should be memoized or whether callers should be audited for reference-equality assumptions.

**3. `use-session-commands.tsx` loses memoization entirely**

In `use-session-commands.tsx:78-82`, `messages` and `userMessages` are plain functions (`() => ...`), not `createMemo`. The original code returned `[]` (a new array literal) on every call when `!id`, which was already a bug. The PR replaces this with `readSessionMessages(...)` which returns `emptyMessages` (stable) when `!id`, but `sync.data.message[id]` (unstable reference) when the id exists. Since these are plain functions called on every render, any downstream effect or memo that depends on reference equality will fire continuously. This is a pre-existing issue, but the PR touches this code and should either fix it or leave a `// TODO: memoize` comment.

### P2 — Moderate

**4. Test coverage for `readUserMessages` with non-array input is missing**

`session-messages.test.ts` tests `readSessionMessages` with `undefined`, `null`, and `{}`, but `readUserMessages` is only tested with a valid array output from `readSessionMessages`. Add a test case for `readUserMessages(undefined)` and `readUserMessages("not an array")` to document and lock in the defensive behavior.

**5. `readUserMessages` type signature is overly permissive**

```ts
export function readUserMessages(messages: readonly Message[] | undefined): UserMessage[]
```

The `undefined` union is only needed because `readSessionMessages` could theoretically return something unexpected. But since `readSessionMessages` is guaranteed to return `Message[]`, the `| undefined` in `readUserMessages` signals to callers that they need to handle `undefined`, which is not actually a real case. Consider narrowing to `readonly Message[]` and handling the guard internally with an assertion, or document why `undefined` is part of the public contract.

**6. `emptyMessages` and `emptyUserMessages` are exported mutable arrays**

In `session-messages.ts:3-4`:

```ts
export const emptyMessages: Message[] = []
export const emptyUserMessages: UserMessage[] = []
```

These are exported as mutable arrays. While the current code never mutates them, future developers could accidentally `.push()` into these shared globals, causing cross-session pollution. Freeze them with `Object.freeze([])` or cast through `as const` to make the type system reject mutations.

**7. PR description claims tests pass but `test:unit` is overly broad**

The PR description lists `bun run test:unit` as a verification step. Given that this PR only touches session message reading, running the full unit suite is wasteful and doesn't demonstrate that the specific new tests were actually executed. The verification steps should list the focused test command (which they do as the first line), but the second line is redundant noise.

### P3 — Minor / Nitpick

**8. Inconsistent import grouping in `session-context-tab.tsx`**

The import of `session-messages` is placed after `session-layout` but before `./session-context-metrics`, mixing absolute (`@/pages/session/...`) and relative (`./...`) imports. The file's existing convention groups relative imports after absolute ones. Move the `session-messages` import to the absolute import group (before `session-layout` or after it, but keep the relative imports together).

**9. `session-messages.ts` file name is ambiguous**

`session-messages.ts` sounds like it contains message data structures or types. A name like `session-message-readers.ts` or `session-message-cache.ts` would more clearly communicate that this file is about safe cache access patterns, not message domain models.

**10. `readSessionMessages` cast is a type-system bypass**

```ts
return Array.isArray(value) ? (value as Message[]) : emptyMessages
```

The `as Message[]` cast assumes that every array in the cache is a well-formed `Message[]`. If the cache ever contains an array of primitives or malformed objects, this cast hides the problem. Consider a runtime shape check (e.g., `value.every(isMessage)`) or at minimum document the trust boundary with a comment explaining why the cast is safe.

**11. Missing newline at end of `session-messages.test.ts`**

The test file ends at line 31 with no trailing newline. Most editors and git diff tools flag this. Add a trailing newline for consistency with the rest of the codebase.

packages/app/src/components/session/session-context-tab.tsx
```diff
@@ -10,10 +10,11 @@ import { StickyAccordionHeader } from "@opencode-ai/ui/sticky-accordion-header"
 import { File } from "@opencode-ai/ui/file"
 import { Markdown } from "@opencode-ai/ui/markdown"
 import { ScrollView } from "@opencode-ai/ui/scroll-view"
-import type { Message, Part, UserMessage } from "@opencode-ai/sdk/v2/client"
+import type { Message, Part } from "@opencode-ai/sdk/v2/client"
 import { useLanguage } from "@/context/language"
```
Astro-Han, line 14: **[P3]** Import order: this absolute import is placed between relative imports. Group absolute imports before relative ones per the file's existing convention.

packages/app/src/pages/session/session-messages.ts
```diff
@@ -0,0 +1,14 @@
+import type { Message, UserMessage } from "@opencode-ai/sdk/v2/client"
+
+export const emptyMessages: Message[] = []
```
Astro-Han: **[P2]** Exported mutable shared arrays. Future code could .push() into these, causing cross-session pollution. Freeze them: export const emptyMessages: Message[] = Object.freeze([]).

packages/app/src/pages/session/session-messages.ts
```diff
@@ -0,0 +1,14 @@
...
+export const emptyUserMessages: UserMessage[] = []
+
+export function readSessionMessages(value: unknown): Message[] {
+  return Array.isArray(value) ? (value as Message[]) : emptyMessages
```
Astro-Han, line 7: **[P1]** Silently swallowing non-array truthy values makes future cache corruption impossible to debug. Consider console.warn when value is truthy but not an array.

**[P3]** Type-system bypass. This assumes every array in the cache is a well-formed Message[]. A runtime shape check or at least a comment documenting the trust boundary would be safer.

packages/app/src/pages/session/session-messages.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export function readUserMessages(messages: readonly Message[] | undefined): UserMessage[] {
+  if (!Array.isArray(messages)) return emptyUserMessages
+  const users = messages.filter((message): message is UserMessage => message.role === "user")
+  return users.length > 0 ? users : emptyUserMessages
```
Astro-Han, line 17: **[P1]** users is a newly allocated array on every call. Callers inside createMemo with { equals: same } are safe, but any reference-equality consumer will re-trigger. Consider whether this should be memoized or documented.

packages/app/src/pages/session/session-messages.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    expect(readSessionMessages({})).toEqual([])
+  })
+
+  test("preserves loaded message arrays", () => {
+    const loaded = [message("msg_1", "user"), message("msg_2", "assistant")]
+
+    expect(readSessionMessages(loaded)).toBe(loaded)
+  })
+
+  test("filters user messages from a safe message list", () => {
```
Astro-Han, line 29: **[P2]** Missing test coverage for readUserMessages with non-array input (undefined, string, etc.). Add defensive test cases to lock in the behavior.

packages/app/src/pages/session/use-session-commands.tsx
```diff
@@ -16,6 +16,7 @@ import { showToast } from "@opencode-ai/ui/toast"
 import { findLast } from "@opencode-ai/util/array"
 import { canCloseSessionTab, closeSessionTab } from "@/pages/session/close-session-tab"
 import { createSessionTabs } from "@/pages/session/helpers"
+import { readSessionMessages, readUserMessages } from "@/pages/session/session-messages"
 import { toggleDesktopTerminal } from "@/pages/session/terminal-shell-tab"
```
Astro-Han, line 20: **[P1]** These are plain functions, not createMemo. When id exists, sync.data.message[id] returns an unstable reference, causing continuous re-renders for any downstream reference-equality consumer. Pre-existing bug, but since this PR touches the code it should either be fixed or annotated with TODO: memoize.

packages/app/src/pages/session/session-messages.ts
```diff
@@ -0,0 +1,14 @@
...
+export const emptyMessages: Message[] = []
+export const emptyUserMessages: UserMessage[] = []
+
+export function readSessionMessages(value: unknown): Message[] {
+  return Array.isArray(value) ? (value as Message[]) : emptyMessages
+}
+
+export function readUserMessages(messages: readonly Message[] | undefined): UserMessage[] {
+  if (!Array.isArray(messages)) return emptyUserMessages
+  const users = messages.filter((message): message is UserMessage => message.role === "user")
```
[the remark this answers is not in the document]
Astro-Han: Fixed in ef1dce8. `readUserMessages` now treats the input as unknown, checks each array element before reading `role`, and the test covers malformed entries.

packages/app/src/pages/session/session-messages.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    role,
+    time: { created: 1 },
+  }) as Message
+
+describe("session message readers", () => {
+  test("returns a stable empty list for missing or invalid session cache values", () => {
+    expect(readSessionMessages(undefined)).toEqual([])
+    expect(readSessionMessages(null)).toEqual([])
+    expect(readSessionMessages({})).toEqual([])
+  })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in ef1dce8. The test now asserts stable identity for invalid cache fallbacks and verifies the fallback is frozen.

packages/app/src/pages/session/session-messages.ts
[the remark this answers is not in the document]
Astro-Han: Fixed in ef1dce8. `readUserMessages` now uses an element guard before checking `role`, while keeping the empty-user fallback behavior.

packages/app/src/components/session/session-context-tab.tsx
Astro-Han: Leaving unchanged after checking the file. This is an absolute `@/...` import and it is still before the relative `./...` imports, which matches the file existing grouping.

packages/app/src/pages/session/session-messages.ts
Astro-Han: Fixed in ef1dce8. The shared empty fallbacks are now runtime-frozen, and tests assert `Object.isFrozen` for both session and user fallbacks. I kept the exported array type compatible with existing callers to avoid broad readonly churn in this bugfix.

packages/app/src/pages/session/session-messages.ts
Astro-Han: Partially handled in ef1dce8: `readUserMessages` now accepts unknown input and ignores malformed array entries before reading `role`, with tests. I am not adding `console.warn` here because this helper runs in render/memo paths and the intended behavior is a quiet renderer fallback for corrupted cache values. If we want diagnostics, that should be a separate cache-integrity logging path, not a UI reader side effect.

packages/app/src/pages/session/session-messages.ts
Astro-Han: Leaving unchanged. The allocation only happens for valid arrays, and changed createMemo callers already use `{ equals: same }`. I do not see a concrete reference-equality consumer here, so memoizing or adding a TODO would be speculative.

packages/app/src/pages/session/session-messages.test.ts
Astro-Han: Fixed in ef1dce8. Added invalid-input coverage for `readUserMessages`, including stable empty identity for undefined and string input.

PR #285, opened 2026-04-28
fix: include release version in installer artifact names

### LINKED ISSUE #256: [Feature] Include version in installer artifact filenames

When I install or download a new PawWork release, I want the installer file to clearly show which version it is, so I can keep multiple versions side by side, share a specific build with someone, or roll back to a previous installer if a new one is broken.

### What do you do today?

Today every release ships installers named `pawwork-mac-arm64.dmg` / `pawwork-mac-x64.dmg` / `pawwork-win-x64.exe`. Downloading the new release into the same Downloads folder overwrites the previous build with the same name, so I lose the old installer. Renaming after the fact is manual and forgettable, and screenshots of the file name from bug reports do not show which version the user is on.

Each release artifact filename embeds the version, e.g. `pawwork-0.2.14-mac-arm64.dmg`, `pawwork-0.2.14-win-x64.exe`. Different versions coexist in the Downloads folder, the version is visible from the file name in finder / explorer screenshots, and the in-app electron-updater upgrade path keeps working across the rename.

Global users

### Extra context

Implementation pointers (for whoever picks this up):

- `packages/desktop-electron/electron-builder.config.ts:55` artifactName template currently is `pawwork-${os}-${arch}.${ext}`. Switch to `pawwork-${version}-${os}-${arch}.${ext}`.
- `packages/desktop-electron/electron-builder-app-update.test.ts:48` snapshot assertion needs to follow.
- `packages/desktop-electron/scripts/verify-release.test.ts` has 16+ hardcoded asset names and simulated updater yml fixtures that need to be regenerated for the new template.
- electron-updater consumes the asset URL from `latest-mac.yml` / `latest.yml`, both of which electron-builder regenerates automatically. The updater path should keep working across the rename, but it is unverified — the upgrade jump from a `0.2.x` build (old filename) to the first `0.2.y` build (new filename) must be smoke-tested manually before shipping.
- Should ship in a release after `v0.2.14`, not bundled with a hotfix.

Astro-Han:
## Summary

- Move desktop release artifact names to platform-first, version-last naming: `pawwork-${os}-${arch}-${version}.${ext}`.
- Update release verification to derive expected installer and updater asset names from the release tag/version instead of hardcoded old filenames.
- Bump PawWork release packages and lockfile metadata to the first CalVer release version, `2026.4.28`.

Closes #256.

## Verification

- `bun install --frozen-lockfile`
- `bun test electron-builder-app-update.test.ts scripts/verify-release.test.ts` from `packages/desktop-electron`
- `bun run typecheck:release` from `packages/desktop-electron`

## Release Follow-up

The first CalVer release still needs the release workflow smoke that verifies an installed `v0.2.14` app can discover, download, and apply `v2026.4.28` through auto-update using the renamed assets.

## Release Notes

**Chores**
- All packages and components updated from version 0.2.14 to 2026.4.28, adopting calendar versioning (CalVer, YYYY.M.D format) across the entire application suite.
- Desktop application release artifacts now include version identifiers in installer and updater asset filenames for all distribution channels.
- Release verification and validation processes updated to support the new calendar versioning standards and enforce corresponding asset naming conventions.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -26,7 +26,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "0.2.14",
+      "version": "2026.4.28",
       "dependencies": {
         "@kobalte/core": "catalog:",
         "@opencode-ai/sdk": "workspace:*",
@@ -110,7 +110,7 @@
     },
     "packages/desktop-electron": {
       "name": "@opencode-ai/desktop-electron",
-      "version": "0.2.14",
+      "version": "2026.4.28",
       "dependencies": {
         "@opencode-ai/util": "workspace:*",
         "effect": "catalog:",
@@ -175,7 +175,7 @@
     },
     "packages/opencode": {
       "name": "opencode",
-      "version": "0.2.14",
+      "version": "2026.4.28",
       "bin": {
         "opencode": "./bin/opencode",
       },
@@ -402,7 +402,7 @@
     },
     "packages/util": {
       "name": "@opencode-ai/util",
-      "version": "0.2.14",
+      "version": "2026.4.28",
       "dependencies": {
         "@opencode-ai/core": "workspace:*",
         "zod": "catalog:",
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opencode-ai/app",
-  "version": "0.2.14",
+  "version": "2026.4.28",
   "description": "",
   "type": "module",
   "exports": {
--- a/packages/desktop-electron/electron-builder-app-update.test.ts
+++ b/packages/desktop-electron/electron-builder-app-update.test.ts
@@ -10,7 +10,11 @@ import { serializeAppUpdateConfig } from "./scripts/write-app-update-config"
 const roots: string[] = []
 type AfterPackContext = Parameters<Extract<NonNullable<Configuration["afterPack"]>, (...args: any[]) => unknown>>[0]
 
-function macAfterPackContext(appOutDir: string, appBundleName: string, electronPlatformName = "darwin"): AfterPackContext {
+function macAfterPackContext(
+  appOutDir: string,
+  appBundleName: string,
+  electronPlatformName = "darwin",
+): AfterPackContext {
   return {
     appOutDir,
     electronPlatformName,
@@ -50,6 +54,12 @@ describe("electron builder app-update config", () => {
     expect(createConfig("prod").mac?.extendInfo).toMatchObject({
       LSHasLocalizedDisplayName: true,
     })
+  })
+
+  test("all channels share the versioned artifact name", () => {
+    expect(createConfig("dev").artifactName).toBe("pawwork-${os}-${arch}-${version}.${ext}")
+    expect(createConfig("beta").artifactName).toBe("pawwork-${os}-${arch}-${version}.${ext}")
+    expect(createConfig("prod").artifactName).toBe("pawwork-${os}-${arch}-${version}.${ext}")
   })
 
   test("afterPack writes app-update.yml to the packager-reported macOS resources path", async () => {
--- a/packages/desktop-electron/electron-builder-app-update.test.ts
+++ b/packages/desktop-electron/electron-builder-app-update.test.ts
@@ -45,7 +45,7 @@ describe("electron builder app-update config", () => {
     const config = createConfig("prod")
     expect(config.productName).toBe("PawWork")
     expect(config.appId).toBe("ai.pawwork.desktop")
-    expect(config.artifactName).toBe("pawwork-${os}-${arch}.${ext}")
+    expect(config.artifactName).toBe("pawwork-${os}-${arch}-${version}.${ext}")
     expect(config.publish).toMatchObject({ owner: "Astro-Han", repo: "pawwork" })
     expect(createConfig("prod").mac?.extendInfo).toMatchObject({
       LSHasLocalizedDisplayName: true,
--- a/packages/desktop-electron/electron-builder.config.ts
+++ b/packages/desktop-electron/electron-builder.config.ts
@@ -52,7 +52,7 @@ async function writeLocalizedMacDisplayName(resourcesDir: string, channel: Chann
 }
 
 const getBase = (): Configuration => ({
-  artifactName: "pawwork-${os}-${arch}.${ext}",
+  artifactName: "pawwork-${os}-${arch}-${version}.${ext}",
   directories: {
     output: "dist",
     buildResources: "resources",
--- a/packages/desktop-electron/scripts/verify-release.test.ts
+++ b/packages/desktop-electron/scripts/verify-release.test.ts
@@ -10,6 +10,8 @@ import {
   normalizeTag,
   parseUpdaterFileUrls,
   readStartupLogFile,
+  releaseAssetNames,
+  releaseUpdaterAssetNames,
   verifyReleasePayload,
   verifyStartupLog,
   type GithubRelease,
@@ -81,6 +83,25 @@ describe("verify-release", () => {
     expect(() => normalizeTag("2026.4.28-hotfix.1")).toThrow("Invalid release tag")
   })
 
+  test("derives release and updater asset names from the CalVer version", () => {
+    expect(releaseAssetNames("2026.4.28")).toEqual([
+      "pawwork-mac-arm64-2026.4.28.dmg",
+      "pawwork-mac-arm64-2026.4.28.zip",
+      "pawwork-mac-arm64-2026.4.28.zip.blockmap",
+      "pawwork-mac-x64-2026.4.28.dmg",
+      "pawwork-mac-x64-2026.4.28.zip",
+      "pawwork-mac-x64-2026.4.28.zip.blockmap",
+      "pawwork-win-x64-2026.4.28.exe",
+      "pawwork-win-x64-2026.4.28.exe.blockmap",
+      "latest.yml",
+      "latest-mac.yml",
+    ])
+    expect(releaseUpdaterAssetNames("2026.4.28")).toEqual({
+      "latest.yml": ["pawwork-win-x64-2026.4.28.exe"],
+      "latest-mac.yml": ["pawwork-mac-arm64-2026.4.28.zip", "pawwork-mac-x64-2026.4.28.zip"],
+    })
+  })
+
   test("parses updater file urls and path entries", () => {
     expect(
       parseUpdaterFileUrls(`version: 2026.4.28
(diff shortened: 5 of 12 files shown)


packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -21,18 +21,28 @@ type VerificationInput = {
...
-  "pawwork-mac-arm64.zip.blockmap",
-  "pawwork-mac-x64.dmg",
-  "pawwork-mac-x64.zip",
-  "pawwork-mac-x64.zip.blockmap",
-  "pawwork-win-x64.exe",
-  "pawwork-win-x64.exe.blockmap",
-  "latest.yml",
-  "latest-mac.yml",
-]
+export function releaseAssetNames(version: string) {
```
Astro-Han, line 36: **[P0]** `normalizeTag` still enforces `^v\d+\.\d+\.\d+$`, which rejects CalVer tags like `v2026.4.28`. The test file was updated to expect `v2026.4.28` to pass, but the regex in `normalizeTag` (line 200) has not been changed. This is a functional bug — the new version will throw at runtime.

Please update the regex to accept CalVer (e.g. `^v\d{4}\.\d{1,2}\.\d{1,2}$` or a more permissive pattern).

Astro-Han, line 36: **[P1]** `releaseAssetNames` is exported but `windowsUpdaterAsset` and `macUpdaterAssets` are not. Since all three are pure helper functions with the same coupling to the naming convention, consider exporting them consistently so tests can reference the same source of truth instead of hardcoding the strings.

Alternatively, keep them all internal and add a single exported `expectedAssets(version)` object that bundles everything.

Astro-Han, line 36: **[P2]** The old `REQUIRED_ASSETS` constant is deleted but there is no single source of truth for the platform/arch matrix (`mac-arm64`, `mac-x64`, `win-x64`). If a fourth platform is added later, the names must be updated in three places: `releaseAssetNames`, `windowsUpdaterAsset`, and `macUpdaterAssets`.

Consider extracting the matrix into a small data structure (e.g. `const PLATFORMS = [...]`) and deriving all three functions from it.

packages/desktop-electron/scripts/verify-release.ts
```diff
@@ -170,23 +180,25 @@ export function verifyStartupLog(source: string, expectedTag: string) {
 export function verifyReleasePayload(input: VerificationInput) {
   const failures: string[] = []
   const assetNames = new Set(input.release.assets.map((asset) => asset.name))
+  const version = releaseVersion(input.release.tag_name)
```
Astro-Han: **[P2]** `releaseVersion` calls `normalizeTag` and then slices off the leading `v`. If `normalizeTag` throws, the error bubbles up as a string in the `failures` array (line 156). That is fine for `verifyStartupLog`, but in `verifyReleasePayload` the `version` variable is used on line 187 before any startup-log check. If the tag is invalid, `releaseAssetNames(version)` will not execute because the exception is caught... actually wait, it is not caught here.

`releaseVersion` throws uncaught inside `verifyReleasePayload`, which would crash the caller rather than returning a failures array. The old code did not call `normalizeTag` in this path. Please either catch the error inside `verifyReleasePayload` and push it to `failures`, or validate the tag earlier.

packages/desktop-electron/scripts/verify-release.test.ts
```diff
@@ -71,54 +71,56 @@ const baseRelease: GithubRelease = {
 
 describe("verify-release", () => {
   test("normalizes release tags", () => {
```
Astro-Han, line 75: **[P1]** The new test cases for `normalizeTag` (lines 79–80) assert that `2026.4.28.1` and `2026.4.28-hotfix.1` throw. However, the regex in `normalizeTag` has not actually been updated to accept CalVer, so these tests will pass for the wrong reason (they still fail the old SemVer regex).

Please add a test that `v2026.4.28` *passes* `normalizeTag`, and update the regex first.

Astro-Han, line 75: **[P2]** The `baseRelease` fixture is updated to `v2026.4.28`, but the `browser_download_url` values still use `https://example.com/...`. That is fine for unit tests, but consider adding one test that exercises `releaseAssetNames` directly to ensure the naming convention is mechanically correct, rather than relying on the indirect coverage through `verifyReleasePayload`.

packages/desktop-electron/electron-builder.config.ts
```diff
@@ -52,7 +52,7 @@ async function writeLocalizedMacDisplayName(resourcesDir: string, channel: Chann
 }
 
 const getBase = (): Configuration => ({
-  artifactName: "pawwork-${os}-${arch}.${ext}",
+  artifactName: "pawwork-${os}-${arch}-${version}.${ext}",
```
Astro-Han, line 55: **[P2]** The `artifactName` template now includes `${version}`. electron-builder will substitute this from `package.json` version at build time, which is correct. However, please confirm that the `latest.yml` and `latest-mac.yml` metadata files emitted by electron-builder will also reference the versioned filenames — the auto-updater reads those files to find the download URL.

If electron-builder emits bare filenames in `latest.yml` (which it should, based on `artifactName`), this is fine. If it strips the version suffix, the updater will 404. A quick local `electron-builder --dir` smoke test would remove the doubt.

Astro-Han (commented):
## Review Summary

Good change overall — the CalVer migration and dynamic asset naming are well-scoped. A few nits below, mostly around edge cases and test coverage.

### Severity Legend
- **P0** — Must fix before merge (blocking)
- **P1** — Should fix before merge (high confidence)
- **P2** — Consider fixing (preference / maintainability)
- **P3** — Nitpick / style / optional

packages/desktop-electron/electron-builder-app-update.test.ts
```diff
@@ -45,7 +45,7 @@ describe("electron builder app-update config", () => {
     const config = createConfig("prod")
     expect(config.productName).toBe("PawWork")
     expect(config.appId).toBe("ai.pawwork.desktop")
-    expect(config.artifactName).toBe("pawwork-${os}-${arch}.${ext}")
+    expect(config.artifactName).toBe("pawwork-${os}-${arch}-${version}.${ext}")
```
Astro-Han, line 52: **[P3]** The assertion checks the exact template string `pawwork-${os}-${arch}-${version}.${ext}`. This is a good regression guard, but consider also asserting that `createConfig("dev")` and `createConfig("beta")` use the same `artifactName`, since the template is defined in `getBase()` and shared across channels.

packages/app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@opencode-ai/app",
-  "version": "0.2.14",
+  "version": "2026.4.28",
```
Astro-Han, line 3: **[P3]** Version bump from `0.2.14` to `2026.4.28` is repeated identically across four `package.json` files and `bun.lock`. This is correct but fragile — a future bump requires touching five files.

Not blocking for this PR, but consider whether a root-level `version` field or a workspace-version-sync script would reduce toil.

bun.lock
```diff
@@ -26,7 +26,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "0.2.14",
+      "version": "2026.4.28",
```
Astro-Han, line 29: **[P3]** `bun.lock` diffs show version bumps for all workspace packages. Please confirm this was generated by `bun install` (not hand-edited) so the lockfile hash and transitive dependency metadata remain valid.

packages/desktop-electron/scripts/verify-release.ts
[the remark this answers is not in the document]
Astro-Han: Checked the current file: `releaseVersion` is defined in this module and the current tests/typecheck pass. No code change needed for this item.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed in 327fe5a8f. The old pattern did accept `v2026.4.28`, but it was too implicit for the new contract. `normalizeTag` now accepts explicit CalVer `vYYYY.M.D` and rejects four-segment and prerelease suffix tags.

Astro-Han: Fixed in 327fe5a8f by replacing the separate updater helpers with exported `releaseUpdaterAssetNames(version)`, derived from the same target matrix as `releaseAssetNames(version)`. Tests now cover both.

Astro-Han: Fixed in 327fe5a8f. Added a single `RELEASE_TARGETS` matrix and derive release assets plus updater metadata assets from it.

packages/desktop-electron/scripts/verify-release.ts
Astro-Han: Fixed in 327fe5a8f. `verifyReleasePayload` now catches invalid tags and returns the error in `failures` instead of throwing. Added direct coverage for `v2026.4.28.1`.

packages/desktop-electron/scripts/verify-release.test.ts
Astro-Han: Fixed in 327fe5a8f. The pass case for `v2026.4.28` is covered, and the regex is now the explicit CalVer form.

Astro-Han: Fixed in 327fe5a8f. Added a direct test for `releaseAssetNames` and `releaseUpdaterAssetNames` so the naming convention is checked mechanically.

packages/desktop-electron/electron-builder.config.ts
Astro-Han: Verified locally after the fix. The mac arm64 zip target produced `dist/pawwork-mac-arm64-2026.4.28.zip`, `dist/pawwork-mac-arm64-2026.4.28.zip.blockmap`, and `dist/latest-mac.yml`; both `url` and `path` in `latest-mac.yml` reference `pawwork-mac-arm64-2026.4.28.zip`.

packages/desktop-electron/electron-builder-app-update.test.ts
Astro-Han: Fixed in 327fe5a8f. Added coverage that dev, beta, and prod all share the versioned `artifactName` template.

packages/app/package.json
Astro-Han: Deferred by YAGNI for this PR. The current release bump has a small known file set, and adding a root version source or sync script would introduce release-process behavior outside #256. This can be a separate tooling issue if the repeated daily bumps make it painful.

bun.lock
Astro-Han: Confirmed. `bun install --frozen-lockfile` passed after the version bump, and the lockfile diff is workspace version metadata only.

PR #286, opened 2026-04-28
fix(app): stabilize session timeline scroll

Astro-Han:
## Summary

- Keep stale message hashes from being re-applied while the session composer clears the hash after sending.
- Render the complete initial 10-turn session window on session switches instead of staging from a single turn.
- Add E2E coverage for continuing from an old message hash and for switching between long sessions.

## Why

Two session timeline issues could make the UI feel unstable:

- Continuing a conversation from an old message hash could clear the hash and then immediately re-apply the stale hash, sending the timeline back toward the old message.
- Switching sessions rendered the capped 10-turn window in visible stages, which could look like a flicker.

## Related Issue

No linked issue. Reported directly during debugging.

## How To Verify

List commands:

(cd packages/app && bun playwright test e2e/session/session-scroll-position.spec.ts)
(cd packages/app && bun playwright test e2e/session/session-send-stability.spec.ts)
(cd packages/app && bun run typecheck)

## Screenshots or Recordings

Not attached. This is interaction stability behavior covered by E2E checks.

## Checklist

- [x] I linked related issue, or stated why no issue
- [x] This PR has type, scope, and priority labels
- [x] I listed verification steps
- [x] I considered screenshots or recordings for visible UI changes

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests validating session timeline scrolling and message rendering across navigation and session switching.

* **Bug Fixes**
  * Improved scroll position handling when navigating via URL hash and returning to recent messages.
  * Enhanced initial message visibility during session switches for smoother transitions.

### DIFF
--- a/packages/app/e2e/session/session-scroll-position.spec.ts
+++ b/packages/app/e2e/session/session-scroll-position.spec.ts
@@ -1,0 +1,185 @@
+import type { Page } from "@playwright/test"
+import { test, expect } from "../fixtures"
+import { withSession } from "../actions"
+import { promptSelector, sessionItemSelector, sessionMessageItemSelector, sessionTurnListSelector } from "../selectors"
+import { createSdk } from "../utils"
+
+type Sdk = ReturnType<typeof createSdk>
+
+function timelineMetrics(page: Page) {
+  return page.evaluate((turnListSelector) => {
+    const list = document.querySelector(turnListSelector)
+    const viewport = list?.closest(".scroll-view__viewport")
+    if (!(viewport instanceof HTMLElement)) throw new Error("session timeline viewport not found")
+    return {
+      top: viewport.scrollTop,
+      height: viewport.scrollHeight,
+      client: viewport.clientHeight,
+      distanceFromBottom: viewport.scrollHeight - viewport.clientHeight - viewport.scrollTop,
+    }
+  }, sessionTurnListSelector)
+}
+
+async function scrollTimelineToBottom(page: Page) {
+  await page.evaluate((turnListSelector) => {
+    const list = document.querySelector(turnListSelector)
+    const viewport = list?.closest(".scroll-view__viewport")
+    if (!(viewport instanceof HTMLElement)) throw new Error("session timeline viewport not found")
+    viewport.scrollTop = viewport.scrollHeight
+  }, sessionTurnListSelector)
+}
+
+async function sendVisiblePrompt(input: { page: Page; text: string }) {
+  const prompt = input.page.locator(promptSelector)
+  await expect(prompt).toBeVisible()
+  await prompt.click()
+  await input.page.keyboard.insertText(input.text)
+  await expect.poll(async () => (await prompt.textContent())?.replace(/\u200B/g, "").trim()).toBe(input.text)
+  await input.page.keyboard.press("Enter")
+}
+
+async function seedSessionTurns(input: { sdk: Sdk; sessionID: string; count: number }) {
+  for (let i = 0; i < input.count; i++) {
+    await input.sdk.session.promptAsync({
+      sessionID: input.sessionID,
+      noReply: true,
+      parts: [
+        {
+          type: "text",
+          text: `seed turn ${i}\n${Array.from({ length: 16 }, (_, line) => `line ${line} ${"content ".repeat(8)}`).join("\n")}`,
+        },
+      ],
+    })
+  }
+}
+
+async function installMessageCountProbe(page: Page) {
+  await page.evaluate(
+    ({ maxSamples, messageSelector }) => {
+      const read = () => ({
+        url: window.location.href,
+        messages: document.querySelectorAll(messageSelector).length,
+      })
+      const samples = [read()]
+      const push = () => {
+        const next = read()
+        const prev = samples[samples.length - 1]
+        if (prev && prev.url === next.url && prev.messages === next.messages) return
+        if (samples.length < maxSamples) samples.push(next)
+      }
+      let frame = requestAnimationFrame(function tick() {
+        push()
+        frame = requestAnimationFrame(tick)
+      })
+      const observer = new MutationObserver(push)
+      observer.observe(document.body, { childList: true, subtree: true })
+      const win = window as typeof window & {
+        __opencode_e2e?: Record<string, unknown> & {
+          messageCountProbe?: { stop: () => unknown }
+        }
+      }
+      win.__opencode_e2e = {
+        ...win.__opencode_e2e,
+        messageCountProbe: {
+          stop() {
+            cancelAnimationFrame(frame)
+            observer.disconnect()
+            push()
+            delete win.__opencode_e2e?.messageCountProbe
+            return samples
+          },
+        },
+      }
+    },
+    { maxSamples: 256, messageSelector: sessionMessageItemSelector },
+  )
+}
+
+async function stopMessageCountProbe(page: Page) {
+  return page.evaluate(() => {
+    const win = window as typeof window & {
+      __opencode_e2e?: {
+        messageCountProbe?: { stop: () => unknown }
+      }
+    }
+    const probe = win.__opencode_e2e?.messageCountProbe
+    if (!probe) throw new Error("message count probe was not installed")
+    return probe.stop()
+  }) as Promise<Array<{ url: string; messages: number }>>
+}
+
+test("keeps the latest turn in view when sending from an old message hash", async ({ page, project }) => {
+  test.setTimeout(120_000)
+
+  await project.open()
+  const sdk = project.sdk
+
+  await withSession(sdk, `e2e scroll position ${Date.now()}`, async (session) => {
+    project.trackSession(session.id)
+
+    await seedSessionTurns({ sdk, sessionID: session.id, count: 14 })
+
+    await project.gotoSession(session.id)
+    await expect(page.locator(sessionMessageItemSelector)).toHaveCount(10, { timeout: 30_000 })
+    await scrollTimelineToBottom(page)
+    await expect.poll(async () => (await timelineMetrics(page)).distanceFromBottom).toBeLessThan(20)
+
+    const ids = await page.locator(sessionMessageItemSelector).evaluateAll((items) =>
+      items.map((item) => (item instanceof HTMLElement ? item.dataset.messageId : undefined)).filter(Boolean),
+    )
+    const oldID = ids[2]
+    if (!oldID) throw new Error("expected an older rendered message id")
+
+    await page.goto(`${page.url()}#message-${oldID}`)
+    await expect(page.locator(`#message-${oldID}`)).toBeVisible()
+    await expect.poll(async () => (await timelineMetrics(page)).distanceFromBottom).toBeGreaterThan(100)
+
+    const token = `scroll_latest_${Date.now()}`
+    const beforeCount = await page.locator(sessionMessageItemSelector).count()
+    await sendVisiblePrompt({ page, text: `reply with ${token}` })
+    await expect(page.locator(sessionMessageItemSelector)).toHaveCount(beforeCount + 1, { timeout: 30_000 })
+
+    await expect.poll(() => page.url()).not.toContain("#message-")
+    await expect.poll(async () => (await timelineMetrics(page)).distanceFromBottom, { timeout: 30_000 }).toBeLessThan(40)
+    const rendered = await page.locator(sessionMessageItemSelector).evaluateAll((items) =>
+      items.map((item) => (item instanceof HTMLElement ? item.dataset.messageId : undefined)).filter(Boolean),
+    )
+    expect(rendered.at(-1)).not.toBe(oldID)
+  })
+})
+
+test("renders the full initial session window when switching sessions", async ({ page, project }) => {
+  test.setTimeout(120_000)
+
+  await project.open()
+  const sdk = project.sdk
+
+  await withSession(sdk, `e2e switch source ${Date.now()}`, async (first) => {
+    project.trackSession(first.id)
+    await withSession(sdk, `e2e switch target ${Date.now()}`, async (second) => {
+      project.trackSession(second.id)
+      await seedSessionTurns({ sdk, sessionID: first.id, count: 14 })
+      await seedSessionTurns({ sdk, sessionID: second.id, count: 14 })
+
+      await project.gotoSession(first.id)
+      await expect(page.locator(sessionMessageItemSelector)).toHaveCount(10, { timeout: 30_000 })
+      await expect(page.locator(sessionItemSelector(second.id))).toBeVisible({ timeout: 30_000 })
+
+      await installMessageCountProbe(page)
+      await page.locator(sessionItemSelector(second.id)).click()
+      await expect(page).toHaveURL(new RegExp(`/session/${second.id}(?:[?#]|$)`))
+      await expect(page.locator(sessionMessageItemSelector)).toHaveCount(10, { timeout: 30_000 })
+      const samples = await stopMessageCountProbe(page)
+      const rendered = await page.locator(sessionMessageItemSelector).evaluateAll((items) =>
+        items.map((item) => (item instanceof HTMLElement ? item.dataset.messageId : undefined)).filter(Boolean),
+      )
+      const secondMessages = await sdk.session.messages({ sessionID: second.id, limit: 100 }).then((r) => r.data ?? [])
+      const secondIDs = new Set(secondMessages.filter((item) => item.info.role === "user").map((item) => item.info.id))
+
+      const switched = samples.filter((sample) => sample.url.includes(`/session/${second.id}`))
+      expect(switched.length).toBeGreaterThan(0)
+      expect(rendered.every((id) => secondIDs.has(id))).toBe(true)
+      expect(switched.filter((sample) => sample.messages > 0 && sample.messages < 10)).toEqual([])
+    })
+  })
+})
--- a/packages/app/e2e/session/session-scroll-position.spec.ts
+++ b/packages/app/e2e/session/session-scroll-position.spec.ts
@@ -6,18 +6,33 @@ import { createSdk } from "../utils"
 
 type Sdk = ReturnType<typeof createSdk>
 
+const INITIAL_SESSION_WINDOW_MESSAGES = 10
+
+type TimelineMetrics = {
+  top: number
+  height: number
+  client: number
+  distanceFromBottom: number
+}
+
 function timelineMetrics(page: Page) {
   return page.evaluate((turnListSelector) => {
     const list = document.querySelector(turnListSelector)
     const viewport = list?.closest(".scroll-view__viewport")
-    if (!(viewport instanceof HTMLElement)) throw new Error("session timeline viewport not found")
+    if (!(viewport instanceof HTMLElement)) return null
     return {
       top: viewport.scrollTop,
       height: viewport.scrollHeight,
       client: viewport.clientHeight,
       distanceFromBottom: viewport.scrollHeight - viewport.clientHeight - viewport.scrollTop,
     }
-  }, sessionTurnListSelector)
+  }, sessionTurnListSelector) as Promise<TimelineMetrics | null>
+}
+
+async function expectTimelineMetrics(page: Page) {
+  const metrics = await timelineMetrics(page)
+  expect(metrics, "session timeline viewport should exist").not.toBeNull()
+  return metrics!
 }
 
 async function scrollTimelineToBottom(page: Page) {
@@ -120,9 +135,11 @@ test("keeps the latest turn in view when sending from an old message hash", asyn
     await seedSessionTurns({ sdk, sessionID: session.id, count: 14 })
 
     await project.gotoSession(session.id)
-    await expect(page.locator(sessionMessageItemSelector)).toHaveCount(10, { timeout: 30_000 })
+    await expect(page.locator(sessionMessageItemSelector)).toHaveCount(INITIAL_SESSION_WINDOW_MESSAGES, {
+      timeout: 30_000,
+    })
     await scrollTimelineToBottom(page)
-    await expect.poll(async () => (await timelineMetrics(page)).distanceFromBottom).toBeLessThan(20)
+    await expect.poll(async () => (await expectTimelineMetrics(page)).distanceFromBottom).toBeLessThan(20)
 
     const ids = await page.locator(sessionMessageItemSelector).evaluateAll((items) =>
       items.map((item) => (item instanceof HTMLElement ? item.dataset.messageId : undefined)).filter(Boolean),
@@ -132,7 +149,7 @@ test("keeps the latest turn in view when sending from an old message hash", asyn
 
     await page.goto(`${page.url()}#message-${oldID}`)
     await expect(page.locator(`#message-${oldID}`)).toBeVisible()
-    await expect.poll(async () => (await timelineMetrics(page)).distanceFromBottom).toBeGreaterThan(100)
+    await expect.poll(async () => (await expectTimelineMetrics(page)).distanceFromBottom).toBeGreaterThan(100)
 
     const token = `scroll_latest_${Date.now()}`
     const beforeCount = await page.locator(sessionMessageItemSelector).count()
@@ -140,7 +157,9 @@ test("keeps the latest turn in view when sending from an old message hash", asyn
     await expect(page.locator(sessionMessageItemSelector)).toHaveCount(beforeCount + 1, { timeout: 30_000 })
 
     await expect.poll(() => page.url()).not.toContain("#message-")
-    await expect.poll(async () => (await timelineMetrics(page)).distanceFromBottom, { timeout: 30_000 }).toBeLessThan(40)
+    await expect
+      .poll(async () => (await expectTimelineMetrics(page)).distanceFromBottom, { timeout: 30_000 })
+      .toBeLessThan(40)
     const rendered = await page.locator(sessionMessageItemSelector).evaluateAll((items) =>
       items.map((item) => (item instanceof HTMLElement ? item.dataset.messageId : undefined)).filter(Boolean),
     )
@@ -162,24 +181,32 @@ test("renders the full initial session window when switching sessions", async ({
       await seedSessionTurns({ sdk, sessionID: second.id, count: 14 })
 
       await project.gotoSession(first.id)
-      await expect(page.locator(sessionMessageItemSelector)).toHaveCount(10, { timeout: 30_000 })
+      await expect(page.locator(sessionMessageItemSelector)).toHaveCount(INITIAL_SESSION_WINDOW_MESSAGES, {
+        timeout: 30_000,
+      })
       await expect(page.locator(sessionItemSelector(second.id))).toBeVisible({ timeout: 30_000 })
 
       await installMessageCountProbe(page)
       await page.locator(sessionItemSelector(second.id)).click()
       await expect(page).toHaveURL(new RegExp(`/session/${second.id}(?:[?#]|$)`))
-      await expect(page.locator(sessionMessageItemSelector)).toHaveCount(10, { timeout: 30_000 })
+      await expect(page.locator(sessionMessageItemSelector)).toHaveCount(INITIAL_SESSION_WINDOW_MESSAGES, {
+        timeout: 30_000,
+      })
       const samples = await stopMessageCountProbe(page)
       const rendered = await page.locator(sessionMessageItemSelector).evaluateAll((items) =>
         items.map((item) => (item instanceof HTMLElement ? item.dataset.messageId : undefined)).filter(Boolean),
       )
       const secondMessages = await sdk.session.messages({ sessionID: second.id, limit: 100 }).then((r) => r.data ?? [])
-      const secondIDs = new Set(secondMessages.filter((item) => item.info.role === "user").map((item) => item.info.id))
+      const secondIDs = new Set(secondMessages.map((item) => item.info.id))
 
       const switched = samples.filter((sample) => sample.url.includes(`/session/${second.id}`))
       expect(switched.length).toBeGreaterThan(0)
       expect(rendered.every((id) => secondIDs.has(id))).toBe(true)
-      expect(switched.filter((sample) => sample.messages > 0 && sample.messages < 10)).toEqual([])
+      expect(
+        switched.filter(
+          (sample) => sample.messages > 0 && sample.messages < INITIAL_SESSION_WINDOW_MESSAGES,
+        ),
+      ).toEqual([])
     })
   })
 })
--- a/packages/app/e2e/session/session-scroll-position.spec.ts
+++ b/packages/app/e2e/session/session-scroll-position.spec.ts
@@ -36,12 +36,14 @@ async function expectTimelineMetrics(page: Page) {
 }
 
 async function scrollTimelineToBottom(page: Page) {
-  await page.evaluate((turnListSelector) => {
+  const found = await page.evaluate((turnListSelector) => {
     const list = document.querySelector(turnListSelector)
     const viewport = list?.closest(".scroll-view__viewport")
-    if (!(viewport instanceof HTMLElement)) throw new Error("session timeline viewport not found")
+    if (!(viewport instanceof HTMLElement)) return false
     viewport.scrollTop = viewport.scrollHeight
+    return true
   }, sessionTurnListSelector)
+  expect(found, "session timeline viewport should exist").toBe(true)
 }
 
 async function sendVisiblePrompt(input: { page: Page; text: string }) {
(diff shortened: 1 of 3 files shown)


Astro-Han (commented):
## Review

### P0 — Potential correctness issue

**`use-session-hash-scroll.ts` — `clearing` flag not reset on hash change**

The new `else if (clearing) { return }` guard prevents re-applying a stale hash while `clearing === true`, but `clearing` is only ever set to `false` when `!hash`. If the user navigates to a *different* hash (e.g. `#message-A` → `#message-B`) while `clearing` is still `true`, the effect bails out and the new hash is never applied. The flag should probably be keyed to the specific hash being cleared, or reset whenever `hash` changes, not just when it becomes empty.

---

### P1 — Maintainability / test hygiene

**`session-scroll-position.spec.ts` — `seedSessionTurns` uses `noReply: true` then asserts on assistant message count**

`seedSessionTurns` seeds 14 user turns with `noReply: true`, so there are no assistant messages. The test then asserts `toHaveCount(10)` on `sessionMessageItemSelector` (which matches `[data-message-id]`). This works only because the selector matches *user* messages too, but the count "10" is the capped window size, not a message-type count. A comment or a more explicit assertion (e.g. counting only `.user` or `.assistant` roles) would make the intent obvious and prevent future breakage if the selector scope changes.

**`session-scroll-position.spec.ts` — `secondIDs` filters on `role === "user"` but `rendered` includes all messages**

The test asserts `rendered.every((id) => secondIDs.has(id))`. Since `secondIDs` is built from user messages only, this will fail if any assistant message is rendered in the window. The test happens to pass because the 10-turn window ends on a user message boundary, but this coupling is fragile. Assert against all message IDs, not just user ones.

**`session-send-stability.spec.ts` — `MAX_STABILITY_SAMPLES = 256` is a magic number with no explanation**

Why 256? If a test generates more than 256 mutations/samplings, the probe silently drops data. A brief comment on how this bound was chosen (e.g. "empirically ~120 samples for a 14-turn session switch") would help future maintainers know whether to bump it.

---

### P2 — Code clarity / defensive programming

**`use-session-hash-scroll.ts` — `clearing` is a module-level mutable flag in a hook factory**

`let clearing = false` lives outside the reactive system. Solid effects can re-run out of order or be batched; a plain `let` flag is prone to stale-state bugs if multiple rapid hash changes interleave. Consider whether `clearing` should be a signal or at least scoped to a ref object so its lifetime is explicit.

**`session-scroll-position.spec.ts` — `timelineMetrics` throws if viewport not found**

`throw new Error("session timeline viewport not found")` inside `page.evaluate` produces a generic Playwright error with no stack context. Prefer returning a sentinel (e.g. `null`) and asserting on it in the test body so the failure message points to the right line.

**`session-scroll-position.spec.ts` — `installMessageCountProbe` leaks `maxSamples` into page scope**

The probe object is injected with `{ maxSamples: 256, messageSelector: ... }`. `maxSamples` is not used inside the evaluated function — the closure already captures `maxSamples` from the outer scope. The parameter is dead weight and confusing; remove it or use it.

**`session-send-stability.spec.ts` — `recordRemovedMount` mutates `unmountCount` even when bounded**

`pushBounded` silently drops samples once `maxSamples` is reached, but `unmountCount` keeps incrementing. This means `unmountCount` can exceed `unmounts.length`, which is surprising. Either cap `unmountCount` alongside the sample list or document the intentional divergence.

---

### P3 — Nitpicks

**`message-timeline.tsx` — `stageCfg` change from `{ init: 1, batch: 3 }` to `{ init: 10, batch: 3 }` is unexplained in code**

The JSDoc on `createTimelineStaging` says "Defer-mounts small timeline windows so revealing older turns does not block first paint". Bumping `init` from 1 to 10 means the first paint now mounts up to 10 turns immediately. This trades first-paint speed for no-flicker switching. A one-line comment on `stageCfg` explaining the trade-off ("init = window cap so the full visible window renders in one frame") would save the next reader a git-blame trip.

**`session-scroll-position.spec.ts` — `Date.now()` in session titles makes debugging flaky failures harder**

If a test flakes, the session title is different on every run, so correlating logs across retries is painful. Consider a deterministic suffix (e.g. `test.info().title` or a counter) or at least logging the generated title before use.

**`session-send-stability.spec.ts` — `page.keyboard.type` vs `page.keyboard.insertText` inconsistency**

`session-scroll-position.spec.ts` uses `insertText` while `session-send-stability.spec.ts` uses `type`. They behave differently (`type` fires `keydown`/`keyup`, `insertText` does not). If the app specifically needs key events for the composer, `type` is correct; if not, standardise on one. Either way, a comment explaining the choice would help.

Astro-Han (commented):
## Review

### P0 — Potential correctness issue

**`use-session-hash-scroll.ts` — `clearing` flag not reset on hash change**

The new `else if (clearing) { return }` guard prevents re-applying a stale hash while `clearing === true`, but `clearing` is only ever set to `false` when `!hash`. If the user navigates to a *different* hash (e.g. `#message-A` → `#message-B`) while `clearing` is still `true`, the effect bails out and the new hash is never applied. The flag should probably be keyed to the specific hash being cleared, or reset whenever `hash` changes, not just when it becomes empty.

---

### P1 — Maintainability / test hygiene

**`session-scroll-position.spec.ts` — `seedSessionTurns` uses `noReply: true` then asserts on assistant message count**

`seedSessionTurns` seeds 14 user turns with `noReply: true`, so there are no assistant messages. The test then asserts `toHaveCount(10)` on `sessionMessageItemSelector` (which matches `[data-message-id]`). This works only because the selector matches *user* messages too, but the count "10" is the capped window size, not a message-type count. A comment or a more explicit assertion (e.g. counting only `.user` or `.assistant` roles) would make the intent obvious and prevent future breakage if the selector scope changes.

**`session-scroll-position.spec.ts` — `secondIDs` filters on `role === \"user\"` but `rendered` includes all messages**

The test asserts `rendered.every((id) => secondIDs.has(id))`. Since `secondIDs` is built from user messages only, this will fail if any assistant message is rendered in the window. The test happens to pass because the 10-turn window ends on a user message boundary, but this coupling is fragile. Assert against all message IDs, not just user ones.

**`session-send-stability.spec.ts` — `MAX_STABILITY_SAMPLES = 256` is a magic number with no explanation**

Why 256? If a test generates more than 256 mutations/samplings, the probe silently drops data. A brief comment on how this bound was chosen (e.g. "empirically ~120 samples for a 14-turn session switch") would help future maintainers know whether to bump it.

---

### P2 — Code clarity / defensive programming

**`use-session-hash-scroll.ts` — `clearing` is a module-level mutable flag in a hook factory**

`let clearing = false` lives outside the reactive system. Solid effects can re-run out of order or be batched; a plain `let` flag is prone to stale-state bugs if multiple rapid hash changes interleave. Consider whether `clearing` should be a signal or at least scoped to a ref object so its lifetime is explicit.

**`session-scroll-position.spec.ts` — `timelineMetrics` throws if viewport not found**

`throw new Error("session timeline viewport not found")` inside `page.evaluate` produces a generic Playwright error with no stack context. Prefer returning a sentinel (e.g. `null`) and asserting on it in the test body so the failure message points to the right line.

**`session-scroll-position.spec.ts` — `installMessageCountProbe` leaks `maxSamples` into page scope**

The probe object is injected with `{ maxSamples: 256, messageSelector: ... }`. `maxSamples` is not used inside the evaluated function — the closure already captures `maxSamples` from the outer scope. The parameter is dead weight and confusing; remove it or use it.

**`session-send-stability.spec.ts` — `recordRemovedMount` mutates `unmountCount` even when bounded**

`pushBounded` silently drops samples once `maxSamples` is reached, but `unmountCount` keeps incrementing. This means `unmountCount` can exceed `unmounts.length`, which is surprising. Either cap `unmountCount` alongside the sample list or document the intentional divergence.

---

### P3 — Nitpicks

**`message-timeline.tsx` — `stageCfg` change from `{ init: 1, batch: 3 }` to `{ init: 10, batch: 3 }` is unexplained in code**

The JSDoc on `createTimelineStaging` says "Defer-mounts small timeline windows so revealing older turns does not block first paint". Bumping `init` from 1 to 10 means the first paint now mounts up to 10 turns immediately. This trades first-paint speed for no-flicker switching. A one-line comment on `stageCfg` explaining the trade-off ("init = window cap so the full visible window renders in one frame") would save the next reader a git-blame trip.

**`session-scroll-position.spec.ts` — `Date.now()` in session titles makes debugging flaky failures harder**

If a test flakes, the session title is different on every run, so correlating logs across retries is painful. Consider a deterministic suffix (e.g. `test.info().title` or a counter) or at least logging the generated title before use.

**`session-send-stability.spec.ts` — `page.keyboard.type` vs `page.keyboard.insertText` inconsistency**

`session-scroll-position.spec.ts` uses `insertText` while `session-send-stability.spec.ts` uses `type`. They behave differently (`type` fires `keydown`/`keyup`, `insertText` does not). If the app specifically needs key events for the composer, `type` is correct; if not, standardise on one. Either way, a comment explaining the choice would help.

Astro-Han (commented):
## Review

### P0 — Potential correctness issue

**`use-session-hash-scroll.ts` — `clearing` flag not reset on hash change**

The new `else if (clearing) { return }` guard prevents re-applying a stale hash while `clearing === true`, but `clearing` is only ever set to `false` when `!hash`. If the user navigates to a *different* hash (e.g. `#message-A` → `#message-B`) while `clearing` is still `true`, the effect bails out and the new hash is never applied. The flag should probably be keyed to the specific hash being cleared, or reset whenever `hash` changes, not just when it becomes empty.

---

### P1 — Maintainability / test hygiene

**`session-scroll-position.spec.ts` — `seedSessionTurns` uses `noReply: true` then asserts on assistant message count**

`seedSessionTurns` seeds 14 user turns with `noReply: true`, so there are no assistant messages. The test then asserts `toHaveCount(10)` on `sessionMessageItemSelector` (which matches `[data-message-id]`). This works only because the selector matches *user* messages too, but the count "10" is the capped window size, not a message-type count. A comment or a more explicit assertion (e.g. counting only `.user` or `.assistant` roles) would make the intent obvious and prevent future breakage if the selector scope changes.

**`session-scroll-position.spec.ts` — `secondIDs` filters on `role === \"user\"` but `rendered` includes all messages**

The test asserts `rendered.every((id) => secondIDs.has(id))`. Since `secondIDs` is built from user messages only, this will fail if any assistant message is rendered in the window. The test happens to pass because the 10-turn window ends on a user message boundary, but this coupling is fragile. Assert against all message IDs, not just user ones.

**`session-send-stability.spec.ts` — `MAX_STABILITY_SAMPLES = 256` is a magic number with no explanation**

Why 256? If a test generates more than 256 mutations/samplings, the probe silently drops data. A brief comment on how this bound was chosen (e.g. "empirically ~120 samples for a 14-turn session switch") would help future maintainers know whether to bump it.

---

### P2 — Code clarity / defensive programming

**`use-session-hash-scroll.ts` — `clearing` is a module-level mutable flag in a hook factory**

`let clearing = false` lives outside the reactive system. Solid effects can re-run out of order or be batched; a plain `let` flag is prone to stale-state bugs if multiple rapid hash changes interleave. Consider whether `clearing` should be a signal or at least scoped to a ref object so its lifetime is explicit.

**`session-scroll-position.spec.ts` — `timelineMetrics` throws if viewport not found**

`throw new Error("session timeline viewport not found")` inside `page.evaluate` produces a generic Playwright error with no stack context. Prefer returning a sentinel (e.g. `null`) and asserting on it in the test body so the failure message points to the right line.

**`session-scroll-position.spec.ts` — `installMessageCountProbe` leaks `maxSamples` into page scope**

The probe object is injected with `{ maxSamples: 256, messageSelector: ... }`. `maxSamples` is not used inside the evaluated function — the closure already captures `maxSamples` from the outer scope. The parameter is dead weight and confusing; remove it or use it.

**`session-send-stability.spec.ts` — `recordRemovedMount` mutates `unmountCount` even when bounded**

`pushBounded` silently drops samples once `maxSamples` is reached, but `unmountCount` keeps incrementing. This means `unmountCount` can exceed `unmounts.length`, which is surprising. Either cap `unmountCount` alongside the sample list or document the intentional divergence.

---

### P3 — Nitpicks

**`message-timeline.tsx` — `stageCfg` change from `{ init: 1, batch: 3 }` to `{ init: 10, batch: 3 }` is unexplained in code**

The JSDoc on `createTimelineStaging` says "Defer-mounts small timeline windows so revealing older turns does not block first paint". Bumping `init` from 1 to 10 means the first paint now mounts up to 10 turns immediately. This trades first-paint speed for no-flicker switching. A one-line comment on `stageCfg` explaining the trade-off ("init = window cap so the full visible window renders in one frame") would save the next reader a git-blame trip.

**`session-scroll-position.spec.ts` — `Date.now()` in session titles makes debugging flaky failures harder**

If a test flakes, the session title is different on every run, so correlating logs across retries is painful. Consider a deterministic suffix (e.g. `test.info().title` or a counter) or at least logging the generated title before use.

**`session-send-stability.spec.ts` — `page.keyboard.type` vs `page.keyboard.insertText` inconsistency**

`session-scroll-position.spec.ts` uses `insertText` while `session-send-stability.spec.ts` uses `type`. They behave differently (`type` fires `keydown`/`keyup`, `insertText` does not). If the app specifically needs key events for the composer, `type` is correct; if not, standardise on one. Either way, a comment explaining the choice would help.

Astro-Han (commented):
## Review

### P0 — Potential correctness issue

**`use-session-hash-scroll.ts` — `clearing` flag not reset on hash change**

The new `else if (clearing) { return }` guard prevents re-applying a stale hash while `clearing === true`, but `clearing` is only ever set to `false` when `!hash`. If the user navigates to a *different* hash (e.g. `#message-A` → `#message-B`) while `clearing` is still `true`, the effect bails out and the new hash is never applied. The flag should probably be keyed to the specific hash being cleared, or reset whenever `hash` changes, not just when it becomes empty.

---

### P1 — Maintainability / test hygiene

**`session-scroll-position.spec.ts` — `seedSessionTurns` uses `noReply: true` then asserts on assistant message count**

`seedSessionTurns` seeds 14 user turns with `noReply: true`, so there are no assistant messages. The test then asserts `toHaveCount(10)` on `sessionMessageItemSelector` (which matches `[data-message-id]`). This works only because the selector matches *user* messages too, but the count "10" is the capped window size, not a message-type count. A comment or a more explicit assertion (e.g. counting only `.user` or `.assistant` roles) would make the intent obvious and prevent future breakage if the selector scope changes.

**`session-scroll-position.spec.ts` — `secondIDs` filters on `role === "user"` but `rendered` includes all messages**

The test asserts `rendered.every((id) => secondIDs.has(id))`. Since `secondIDs` is built from user messages only, this will fail if any assistant message is rendered in the window. The test happens to pass because the 10-turn window ends on a user message boundary, but this coupling is fragile. Assert against all message IDs, not just user ones.

**`session-send-stability.spec.ts` — `MAX_STABILITY_SAMPLES = 256` is a magic number with no explanation**

Why 256? If a test generates more than 256 mutations/samplings, the probe silently drops data. A brief comment on how this bound was chosen (e.g. "empirically ~120 samples for a 14-turn session switch") would help future maintainers know whether to bump it.

---

### P2 — Code clarity / defensive programming

**`use-session-hash-scroll.ts` — `clearing` is a module-level mutable flag in a hook factory**

`let clearing = false` lives outside the reactive system. Solid effects can re-run out of order or be batched; a plain `let` flag is prone to stale-state bugs if multiple rapid hash changes interleave. Consider whether `clearing` should be a signal or at least scoped to a ref object so its lifetime is explicit.

**`session-scroll-position.spec.ts` — `timelineMetrics` throws if viewport not found**

`throw new Error("session timeline viewport not found")` inside `page.evaluate` produces a generic Playwright error with no stack context. Prefer returning a sentinel (e.g. `null`) and asserting on it in the test body so the failure message points to the right line.

**`session-scroll-position.spec.ts` — `installMessageCountProbe` leaks `maxSamples` into page scope**

The probe object is injected with `{ maxSamples: 256, messageSelector: ... }`. `maxSamples` is not used inside the evaluated function — the closure already captures `maxSamples` from the outer scope. The parameter is dead weight and confusing; remove it or use it.

**`session-send-stability.spec.ts` — `recordRemovedMount` mutates `unmountCount` even when bounded**

`pushBounded` silently drops samples once `maxSamples` is reached, but `unmountCount` keeps incrementing. This means `unmountCount` can exceed `unmounts.length`, which is surprising. Either cap `unmountCount` alongside the sample list or document the intentional divergence.

---

### P3 — Nitpicks

**`message-timeline.tsx` — `stageCfg` change from `{ init: 1, batch: 3 }` to `{ init: 10, batch: 3 }` is unexplained in code**

The JSDoc on `createTimelineStaging` says "Defer-mounts small timeline windows so revealing older turns does not block first paint". Bumping `init` from 1 to 10 means the first paint now mounts up to 10 turns immediately. This trades first-paint speed for no-flicker switching. A one-line comment on `stageCfg` explaining the trade-off ("init = window cap so the full visible window renders in one frame") would save the next reader a git-blame trip.

**`session-scroll-position.spec.ts` — `Date.now()` in session titles makes debugging flaky failures harder**

If a test flakes, the session title is different on every run, so correlating logs across retries is painful. Consider a deterministic suffix (e.g. `test.info().title` or a counter) or at least logging the generated title before use.

**`session-send-stability.spec.ts` — `page.keyboard.type` vs `page.keyboard.insertText` inconsistency**

`session-scroll-position.spec.ts` uses `insertText` while `session-send-stability.spec.ts` uses `type`. They behave differently (`type` fires `keydown`/`keyup`, `insertText` does not). If the app specifically needs key events for the composer, `type` is correct; if not, standardise on one. Either way, a comment explaining the choice would help.

Astro-Han (commented):
## Review

### P0 — Potential correctness issue

**`use-session-hash-scroll.ts` — `clearing` flag not reset on hash change**

The new `else if (clearing) { return }` guard prevents re-applying a stale hash while `clearing === true`, but `clearing` is only ever set to `false` when `!hash`. If the user navigates to a *different* hash (e.g. `#message-A` → `#message-B`) while `clearing` is still `true`, the effect bails out and the new hash is never applied. The flag should probably be keyed to the specific hash being cleared, or reset whenever `hash` changes, not just when it becomes empty.

---

### P1 — Maintainability / test hygiene

**`session-scroll-position.spec.ts` — `seedSessionTurns` uses `noReply: true` then asserts on assistant message count**

`seedSessionTurns` seeds 14 user turns with `noReply: true`, so there are no assistant messages. The test then asserts `toHaveCount(10)` on `sessionMessageItemSelector` (which matches `[data-message-id]`). This works only because the selector matches *user* messages too, but the count "10" is the capped window size, not a message-type count. A comment or a more explicit assertion (e.g. counting only `.user` or `.assistant` roles) would make the intent obvious and prevent future breakage if the selector scope changes.

**`session-scroll-position.spec.ts` — `secondIDs` filters on `role === "user"` but `rendered` includes all messages**

The test asserts `rendered.every((id) => secondIDs.has(id))`. Since `secondIDs` is built from user messages only, this will fail if any assistant message is rendered in the window. The test happens to pass because the 10-turn window ends on a user message boundary, but this coupling is fragile. Assert against all message IDs, not just user ones.

**`session-send-stability.spec.ts` — `MAX_STABILITY_SAMPLES = 256` is a magic number with no explanation**

Why 256? If a test generates more than 256 mutations/samplings, the probe silently drops data. A brief comment on how this bound was chosen (e.g. "empirically ~120 samples for a 14-turn session switch") would help future maintainers know whether to bump it.

---

### P2 — Code clarity / defensive programming

**`use-session-hash-scroll.ts` — `clearing` is a module-level mutable flag in a hook factory**

`let clearing = false` lives outside the reactive system. Solid effects can re-run out of order or be batched; a plain `let` flag is prone to stale-state bugs if multiple rapid hash changes interleave. Consider whether `clearing` should be a signal or at least scoped to a ref object so its lifetime is explicit.

**`session-scroll-position.spec.ts` — `timelineMetrics` throws if viewport not found**

`throw new Error("session timeline viewport not found")` inside `page.evaluate` produces a generic Playwright error with no stack context. Prefer returning a sentinel (e.g. `null`) and asserting on it in the test body so the failure message points to the right line.

**`session-scroll-position.spec.ts` — `installMessageCountProbe` leaks `maxSamples` into page scope**

The probe object is injected with `{ maxSamples: 256, messageSelector: ... }`. `maxSamples` is not used inside the evaluated function — the closure already captures `maxSamples` from the outer scope. The parameter is dead weight and confusing; remove it or use it.

**`session-send-stability.spec.ts` — `recordRemovedMount` mutates `unmountCount` even when bounded**

`pushBounded` silently drops samples once `maxSamples` is reached, but `unmountCount` keeps incrementing. This means `unmountCount` can exceed `unmounts.length`, which is surprising. Either cap `unmountCount` alongside the sample list or document the intentional divergence.

---

### P3 — Nitpicks

**`message-timeline.tsx` — `stageCfg` change from `{ init: 1, batch: 3 }` to `{ init: 10, batch: 3 }` is unexplained in code**

The JSDoc on `createTimelineStaging` says "Defer-mounts small timeline windows so revealing older turns does not block first paint". Bumping `init` from 1 to 10 means the first paint now mounts up to 10 turns immediately. This trades first-paint speed for no-flicker switching. A one-line comment on `stageCfg` explaining the trade-off ("init = window cap so the full visible window renders in one frame") would save the next reader a git-blame trip.

**`session-scroll-position.spec.ts` — `Date.now()` in session titles makes debugging flaky failures harder**

If a test flakes, the session title is different on every run, so correlating logs across retries is painful. Consider a deterministic suffix (e.g. `test.info().title` or a counter) or at least logging the generated title before use.

**`session-send-stability.spec.ts` — `page.keyboard.type` vs `page.keyboard.insertText` inconsistency**

`session-scroll-position.spec.ts` uses `insertText` while `session-send-stability.spec.ts` uses `type`. They behave differently (`type` fires `keydown`/`keyup`, `insertText` does not). If the app specifically needs key events for the composer, `type` is correct; if not, standardise on one. Either way, a comment explaining the choice would help.

packages/app/src/pages/session/use-session-hash-scroll.ts
```diff
@@ -144,7 +144,11 @@ export const useSessionHashScroll = (input: {
 
   createEffect(() => {
     const hash = location.hash
-    if (!hash) clearing = false
+    if (!hash) {
+      clearing = false
+    } else if (clearing) {
```
Astro-Han: **[P0]** The `clearing` flag is only reset when `!hash`. If the user navigates from `#message-A` to `#message-B` while `clearing` is still `true`, this effect bails out and the new hash is never applied. Consider keying the flag to the specific hash being cleared, or resetting it whenever `hash` changes.

packages/app/src/pages/session/message-timeline.tsx
```diff
@@ -378,7 +378,7 @@ export function MessageTimeline(props: {
     return language.t("command.session.new")
   })
   const showHeader = createMemo(() => !!(titleValue() || parentID()))
-  const stageCfg = { init: 1, batch: 3 }
+  const stageCfg = { init: 10, batch: 3 }
```
Astro-Han, line 382: **[P3]** The JSDoc says `init` defers mounts to avoid blocking first paint, but bumping from 1 to 10 means the first paint now mounts up to 10 turns immediately. A one-line comment explaining the trade-off (e.g. `init = window cap so the full visible window renders in one frame`) would save the next reader a git-blame trip.

packages/app/e2e/session/session-scroll-position.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+import { test, expect } from "../fixtures"
+import { withSession } from "../actions"
+import { promptSelector, sessionItemSelector, sessionMessageItemSelector, sessionTurnListSelector } from "../selectors"
+import { createSdk } from "../utils"
+
+type Sdk = ReturnType<typeof createSdk>
+
+function timelineMetrics(page: Page) {
+  return page.evaluate((turnListSelector) => {
+    const list = document.querySelector(turnListSelector)
```
Astro-Han, line 20: **[P2]** `throw` inside `page.evaluate` produces a generic Playwright error with no stack context. Prefer returning `null` and asserting on it in the test body so failures point to the right line.

packages/app/e2e/session/session-scroll-position.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+      parts: [
+        {
+          type: "text",
+          text: `seed turn ${i}\n${Array.from({ length: 16 }, (_, line) => `line ${line} ${"content ".repeat(8)}`).join("\n")}`,
+        },
+      ],
+    })
+  }
+}
+
```
Astro-Han, line 72: **[P2]** `maxSamples` is passed into `page.evaluate` but never used inside — the closure already captures it from the outer scope. Either remove the dead parameter or use it consistently.

packages/app/e2e/session/session-scroll-position.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+async function seedSessionTurns(input: { sdk: Sdk; sessionID: string; count: number }) {
+  for (let i = 0; i < input.count; i++) {
+    await input.sdk.session.promptAsync({
+      sessionID: input.sessionID,
+      noReply: true,
+      parts: [
+        {
+          type: "text",
```
Astro-Han, line 65: **[P1]** `noReply: true` means no assistant messages are generated, yet the test later asserts `toHaveCount(10)` on `[data-message-id]` which matches user messages too. The count "10" is the window cap, not an assistant-message count. A comment or more explicit assertion would prevent future breakage if the selector scope changes.

packages/app/e2e/session/session-scroll-position.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  const sdk = project.sdk
+
+  await withSession(sdk, `e2e switch source ${Date.now()}`, async (first) => {
+    project.trackSession(first.id)
+    await withSession(sdk, `e2e switch target ${Date.now()}`, async (second) => {
+      project.trackSession(second.id)
+      await seedSessionTurns({ sdk, sessionID: first.id, count: 14 })
+      await seedSessionTurns({ sdk, sessionID: second.id, count: 14 })
+
+      await project.gotoSession(first.id)
```
Astro-Han, line 185: **[P1]** `secondIDs` is built from `role === "user"` messages only, but `rendered` includes all messages. This will fail if any assistant message appears in the window. Assert against all message IDs, not just user ones.

packages/app/e2e/session/session-scroll-position.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+test("keeps the latest turn in view when sending from an old message hash", async ({ page, project }) => {
+  test.setTimeout(120_000)
+
+  await project.open()
+  const sdk = project.sdk
+
+  await withSession(sdk, `e2e scroll position ${Date.now()}`, async (session) => {
+    project.trackSession(session.id)
+
+    await seedSessionTurns({ sdk, sessionID: session.id, count: 14 })
```
Astro-Han, line 137: **[P3]** `Date.now()` in session titles makes debugging flaky failures harder — the title is different on every run, so correlating logs across retries is painful. Consider a deterministic suffix or logging the title before use.

packages/app/e2e/session/session-scroll-position.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  }, sessionTurnListSelector)
+}
+
+async function sendVisiblePrompt(input: { page: Page; text: string }) {
+  const prompt = input.page.locator(promptSelector)
+  await expect(prompt).toBeVisible()
+  await prompt.click()
+  await input.page.keyboard.insertText(input.text)
+  await expect.poll(async () => (await prompt.textContent())?.replace(/\u200B/g, "").trim()).toBe(input.text)
+  await input.page.keyboard.press("Enter")
```
Astro-Han, line 55: **[P3]** `insertText` does not fire `keydown`/`keyup` events. If the composer specifically needs key events, this is fine; otherwise consider standardising on one approach across test files. A comment explaining the choice would help.

Astro-Han (commented):
## Follow-up — items that couldn't be attached as line comments

**`session-send-stability.spec.ts` — `MAX_STABILITY_SAMPLES = 256` [P1]**

This file is not in the PR diff, so a line comment isn't possible. The bound is unexplained; if a test generates more than 256 mutations the probe silently drops data. A brief comment on how this was chosen (e.g. "empirically ~120 samples for a 14-turn session switch") would help future maintainers.

**`session-send-stability.spec.ts` — `recordRemovedMount` mutates `unmountCount` even when bounded [P2]**

`pushBounded` silently drops samples once `maxSamples` is reached, but `unmountCount` keeps incrementing. This means `unmountCount` can exceed `unmounts.length`, which is surprising. Either cap `unmountCount` alongside the sample list or document the intentional divergence.

**`use-session-hash-scroll.ts` — `let clearing = false` outside the reactive system [P2]**

This line is outside the PR diff hunk, so a line comment isn't possible. A plain `let` flag is prone to stale-state bugs when Solid effects re-run out of order or are batched. Consider a signal or a ref object so the lifetime is explicit.

packages/app/src/pages/session/use-session-hash-scroll.ts
```diff
@@ -144,7 +144,11 @@ export const useSessionHashScroll = (input: {
...
+      return
+    }
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `clearMessageHash` now records the exact `clearingHash`; only that hash is suppressed while it is being cleared, and any different hash resets the guard and proceeds through normal hash scrolling.

packages/app/src/pages/session/use-session-hash-scroll.ts
Astro-Han: Fixed. The guard is now keyed to the specific hash being cleared instead of a boolean, so `#message-A` clearing does not block a later `#message-B` navigation.

packages/app/src/pages/session/message-timeline.tsx
Astro-Han: Fixed. Added the narrow comment explaining that `init` matches the initial window cap so session switches do not reveal partial batches.

packages/app/e2e/session/session-scroll-position.spec.ts
Astro-Han: Fixed. `timelineMetrics` now returns `null` when the viewport is missing, and the test asserts that condition with a named Playwright expectation before reading metrics.

packages/app/e2e/session/session-scroll-position.spec.ts
Astro-Han: Keeping this as is. The `page.evaluate` callback runs in the browser context and cannot capture the Node-side `maxSamples` binding; the destructured `maxSamples` parameter is used by `push` to bound samples.

packages/app/e2e/session/session-scroll-position.spec.ts
Astro-Han: Fixed. Replaced the literal count with `INITIAL_SESSION_WINDOW_MESSAGES`, making the assertion about the window cap rather than assistant-message count.

packages/app/e2e/session/session-scroll-position.spec.ts
Astro-Han: Fixed. The expected id set now includes all messages from the target session instead of filtering to user messages only.

packages/app/e2e/session/session-scroll-position.spec.ts
Astro-Han: Keeping this as is. The title is not the test identifier, and the generated session id is what the test logs and selectors use. The timestamp keeps session titles distinct during retries or local reused state without adding logging noise.

packages/app/e2e/session/session-scroll-position.spec.ts
Astro-Han: Keeping this as is. The test only needs key events for submit, which is covered by `keyboard.press("Enter")`; `insertText` keeps text entry deterministic for the contenteditable prompt.

Astro-Han:
Follow-up outcomes for non-inline items:

- `session-send-stability.spec.ts` sample bound and `unmountCount`: left unchanged. This file is outside the PR diff, and the current shape intentionally separates total detected unmounts from the bounded sample payload. Changing it would expand this PR beyond the session scroll fix.
- `use-session-hash-scroll.ts` local mutable guard: kept local, but narrowed from a boolean to `clearingHash`. It is scoped to one hook instance and only gates hash effects; making it a signal would add reactive state without a current caller benefit.

packages/app/e2e/session/session-scroll-position.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+
+      await project.gotoSession(first.id)
+      await expect(page.locator(sessionMessageItemSelector)).toHaveCount(INITIAL_SESSION_WINDOW_MESSAGES, {
+        timeout: 30_000,
+      })
+      await expect(page.locator(sessionItemSelector(second.id))).toBeVisible({ timeout: 30_000 })
+
+      await installMessageCountProbe(page)
+      await page.locator(sessionItemSelector(second.id)).click()
+      await expect(page).toHaveURL(new RegExp(`/session/${second.id}(?:[?#]|$)`))
```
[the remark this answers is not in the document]
Astro-Han: Fixed by removing the dynamic regex entirely. The assertion now checks the URL pathname with `endsWith`, so session IDs are treated as literal path text.

Astro-Han:
Non-inline CodeRabbit outcomes:

- `scrollTimelineToBottom`: fixed with the same sentinel pattern as `timelineMetrics`.
- Docstring coverage warning: not accepted. This PR follows existing TypeScript E2E style; adding docstrings would be unrelated churn.
- Suggested `ui` label: not accepted. Existing `bug`, `app`, and `P2` labels satisfy the PR template and match the scope.

Issue #283, opened 2026-04-28
[Feature] Make agent tool lifecycle explicit, visible, and schedulable


Astro-Han:
## What task are you trying to do?

We want PawWork's `agent` tool to behave like a schedulable subagent lifecycle, not just a synchronous helper call that eventually returns a text block.

This matters when the main agent delegates slow or parallel work to one or more subagents. The main agent should be able to list active subagents, inspect their results, stop the ones it owns, and recover cleanly when a subagent is canceled, empty, failed, or slow.

## What happens today?

The current `agent` tool is mostly a foreground call: the parent session starts a child session and waits for a final text result. That leaves several product gaps:

- If a subagent is manually stopped, the parent may not receive an explicit `canceled_by_user` result.
- If a subagent completes without final text, the parent can see an empty result instead of an actionable state.
- If a foreground subagent runs for a long time, the parent can appear stuck.
- If multiple subagents exist, the model has no first-class way to list, inspect, or stop the subagents it launched.
- The main session card does not reliably show the latest subagent activity.
- Session exports do not yet provide a clear subagent lifecycle trail for debugging.

## What would a good result look like?

PawWork tracks each `agent` invocation as a structured `SubagentRun` attached to the parent session. The model can use `agent_list`, `agent_output`, and `agent_stop` to coordinate its own subagents. The UI stays simple: the existing agent card only shows the latest useful activity line.

Subagent cancellation, empty output, failure, foreground timeout warning, background completion, and result consumption are explicit states. Background completion injects a notification into the parent session so the main agent does not continue blindly.

## Which audience does this matter to most?

Both. Non-technical users get a reassuring experience where subagents visibly make progress and cancellation works. Technical users get debuggable session exports and predictable model-facing scheduling tools.

## Acceptance criteria

- `agent` supports `run_in_background?: boolean`, defaulting to foreground.
- A structured `SubagentRun` lifecycle is persisted on the parent session's agent tool metadata.
- Foreground and background subagents share the same lifecycle model.
- The model can call `agent_list`, `agent_output`, and `agent_stop`.
- Background subagent concurrency is limited to 5 per parent session.
- User cancellation returns `canceled_by_user`, never an empty result.
- Empty final output returns `completed_empty`, never an empty result.
- A foreground subagent running longer than 10 minutes emits a warning and wakes the parent, but is not killed automatically.
- Background terminal states inject a parent-session notification and trigger auto-resume when safe.
- The existing agent card shows the latest useful activity in one line.
[...]

PR #287, opened 2026-04-28
feat(agent): subagent lifecycle v1 (foreground-only)

Astro-Han:
## Summary

Implements subagent lifecycle v1 (foreground-only) per #283. The agent tool now exposes explicit terminal states (completed / completed_empty / failed / canceled_by_user), caps parallel dispatch at 5 per parent, and ships two new model-callable inspection tools (`agent_list`, `agent_output`) so a parent agent can ask "is my subagent still running, what is it doing, what did it return".

## Why

Today the agent tool returns an empty `<subagent_result>` on cancel and silently swallows model failures, with no way for the dispatching agent to introspect or recover. This change makes lifecycle explicit, queryable, and bounded so common subagent failure modes stop being invisible. Background execution (`run_in_background`, notification injection, abandoned sweep, nested subagents) is intentionally deferred to v1.1.

## Related Issue

#283 (v5 spec, comment id 4332342274).

## Architecture

- `SubtaskPart` (`packages/opencode/src/session/message-v2.ts`) extended with all lifecycle / progress / result fields. Every new field is `.optional()` or `.default(...)` so legacy rows decode as terminal-completed without rewrites.
- New `Session` columns `created_by_agent_tool` + `subagent_type` mark child sessions as agent-tool-created and tag the subagent type.
- New `SubagentRun` service (`packages/opencode/src/session/subagent-run.ts`) is the single writer for lifecycle fields. Parent slot reservation is gated by `Semaphore.makeUnsafe(1)` per parent at MAX_ACTIVE=5; per-row mutex serializes lifecycle writes.
- Single-writer guard via `Context.Reference<boolean>` in `subagent-run-context.ts`. `Session.updatePart` rejects non-writer SubtaskPart writes that mutate lifecycle fields with `SubagentRunGuardViolation`.
- `AgentTool.execute` restructured into a scoped 8-step flow with `Effect.addFinalizer` for `releaseSlot` + listener cleanup on every exit path, and an outer `Effect.catch` that finalizes a started SubtaskPart to `failed` so pre-prompt errors cannot strand the row in `running`.
- `AgentPromptOps.wasInterrupted(sessionID)` query (set by `SessionRunState.onInterrupt` firing on the child runner) deterministically distinguishes child-runner abort from natural completion or model failure.
- `agent_list` filters by lifecycle status (running / completed / completed_empty / failed / canceled_by_user / all_active / all). `agent_output` accepts XOR(`subagent_session_id`, `tool_call_id`), validates parent ownership and `createdByAgentTool`, and marks terminal rows consumed on first read.

## Spec deviations

Two implementable mechanisms in this branch differ slightly from the literal spec text; both preserve the spec's intent:

1. **Single-writer enforcement uses a shared `Context.Reference` + inline check, not a Layer wrapper.** The spec sketch was a circular layer dep; the Reference lives in its own module so both writers and the guard observe it without circularity.
2. **Direct child-abort detection uses an additive `AgentPromptOps.wasInterrupted(sessionID)` query, not a subscribe API.** The current `SessionRunState` exposes no such API; the per-Instance `Set` written from `loop()`'s `onInterrupt` arg gives equivalent semantics with smaller surface.

## Not in this PR (deferred / known limitations)

- **UI agent card rendering (Task 15 in plan).** `packages/app/` is mid-rewrite and does not yet render `SubtaskPart` at all. Lifecycle data is queryable via `agent_list` / `agent_output`; the visual card is a follow-up.
- **`reserveSlot` cap accuracy after host restart.** In-memory `activeCounts` resets on restart; persisted `running` rows are not re-counted. Stranded `running` rows belong to the v1.1 abandoned-sweep work.
- **Background dispatch (`run_in_background`, `agent_stop`, `block=true`, abandoned, notification injection, nested subagents).** All v1.1.

## How To Verify

\`\`\`bash
bun --cwd packages/opencode run typecheck
bun --cwd packages/opencode test test/session/subagent-run.test.ts \
  test/session/subagent-lifecycle-integration.test.ts \
  test/tool/agent.test.ts \
  test/session/prompt-effect.test.ts
\`\`\`

Targeted suites cover: cap rejection + FIFO contention, lock-ordering invariant, all five terminal states end-to-end, ring-eviction with lifecycle pinning + hard-20 cap, single-writer guard violation, resume validation (parent / createdByAgentTool / subagent_type), backward-compat decode of legacy SubtaskPart rows.

Repo-wide \`turbo test:ci\` left to PR CI.

## Checklist

- [x] I am targeting \`dev\`, and my PR title and commit messages use Conventional Commits in English

## Summary by CodeRabbit

* **New Features**
  * Subagent lifecycle: subtask status, capped event history, activity tracking, consumption semantics, interruption-aware prompts, resume safeguards, and per-parent concurrency limits
  * Two built-in tools: list subagents and fetch subagent results/transcripts

* **Tests**
  * New integration and unit suites covering lifecycle, persistence, concurrency, resume, error sanitization, and formatting

* **Chores**
  * Session schema extended with agent-created marker (default false) and subagent type (nullable)

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -239,6 +239,19 @@ export const CompactionPart = PartBase.extend({
 })
 export type CompactionPart = z.infer<typeof CompactionPart>
 
+export const SubtaskEvent = z.discriminatedUnion("type", [
+  z.object({ type: z.literal("started"), at: z.number() }),
+  z.object({ type: z.literal("tool_started"), tool: z.string(), label: z.string(), at: z.number() }),
+  z.object({ type: z.literal("tool_completed"), tool: z.string(), at: z.number() }),
+  z.object({ type: z.literal("model_wait_started"), at: z.number() }),
+  z.object({ type: z.literal("completed"), at: z.number() }),
+  z.object({ type: z.literal("completed_empty"), at: z.number() }),
+  z.object({ type: z.literal("canceled_by_user"), at: z.number() }),
+  z.object({ type: z.literal("failed"), kind: z.string(), at: z.number() }),
+  z.object({ type: z.literal("consumed"), at: z.number() }),
+])
+export type SubtaskEvent = z.infer<typeof SubtaskEvent>
+
 export const SubtaskPart = PartBase.extend({
   type: z.literal("subtask"),
   prompt: z.string(),
@@ -251,6 +264,34 @@ export const SubtaskPart = PartBase.extend({
     })
     .optional(),
   command: z.string().optional(),
+  // identity (NEW; .optional() so legacy decode succeeds without a primary key)
+  tool_call_id: z.string().optional(),
+  parent_session_id: z.string().optional(),
+  parent_message_id: z.string().optional(),
+  subagent_session_id: z.string().optional(),
+  // lifecycle (NEW)
+  status: z
+    .enum(["running", "completed", "completed_empty", "failed", "canceled_by_user"])
+    .default("completed"),
+  started_at: z.number().optional(),
+  updated_at: z.number().optional(),
+  ended_at: z.number().optional(),
+  consumed_at: z.number().optional(),
+  // progress (NEW)
+  last_activity: z
+    .object({
+      kind: z.enum(["model_waiting", "thinking", "tool"]),
+      label: z.string(),
+      tool: z.string().optional(),
+      at: z.number(),
+    })
+    .optional(),
+  recent_events: z.array(SubtaskEvent).max(20).default([]),
+  // result (NEW)
+  result_summary: z.string().optional(),
+  result_text: z.string().optional(),
+  partial_result: z.string().nullable().optional(),
+  error: z.object({ kind: z.string(), message: z.string() }).optional(),
 }).meta({
   ref: "SubtaskPart",
 })
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -277,15 +277,6 @@ export const SubtaskPart = PartBase.extend({
   updated_at: z.number().optional(),
   ended_at: z.number().optional(),
   consumed_at: z.number().optional(),
-  // progress (NEW)
-  last_activity: z
-    .object({
-      kind: z.enum(["model_waiting", "thinking", "tool"]),
-      label: z.string(),
-      tool: z.string().optional(),
-      at: z.number(),
-    })
-    .optional(),
   recent_events: z.array(SubtaskEvent).max(20).default([]),
   // result (NEW)
   result_summary: z.string().optional(),
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -31,6 +31,11 @@ import type { Provider } from "@/provider"
 import { Permission } from "@/permission"
 import { Global } from "@/global"
 import { Effect, Layer, Option, Context } from "effect"
+import {
+  SubagentRunWriterContext,
+  SubagentRunGuardViolation,
+  lifecycleFieldsChanged,
+} from "./subagent-run-context"
 
 const log = Log.create({ service: "session" })
 
@@ -510,6 +515,26 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
 
     const updatePart = <T extends MessageV2.Part>(part: T): Effect.Effect<T> =>
       Effect.gen(function* () {
+        if (part.type === "subtask") {
+          const isWriter = yield* SubagentRunWriterContext
+          if (!isWriter) {
+            const existing = yield* getPart({
+              sessionID: part.sessionID,
+              messageID: part.messageID,
+              partID: part.id,
+            }).pipe(Effect.catch(() => Effect.succeed(undefined)))
+            if (
+              lifecycleFieldsChanged(
+                existing as unknown as Record<string, unknown> | undefined,
+                part as unknown as Record<string, unknown>,
+              )
+            ) {
+              return yield* Effect.die(
+                new SubagentRunGuardViolation((part as { tool_call_id?: string }).tool_call_id),
+              )
+            }
+          }
+        }
         yield* Effect.sync(() =>
           SyncEvent.run(MessageV2.Event.PartUpdated, {
             sessionID: part.sessionID,
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -518,11 +518,13 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
         if (part.type === "subtask") {
           const isWriter = yield* SubagentRunWriterContext
           if (!isWriter) {
+            // No catch: getPart's typed error channel is never; defects (e.g. database failure)
+            // should propagate so the guard can't be silently bypassed by a missing read.
             const existing = yield* getPart({
               sessionID: part.sessionID,
               messageID: part.messageID,
               partID: part.id,
-            }).pipe(Effect.catch(() => Effect.succeed(undefined)))
+            })
             if (
               lifecycleFieldsChanged(
                 existing as unknown as Record<string, unknown> | undefined,
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -525,9 +525,14 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
               messageID: part.messageID,
               partID: part.id,
             })
+            // Only police mutations: first writes (existing === undefined) are allowed so
+            // Session.fork() / migration / import paths can clone historical SubtaskParts with
+            // their persisted lifecycle values. Once a part exists, lifecycle fields are frozen
+            // for non-writers.
             if (
+              existing &&
               lifecycleFieldsChanged(
-                existing as unknown as Record<string, unknown> | undefined,
+                existing as unknown as Record<string, unknown>,
                 part as unknown as Record<string, unknown>,
               )
             ) {
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -68,6 +68,8 @@ export function fromRow(row: SessionRow): Info {
     workspaceID: row.workspace_id ?? undefined,
     directory: row.directory,
     parentID: row.parent_id ?? undefined,
+    createdByAgentTool: row.created_by_agent_tool ?? false,
+    subagentType: row.subagent_type ?? null,
     title: row.title,
     skill: row.skill ?? undefined,
     version: row.version,
@@ -90,6 +92,8 @@ export function toRow(info: Info) {
     project_id: info.projectID,
     workspace_id: info.workspaceID,
     parent_id: info.parentID,
+    created_by_agent_tool: info.createdByAgentTool,
+    subagent_type: info.subagentType,
     slug: info.slug,
     directory: info.directory,
     title: info.title,
@@ -127,6 +131,8 @@ export const Info = z
     workspaceID: WorkspaceID.zod.optional(),
     directory: z.string(),
     parentID: SessionID.zod.optional(),
+    createdByAgentTool: z.boolean().default(false),
+    subagentType: z.string().nullable().default(null),
     skill: z.string().optional(),
     summary: z
       .object({
@@ -189,6 +195,8 @@ export const CreateInput = z
     skill: z.string().optional(),
     permission: Info.shape.permission,
     workspaceID: WorkspaceID.zod.optional(),
+    createdByAgentTool: z.boolean().optional(),
+    subagentType: z.string().nullable().optional(),
   })
   .optional()
 export type CreateInput = z.output<typeof CreateInput>
@@ -349,6 +357,8 @@ export interface Interface {
     skill?: string
     permission?: Permission.Ruleset
     workspaceID?: WorkspaceID
+    createdByAgentTool?: boolean
+    subagentType?: string | null
   }) => Effect.Effect<Info>
   readonly fork: (input: { sessionID: SessionID; messageID?: MessageID }) => Effect.Effect<Info>
   readonly touch: (sessionID: SessionID) => Effect.Effect<void>
@@ -411,6 +421,8 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
       workspaceID?: WorkspaceID
       directory: string
       permission?: Permission.Ruleset
+      createdByAgentTool?: boolean
+      subagentType?: string | null
     }) {
       const ctx = yield* InstanceState.context
       const result: Info = {
@@ -421,6 +433,8 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
         directory: input.directory,
         workspaceID: input.workspaceID,
         parentID: input.parentID,
+        createdByAgentTool: input.createdByAgentTool ?? false,
+        subagentType: input.subagentType ?? null,
         title: input.title ?? createDefaultTitle(!!input.parentID),
         skill: input.skill,
         permission: input.permission,
@@ -535,6 +549,8 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
       skill?: string
       permission?: Permission.Ruleset
       workspaceID?: WorkspaceID
+      createdByAgentTool?: boolean
+      subagentType?: string | null
     }) {
       const directory = yield* InstanceState.directory
       const workspace = yield* InstanceState.workspaceID
@@ -545,6 +561,8 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
         skill: input?.skill,
         permission: input?.permission,
         workspaceID: input?.workspaceID ?? (workspace as WorkspaceID | undefined),
+        createdByAgentTool: input?.createdByAgentTool,
+        subagentType: input?.subagentType,
       })
     })
 
--- a/packages/opencode/src/session/subagent-run-context.ts
+++ b/packages/opencode/src/session/subagent-run-context.ts
@@ -38,26 +38,19 @@ export class SubagentRunGuardViolation extends Error {
  * last_activity / error arrays/objects with identical content is NOT rejected. Only genuine
  * lifecycle mutations trip the guard.
  *
+ * Caller contract: `existing` must be a real persisted part. First-write paths (no row yet)
+ * are allowed unconditionally upstream — Session.fork / migration / import need to seed
+ * historical lifecycle values without going through SubagentRun writers.
+ *
  * Limitation: JSON.stringify is order-sensitive (`{a:1,b:2}` vs `{b:2,a:1}` produce different
  * strings) and drops `undefined` values. This is acceptable here because lifecycle field shapes
  * are produced by SubagentRun writers with stable key order, and `undefined` lifecycle fields
- * are short-circuited above as "no change".
+ * compare equal to `undefined` on both sides.
  */
 export const lifecycleFieldsChanged = (
-  existing: Record<string, unknown> | undefined,
+  existing: Record<string, unknown>,
   next: Record<string, unknown>,
 ): boolean => {
-  if (!existing) {
-    // First write of this part. Reject only if the writer set a non-default lifecycle value.
-    for (const k of LIFECYCLE_KEYS) {
-      const v = next[k]
-      if (v === undefined) continue
-      if (k === "status" && v === "completed") continue
-      if (k === "recent_events" && Array.isArray(v) && v.length === 0) continue
-      return true
-    }
-    return false
-  }
   for (const k of LIFECYCLE_KEYS) {
     if (JSON.stringify(existing[k]) !== JSON.stringify(next[k])) return true
   }
--- a/packages/opencode/src/session/subagent-run-context.ts
+++ b/packages/opencode/src/session/subagent-run-context.ts
@@ -1,0 +1,57 @@
+import { Context } from "effect"
+
+/**
+ * True only inside SubagentRun service writer methods. Read by Session.updatePart
+ * to gate writes that mutate SubtaskPart lifecycle fields. Single shared module
+ * keeps both writers and the guard observing the same Reference without a Layer wrapper.
+ */
+export const SubagentRunWriterContext: Context.Reference<boolean> = Context.Reference<boolean>(
+  "@pawwork/SubagentRunWriterContext",
+  { defaultValue: () => false },
+)
+
+const LIFECYCLE_KEYS = [
+  "status",
+  "started_at",
+  "updated_at",
+  "ended_at",
+  "consumed_at",
+  "last_activity",
+  "recent_events",
+  "result_summary",
+  "result_text",
+  "partial_result",
+  "error",
+] as const
+
+export class SubagentRunGuardViolation {
+  readonly _tag = "SubagentRunGuardViolation"
+  constructor(readonly tool_call_id: string | undefined) {}
+}
+
+/**
+ * Returns true if `next` mutates any lifecycle field relative to `existing`. Compares by value
+ * (JSON.stringify) not by reference, so a static-field update that re-clones recent_events /
+ * last_activity / error arrays/objects with identical content is NOT rejected. Only genuine
+ * lifecycle mutations trip the guard.
+ */
+export const lifecycleFieldsChanged = (
+  existing: Record<string, unknown> | undefined,
+  next: Record<string, unknown>,
+): boolean => {
+  if (!existing) {
+    // First write of this part. Reject only if the writer set a non-default lifecycle value.
+    for (const k of LIFECYCLE_KEYS) {
+      const v = next[k]
+      if (v === undefined) continue
+      if (k === "status" && v === "completed") continue
+      if (k === "recent_events" && Array.isArray(v) && v.length === 0) continue
+      return true
+    }
+    return false
+  }
+  for (const k of LIFECYCLE_KEYS) {
+    if (JSON.stringify(existing[k]) !== JSON.stringify(next[k])) return true
+  }
+  return false
+}
--- a/packages/opencode/src/session/subagent-run-context.ts
+++ b/packages/opencode/src/session/subagent-run-context.ts
@@ -24,9 +24,12 @@ const LIFECYCLE_KEYS = [
   "error",
 ] as const
 
-export class SubagentRunGuardViolation {
+export class SubagentRunGuardViolation extends Error {
   readonly _tag = "SubagentRunGuardViolation"
-  constructor(readonly tool_call_id: string | undefined) {}
+  constructor(readonly tool_call_id: string | undefined) {
+    super(`SubagentRun lifecycle field write outside writer context (tool_call_id=${tool_call_id ?? "?"})`)
+    this.name = "SubagentRunGuardViolation"
+  }
 }
 
 /**
@@ -34,6 +37,11 @@ export class SubagentRunGuardViolation {
  * (JSON.stringify) not by reference, so a static-field update that re-clones recent_events /
  * last_activity / error arrays/objects with identical content is NOT rejected. Only genuine
  * lifecycle mutations trip the guard.
+ *
+ * Limitation: JSON.stringify is order-sensitive (`{a:1,b:2}` vs `{b:2,a:1}` produce different
+ * strings) and drops `undefined` values. This is acceptable here because lifecycle field shapes
+ * are produced by SubagentRun writers with stable key order, and `undefined` lifecycle fields
+ * are short-circuited above as "no change".
  */
 export const lifecycleFieldsChanged = (
   existing: Record<string, unknown> | undefined,
--- a/packages/opencode/src/session/subagent-run-context.ts
+++ b/packages/opencode/src/session/subagent-run-context.ts
@@ -24,7 +24,6 @@ const LIFECYCLE_KEYS = [
   "updated_at",
   "ended_at",
   "consumed_at",
-  "last_activity",
   "recent_events",
   "result_summary",
   "result_text",
@@ -43,7 +42,7 @@ export class SubagentRunGuardViolation extends Error {
 /**
  * Returns true if `next` mutates any lifecycle field relative to `existing`. Compares by value
  * (JSON.stringify) not by reference, so a static-field update that re-clones recent_events /
- * last_activity / error arrays/objects with identical content is NOT rejected. Only genuine
+ * recent_events / error arrays/objects with identical content is NOT rejected. Only genuine
  * lifecycle mutations trip the guard.
  *
  * Caller contract: `existing` must be a real persisted part. First-write paths (no row yet)
--- a/packages/opencode/src/session/subagent-run-context.ts
+++ b/packages/opencode/src/session/subagent-run-context.ts
@@ -10,7 +10,15 @@ export const SubagentRunWriterContext: Context.Reference<boolean> = Context.Refe
   { defaultValue: () => false },
 )
 
+// Includes both lifecycle state (status, timestamps, events, results) and immutable linkage
+// fields (tool_call_id, parent_*_id, subagent_session_id). Linkage fields are set once at
+// `start` / `patchSession` and must never be repointed: rewriting them would corrupt
+// findLatestBySessionID, ownership checks, and the in-memory tool_call_id index.
 const LIFECYCLE_KEYS = [
+  "tool_call_id",
+  "parent_session_id",
+  "parent_message_id",
+  "subagent_session_id",
   "status",
   "started_at",
   "updated_at",
--- a/packages/opencode/src/session/subagent-run.ts
+++ b/packages/opencode/src/session/subagent-run.ts
@@ -57,10 +57,6 @@ export interface Interface {
   readonly start: (input: StartInput) => Effect.Effect<MessageV2.SubtaskPart>
   readonly patchSession: (toolCallID: string, sessionID: SessionID) => Effect.Effect<void>
   readonly recordEvent: (toolCallID: string, event: MessageV2.SubtaskEvent) => Effect.Effect<void>
-  readonly recordActivity: (
-    toolCallID: string,
-    activity: NonNullable<MessageV2.SubtaskPart["last_activity"]>,
-  ) => Effect.Effect<void>
   readonly finalize: (
     toolCallID: string,
     status: TerminalStatus,
@@ -90,6 +86,9 @@ export interface Interface {
 
 export class Service extends Context.Service<Service, Interface>()("@pawwork/SubagentRun") {}
 
+// Default cap on parallel subagent dispatches per parent. Picked to stay within typical
+// per-conversation token / context budgets while still letting parents fan out useful
+// work. Not currently configurable; promoting to Config is a v1.1 follow-up.
 const MAX_ACTIVE = 5
 
 export const layer: Layer.Layer<Service, never, Session.Service> = Layer.effect(
@@ -137,9 +136,12 @@ export const layer: Layer.Layer<Service, never, Session.Service> = Layer.effect(
 
     const releaseSlot = (parentID: SessionID): Effect.Effect<void> =>
       getSlotLock(parentID).withPermits(1)(
-        Effect.sync(() => {
+        Effect.gen(function* () {
           const current = activeCounts.get(parentID) ?? 0
-          if (current > 0) activeCounts.set(parentID, current - 1)
+          // Underflow is always a bug (release without matching reserve). Die so the invariant
+          // violation surfaces instead of silently clipping at 0 and hiding double-release.
+          if (current <= 0) return yield* Effect.die(new Error(`releaseSlot underflow for ${parentID}`))
+          activeCounts.set(parentID, current - 1)
         }),
       )
 
@@ -200,12 +202,15 @@ export const layer: Layer.Layer<Service, never, Session.Service> = Layer.effect(
             updated_at: now,
             recent_events: [{ type: "started" as const, at: now }],
           } satisfies MessageV2.SubtaskPart
+          const persisted = yield* session.updatePart(part)
+          // Index after persistence — if updatePart fails, the cache stays clean instead of
+          // pointing at a row that does not exist in storage.
           partsByToolCall.set(input.tool_call_id, {
             sessionID: input.parent_session_id,
             messageID: input.parent_message_id,
             partID,
           })
-          return yield* session.updatePart(part)
+          return persisted
         }),
       )
 
@@ -214,18 +219,6 @@ export const layer: Layer.Layer<Service, never, Session.Service> = Layer.effect(
         session.updatePart({
           ...existing,
           subagent_session_id: sessionID,
-          updated_at: Date.now(),
-        }).pipe(Effect.asVoid),
-      )
-
-    const recordActivity = (
-      toolCallID: string,
-      activity: NonNullable<MessageV2.SubtaskPart["last_activity"]>,
-    ): Effect.Effect<void> =>
-      withExistingPart(toolCallID, (existing) =>
-        session.updatePart({
-          ...existing,
-          last_activity: activity,
           updated_at: Date.now(),
         }).pipe(Effect.asVoid),
       )
@@ -252,11 +245,13 @@ export const layer: Layer.Layer<Service, never, Session.Service> = Layer.effect(
           // Hard cap: if all 20 entries are lifecycle and a 21st arrives, evict the oldest
           // lifecycle entry so SubtaskEvent.array().max(20) cannot be violated on next decode.
           while (merged.length > 20) merged.shift()
-          // Sort after eviction so chronological order is preserved across lifecycle/progress.
-          merged.sort((a, b) => a.at - b.at)
+          // Stable sort by timestamp, breaking ties on insertion order so events sharing
+          // `Date.now()` (e.g., recordRejected's `started`+`failed` pair) keep deterministic order.
+          const indexed = merged.map((e, i) => ({ e, i }))
+          indexed.sort((a, b) => a.e.at - b.e.at || a.i - b.i)
           yield* session.updatePart({
             ...existing,
-            recent_events: merged,
+            recent_events: indexed.map((x) => x.e),
             updated_at: Date.now(),
           })
         }),
@@ -312,12 +307,14 @@ export const layer: Layer.Layer<Service, never, Session.Service> = Layer.effect(
             ],
             error: { kind: "too_many_active", message: input.reason },
           } satisfies MessageV2.SubtaskPart
+          const persisted = yield* session.updatePart(part)
+          // Index after persistence — same rationale as `start`.
           partsByToolCall.set(input.tool_call_id, {
             sessionID: input.parent_session_id,
             messageID: input.parent_message_id,
             partID,
           })
-          return yield* session.updatePart(part)
+          return persisted
         }),
       )
 
@@ -408,7 +405,6 @@ export const layer: Layer.Layer<Service, never, Session.Service> = Layer.effect(
       start,
       patchSession,
       recordEvent,
-      recordActivity,
       finalize,
       recordRejected,
       setConsumed,
--- a/packages/opencode/src/session/subagent-run.ts
+++ b/packages/opencode/src/session/subagent-run.ts
@@ -1,0 +1,294 @@
+import { Context, Effect, Layer, Semaphore } from "effect"
+import { Bus } from "../bus"
+import * as Session from "./session"
+import { PartID as PartIDNs, type MessageID, type PartID, type SessionID } from "./schema"
+import type { MessageV2 } from "./message-v2"
+import type { ProviderID, ModelID } from "../provider/schema"
+
+export class TooManyActive {
+  readonly _tag = "TooManyActive"
+  constructor(readonly parentID: SessionID) {}
+}
+
+export class NotFound {
+  readonly _tag = "NotFound"
+  constructor(readonly key: string) {}
+}
+
+export type TerminalStatus = "completed" | "completed_empty" | "failed" | "canceled_by_user"
+export type AgentListFilter = TerminalStatus | "running" | "all_active" | "all"
+
+export interface StartInput {
+  parent_session_id: SessionID
+  parent_message_id: MessageID
+  tool_call_id: string
+  description: string
+  prompt: string
+  agent: string
+  subagent_type: string
+  command?: string
+  model?: { providerID: ProviderID; modelID: ModelID }
+}
+
+export interface FinalizeFields {
+  result_text?: string
+  result_summary?: string
+  partial_result?: string | null
+  error?: { kind: string; message: string }
+  ended_at?: number
+}
+
+export interface RejectedInput {
+  parent_session_id: SessionID
+  parent_message_id: MessageID
+  tool_call_id: string
+  description: string
+  prompt: string
+  agent: string
+  subagent_type: string
+  command?: string
+  model?: { providerID: ProviderID; modelID: ModelID }
+  reason: string
+}
+
+export interface Interface {
+  readonly reserveSlot: (parentID: SessionID) => Effect.Effect<void, TooManyActive>
+  readonly releaseSlot: (parentID: SessionID) => Effect.Effect<void>
+  readonly start: (input: StartInput) => Effect.Effect<MessageV2.SubtaskPart>
+  readonly patchSession: (toolCallID: string, sessionID: SessionID) => Effect.Effect<void>
+  readonly recordEvent: (toolCallID: string, event: MessageV2.SubtaskEvent) => Effect.Effect<void>
+  readonly recordActivity: (
+    toolCallID: string,
+    activity: NonNullable<MessageV2.SubtaskPart["last_activity"]>,
+  ) => Effect.Effect<void>
+  readonly finalize: (
+    toolCallID: string,
+    status: TerminalStatus,
+    fields: FinalizeFields,
+  ) => Effect.Effect<void>
+  readonly recordRejected: (input: RejectedInput) => Effect.Effect<MessageV2.SubtaskPart>
+  readonly setConsumed: (toolCallID: string) => Effect.Effect<void>
+  readonly read: (toolCallID: string) => Effect.Effect<MessageV2.SubtaskPart, NotFound>
+  readonly findLatestBySessionID: (
+    parentID: SessionID,
+    subagentSessionID: SessionID,
+  ) => Effect.Effect<MessageV2.SubtaskPart, NotFound>
+  readonly list: (
+    parentID: SessionID,
+    filter: { status: AgentListFilter; limit: number },
+  ) => Effect.Effect<MessageV2.SubtaskPart[]>
+}
+
+export class Service extends Context.Service<Service, Interface>()("@pawwork/SubagentRun") {}
+
+const MAX_ACTIVE = 5
+
+export const layer: Layer.Layer<Service, never, Bus.Service | Session.Service> = Layer.effect(
+  Service,
+  Effect.gen(function* () {
+    const session = yield* Session.Service
+
+    const slotLocks = new Map<SessionID, Semaphore.Semaphore>()
+    const activeCounts = new Map<SessionID, number>()
+    const rowLocks = new Map<string, Semaphore.Semaphore>()
+    const partsByToolCall = new Map<
+      string,
+      { sessionID: SessionID; messageID: MessageID; partID: PartID }
+    >()
+
+    const getSlotLock = (parentID: SessionID) => {
+      const hit = slotLocks.get(parentID)
+      if (hit) return hit
+      const next = Semaphore.makeUnsafe(1)
+      slotLocks.set(parentID, next)
+      return next
+    }
+
+    const getRowLock = (toolCallID: string) => {
+      const hit = rowLocks.get(toolCallID)
+      if (hit) return hit
+      const next = Semaphore.makeUnsafe(1)
+      rowLocks.set(toolCallID, next)
+      return next
+    }
+
+    const reserveSlot = (parentID: SessionID): Effect.Effect<void, TooManyActive> =>
+      getSlotLock(parentID).withPermits(1)(
+        Effect.gen(function* () {
+          const current = activeCounts.get(parentID) ?? 0
+          if (current >= MAX_ACTIVE) {
+            return yield* Effect.fail(new TooManyActive(parentID))
+          }
+          activeCounts.set(parentID, current + 1)
+        }),
+      )
+
+    const releaseSlot = (parentID: SessionID): Effect.Effect<void> =>
+      getSlotLock(parentID).withPermits(1)(
+        Effect.sync(() => {
+          const current = activeCounts.get(parentID) ?? 0
+          if (current > 0) activeCounts.set(parentID, current - 1)
+        }),
+      )
+
+    const readPart = (toolCallID: string) =>
+      Effect.gen(function* () {
+        const ref = partsByToolCall.get(toolCallID)
+        if (!ref) return yield* Effect.fail(new NotFound(toolCallID))
+        const got = yield* session.getPart({
+          sessionID: ref.sessionID,
+          messageID: ref.messageID,
+          partID: ref.partID,
+        })
+        if (!got || got.type !== "subtask") return yield* Effect.fail(new NotFound(toolCallID))
+        return got as MessageV2.SubtaskPart
+      })
+
+    const start = Effect.fn("SubagentRun.start")(function* (input: StartInput) {
+      const partID = PartIDNs.ascending() as PartID
+      const now = Date.now()
+      const part = {
+        type: "subtask" as const,
+        id: partID,
+        sessionID: input.parent_session_id,
+        messageID: input.parent_message_id,
+        prompt: input.prompt,
+        description: input.description,
+        agent: input.agent,
+        model: input.model,
+        command: input.command,
+        tool_call_id: input.tool_call_id,
+        parent_session_id: input.parent_session_id,
+        parent_message_id: input.parent_message_id,
+        subagent_session_id: undefined,
+        status: "running" as const,
+        started_at: now,
+        updated_at: now,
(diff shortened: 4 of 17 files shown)


Astro-Han (commented):
## Review: subagent lifecycle v1

Overall this is a solid, well-tested PR. The scoped 8-step flow, single-writer guard, and explicit terminal states are good architectural choices. Below are issues found at various severity levels.

---

### P0 — Must fix before merge

**1. `agent_output` swallows ALL errors as "not found" via blanket `Effect.catch`**
`agent_output.ts` lines 75-78 use `.pipe(Effect.catch(() => Effect.succeed(null)))` on both `readByToolCallID` and `findLatestBySessionID`. This means a storage I/O error, a JSON parse failure, or any other defect gets silently converted to "subagent not found", which is a false-negative that misleads the model. Narrow the catch to `Effect.catchTag("NotFound", ...)` so real failures surface as defects.

**2. `finalizeAfter` reads the row AFTER checking `wasInterrupted`, creating a race**
In `agent.ts`, `finalizeAfter` does `const interrupted = ops.wasInterrupted(nextSession.id)` first, then `const current = yield* subagentRun.read(ctx.callID!)`. If the child runner completes naturally, then the parent aborts, then `wasInterrupted` returns true (because `onInterrupt` fired), but between that and the `read`, another fiber finalizes the row to `completed`. The `current.status !== "running"` guard at line 391 will skip the finalize, but the `interrupted` variable was already computed as `true`. This is benign because the early-return prevents the wrong path from running, but the logic is fragile: the read should happen first, then `wasInterrupted` should be consulted only if the row is still running. Reorder so `read` happens before `wasInterrupted`.

**3. `recordEvent` sorts by `at` but `at` can collide, making sort order non-deterministic**
In `subagent-run.ts` line 257, `merged.sort((a, b) => a.at - b.at)` can produce unstable ordering when two events share the same timestamp (e.g. `started` and `failed` in `recordRejected` both use `Date.now()`). Add a secondary sort key (array index) to preserve insertion order.

**4. `releaseSlot` can underflow `activeCounts` without warning**
`subagent-run.ts` line 143: `if (current > 0) activeCounts.set(parentID, current - 1)`. If a bug ever causes `releaseSlot` to be called more times than `reserveSlot`, the count silently stays at 0 instead of going negative. An underflow would mask a logic error. Consider asserting `current > 0` and dying if not, or at least logging a warning.

---

### P1 — Should fix before merge

**5. `errorMessage` can leak sensitive JSON payloads before sanitization**
In `agent.ts`, `errorMessage` stringifies arbitrary unknowns with `JSON.stringify(e)`. If the error is a model response containing API keys, tokens, or PII, this gets passed straight into `sanitizeErrorMessage`, which only strips stacks/paths/JSON envelopes — it does NOT strip the raw JSON content inside the envelope. The `JSON_ENVELOPE_RE` regex removes `{...}` blocks, but if `JSON.stringify(e)` produces a flat string without braces (e.g. `"sensitive_token_here"`), it passes through untouched. Consider redacting known sensitive patterns or truncating before JSON.stringify.

**6. `findLatestBySessionID` does a full table scan of all parent messages**
`subagent-run.ts` lines 383-393: `collectSubtaskParts` loads ALL messages in the parent session, then ALL parts of each message, then filters. For a long-running parent session with many messages, this is O(n*m) and gets worse with every subagent dispatch. Add an index or at least a message-time bound to limit the scan.

**7. `agent_output` does not validate `subagent_session_id` belongs to the parent before calling `findLatestBySessionID`**
`agent_output.ts` line 77 calls `findLatestBySessionID(ctx.sessionID, params.subagent_session_id! as SessionID)`. The `findLatestBySessionID` implementation scans the parent's messages and filters by `subagent_session_id`, so it implicitly enforces parent ownership. But if a caller passes a `subagent_session_id` that belongs to a DIFFERENT parent's child, the scan simply returns nothing and the tool reports "not found". This is correct behavior but the error message is misleading — it says "not found or not accessible" but the real issue is cross-parent access. Consider a clearer error.

**8. `synthesizeOutput` for `failed` status does not include `subagent_session_id` in the resume hint**
In `agent.ts`, the `failed` branch at line 119 produces `header = \`status: failed\\nerror: ...\``. The `resumeHint` is still computed (line 84-86) but is only included when `childID` is truthy. For `recordRejected` (cap rejection), `childID` is `undefined` so no resume hint appears — this is correct. But for a `failed` run that DID create a child session (e.g. model error after `start` + `patchSession`), the resume hint IS included, yet the status header says "failed" with no guidance that the session can be resumed. Consider adding a note like "(session still resumable)" to the failed header when `childID` is present.

**9. `MAX_ACTIVE = 5` is hardcoded with no config hook**
`subagent-run.ts` line 94. The PR description says "caps parallel dispatch at 5 per parent" but there's no way to tune this. At minimum, document why 5 was chosen (token budget? UI clutter? memory?). Better: make it a config value with a sensible default.

**10. `agent_list` `formatElapsed` shows `0s` for negative or zero durations**
`agent-list.ts` line 9: `if (ms <= 0) return "0s"`. A negative duration means `started_at` is in the future (clock skew), which is a real bug. Returning "0s" silently hides the skew. Return something like "-" or "clock skew" to signal the anomaly.

---

### P2 — Nice to have

**11. `SubtaskEvent` discriminated union lacks a `details` field for extensibility**
`message-v2.ts`: Events like `tool_started` have `tool` and `label`, but `failed` only has `kind`. If v1.1 wants to add `retry_count`, `duration_ms`, or `stack_trace`, the union shape needs to change. Add an optional `details?: Record<string, unknown>` to each variant now to avoid a migration later.

**12. `recent_events` ring sorts after every append, which is O(n log n) per event**
`subagent-run.ts` line 257 sorts the entire array after every `recordEvent`. With 20 items this is trivial, but if the cap ever increases, this becomes wasteful. Since events are append-only and mostly chronological, a simple `push` + `splice` at the front is O(n) and avoids the sort. Not a real perf issue at n=20, but the sort is unnecessary complexity.

**13. `agent_output` `formatTranscript` truncates at 1000 chars with no indication of total length**
`agent-output.ts` line 37: `body.slice(0, 1000) + "…(truncated)"`. The model doesn't know how much was cut. Include the original length: `…(truncated, 2345 chars total)`.

**14. `lifecycleFieldsChanged` uses `JSON.stringify` for value comparison, which is brittle**
`subagent-run-context.ts` line 55: `JSON.stringify(existing[k]) !== JSON.stringify(next[k])`. This fails for objects with different key orders (`{a:1,b:2}` vs `{b:2,a:1}` stringify differently), and for `undefined` values (JSON.stringify skips them). Use a deep-equal utility or at least document this limitation.

**15. `session.ts` `updatePart` guard does a `getPart` lookup for EVERY subtask part write, even writer writes**
`session.ts` lines 83-98: The guard checks `!isWriter` first, but the `getPart` call and `lifecycleFieldsChanged` computation happen inside the `if (!isWriter)` block — that's correct. However, the `Effect.catch(() => Effect.succeed(undefined))` on `getPart` means a storage error during the guard check is silently swallowed, causing the guard to pass (because `existing` is undefined and `lifecycleFieldsChanged` returns `false` for first writes with default values). A storage error should die, not let the write through.

**16. `AgentTool.execute` outer `Effect.catch` uses `Effect.fail(error)` which loses the original Cause**
`agent.ts` line 466: `return yield* Effect.fail(error)`. This reconstructs the error as a plain failure, losing any `Cause` metadata (stack, annotations, parallel errors). Use `Effect.failCause` to preserve the original cause.

---

### P3 — Nitpicks / style

**17. `agent.ts` `makeReadLastCompletedAssistantText` reads up to 5 messages but only needs 1**
The function scans up to 5 messages looking for the most recent assistant text. Since messages are ordered, reading just the last 1-2 messages would suffice. The `limit: 5` is arbitrary and wastes I/O.

**18. `subagent-run.ts` `collectSubtaskParts` is duplicated logic between `findLatestBySessionID`, `list`, and `readByToolCallID`**
All three methods call `collectSubtaskParts` which does a full parent message scan. Extract a cached/indexed approach or at least memoize within a single Effect scope.

**19. `agent_output.ts` `NOT_FOUND` error is constructed eagerly on every call**
Line 71 creates `new Error(...)` before knowing if it's needed. Move inside the `if (!row)` branch.

**20. `subagent-run.ts` `withExistingPart` silently no-ops when the row is missing**
Line 164-178: If `readPart` returns `NotFound`, the mutation is skipped silently. For `finalize`, `setConsumed`, and `recordEvent`, a missing row is a real bug (e.g. `finalize` called on a non-existent `tool_call_id`). Consider making the no-op behavior explicit per-mutator rather than universal.

**21. `agent-list.ts` `formatRow` uses `p.subagent_session_id ?? "-"` but legacy rows may have `undefined` session ID**
For legacy rows, showing "-" is fine, but the `statusLabel` logic at line 23 treats `consumed_at` as the unread marker. Legacy rows have no `consumed_at`, so they always show as `(unread)` even if they were completed long ago. Consider defaulting `consumed_at` for legacy decoded rows.

**22. `SubagentRunGuardViolation` is thrown via `Effect.die` but carries no stack trace**
`session.ts` line 94-95: `Effect.die(new SubagentRunGuardViolation(...))`. `Effect.die` creates a defect, but the `SubagentRunGuardViolation` class doesn't extend `Error`, so standard stack traces won't be captured. Make it `extends Error`.

**23. `agent.ts` `finalizeAfter` has a comment saying "Do NOT also OR with ctx.abort.aborted" but the code doesn't OR it anymore**
The comment at lines 378-384 is a good explanation of WHY the fix was made, but it reads like a commit message embedded in source. Trim or move to a docstring.

**24. `agent.ts` `synthesizeOutput` `completed_empty` branch wraps an empty string with `<subagent_result>` tags**
Line 112: `body = wrapper("")` produces `<subagent_result>\n\n</subagent_result>`. For an empty result, the model sees an empty wrapper. Consider omitting the wrapper entirely for `completed_empty`.

**25. `subagent-run.ts` `recordRejected` uses `Date.now()` for both `started_at` and `ended_at`, making the run appear instantaneous**
This is semantically correct (the rejection is immediate), but the `recent_events` array contains both `started` and `failed` at the same timestamp. The `recordEvent` sort will order them arbitrarily (see P0 #3). Ensure deterministic ordering.

**26. `agent_output.ts` `formatResult` for `failed` returns `error.message` but `sanitizeErrorMessage` was already applied in `synthesizeOutput`**
The `failed` branch in `formatResult` reads `part.error?.message`, which was already sanitized when `finalize` stored it. But `agent_output` is also callable by the model directly, and the model might want the raw error. The double-sanitization is fine, but there's no way to get the raw error kind+message pair. Consider whether `agent_output` should ever expose unsanitized errors.

**27. `subagent-run.ts` `Interface` has `recordActivity` but it's never called in this PR**
The method exists but has no callers in the diff. Either remove it or add a TODO comment explaining when v1.1 will use it.

**28. `agent.ts` `Effect.scoped` + `Effect.addFinalizer` pattern is elegant but the inner `Effect.gen` at line 287 is 150+ lines deep**
Consider extracting the scoped body into a named function for readability. The nesting makes it hard to trace where `Effect.scoped` ends.

**29. `message-v2.ts` `SubtaskPart` schema has `.meta({ ref: "SubtaskPart" })` but the new fields aren't reflected in any TS type-level constraints**
The `SubtaskPart` type is inferred from Zod, but the `satisfies MessageV2.SubtaskPart` usage in `subagent-run.ts` (lines 203, 315) is a runtime assertion, not compile-time. The `SubtaskPart` type should be exported and used as the return type for `start` and `recordRejected`.

**30. `prompt.ts` `interruptedSessions` is a `Set<SessionID>` but `SessionID` is a branded string; the Set uses reference equality**
This is fine for strings, but if `SessionID` ever becomes an object type, the Set lookup breaks. Document that `SessionID` must remain a string brand.

packages/opencode/src/tool/agent-output.ts
```diff
@@ -0,0 +1,94 @@
...
+        },
+        ctx: Tool.Context,
+      ) =>
+        Effect.gen(function* () {
+          const NOT_FOUND = new Error("subagent not found or not accessible from this parent")
+          const row = params.tool_call_id
+            ? yield* subagentRun
+                .readByToolCallID(ctx.sessionID, params.tool_call_id)
+                .pipe(Effect.catch(() => Effect.succeed(null as MessageV2.SubtaskPart | null)))
+            : yield* subagentRun
```
Astro-Han, line 95: **[P0]** Blanket Effect.catch swallows all errors as not found

Astro-Han, line 95: **[P0]** Blanket `Effect.catch(() => Effect.succeed(null))` swallows storage errors, JSON parse failures, and defects as "not found". Narrow to `Effect.catchTag("NotFound", ...)` so real failures surface as defects instead of misleading the model with a false-negative.

packages/opencode/src/tool/agent.ts
```diff
@@ -54,132 +181,242 @@ export const AgentTool = Tool.define(
...
+                tools: {
+                  agent: false,
+                  ...(canTodo ? {} : { todowrite: false }),
+                  ...Object.fromEntries(
+                    (cfg.experimental?.primary_tools ?? []).map((item) => [item, false]),
+                  ),
+                },
+                parts,
+              })
+              .pipe(
```
Astro-Han, line 412: **[P0]** `wasInterrupted` is read before `subagentRun.read`. If parent aborts after natural completion, `interrupted` is `true` but the row may already be `completed` by the time `read` happens. The `current.status !== "running"` guard prevents the wrong finalize, but the logic is fragile. Reorder: read first, then consult `wasInterrupted` only when still running.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,427 @@
...
+            const idx = merged.findIndex((e) => !LIFECYCLE_KINDS.has(e.type))
+            if (idx < 0) break
+            merged.splice(idx, 1)
+          }
+          // Hard cap: if all 20 entries are lifecycle and a 21st arrives, evict the oldest
+          // lifecycle entry so SubtaskEvent.array().max(20) cannot be violated on next decode.
+          while (merged.length > 20) merged.shift()
+          // Sort after eviction so chronological order is preserved across lifecycle/progress.
+          merged.sort((a, b) => a.at - b.at)
+          yield* session.updatePart({
```
Astro-Han: **[P0]** `merged.sort((a, b) => a.at - b.at)` is unstable when timestamps collide (e.g. `started` and `failed` both use `Date.now()` in `recordRejected`). Add a secondary sort key (original array index) to preserve deterministic insertion order.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,427 @@
...
+          activeCounts.set(parentID, current + 1)
+        }),
+      )
+
+    const releaseSlot = (parentID: SessionID): Effect.Effect<void> =>
+      getSlotLock(parentID).withPermits(1)(
+        Effect.sync(() => {
+          const current = activeCounts.get(parentID) ?? 0
+          if (current > 0) activeCounts.set(parentID, current - 1)
+        }),
```
Astro-Han, line 143: **[P0]** `if (current > 0) activeCounts.set(parentID, current - 1)` silently clips underflows. If a bug causes extra releases, the count stays at 0 and the invariant violation is hidden. Assert `current > 0` and die (or at least log) on underflow so the bug is visible.

packages/opencode/src/tool/agent.ts
```diff
@@ -1,42 +1,169 @@
...
+  /**
+   * After ops.prompt resolves, returns true iff the child session's runner.onInterrupt
+   * fired during execution (parent abort propagated through `cancel()`, OR a user
+   * canceled the child session directly). Returns false on natural completion or model
+   * failure. Synchronous query.
+   */
+  wasInterrupted(sessionID: SessionID): boolean
 }
 
 const id = "agent"
```
Astro-Han, line 26: **[P1]** `errorMessage` calls `JSON.stringify(e)` on arbitrary unknowns before `sanitizeErrorMessage` runs. If the error contains API keys, tokens, or PII in a flat string (no `{...}` envelope), `JSON_ENVELOPE_RE` won't strip it. Sanitize or redact BEFORE stringification, or avoid `JSON.stringify` on opaque errors entirely.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,427 @@
...
+      return part.status === filter
+    }
+
+    const findLatestBySessionID = (
+      parentID: SessionID,
+      subagentSessionID: SessionID,
+    ): Effect.Effect<MessageV2.SubtaskPart, NotFound> =>
+      Effect.gen(function* () {
+        const all = yield* collectSubtaskParts(parentID)
+        const matches = all.filter((p) => p.subagent_session_id === subagentSessionID)
```
Astro-Han: **[P1]** `collectSubtaskParts` loads ALL messages and ALL parts in the parent session. For long sessions this is O(n*m) and grows unbounded. Add a time-bound or message-limit to the scan, or build an index on `subagent_session_id`.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,427 @@
...
+  readonly list: (
+    parentID: SessionID,
+    filter: { status: AgentListFilter; limit: number },
+  ) => Effect.Effect<MessageV2.SubtaskPart[]>
+}
+
+export class Service extends Context.Service<Service, Interface>()("@pawwork/SubagentRun") {}
+
+const MAX_ACTIVE = 5
+
```
Astro-Han, line 91: **[P1]** `MAX_ACTIVE = 5` is hardcoded with no config hook or justification comment. Document why 5 (token budget? memory? UI clutter?), or expose it via `Config`.

packages/opencode/src/tool/agent-list.ts
```diff
@@ -0,0 +1,74 @@
+import z from "zod"
+import { Effect } from "effect"
+import * as Tool from "./tool"
+import { SubagentRun } from "../session/subagent-run"
+import type { MessageV2 } from "../session/message-v2"
+
+const formatElapsed = (ms: number): string => {
+  if (ms <= 0) return "0s"
+  const s = Math.floor(ms / 1000)
```
Astro-Han, line 23: **[P1]** `if (ms <= 0) return "0s"` hides clock skew. A negative duration means `started_at` is in the future. Return something like `"clock skew"` or `"-"` so the anomaly is visible instead of silently coerced.

packages/opencode/src/tool/agent-output.ts
```diff
@@ -0,0 +1,94 @@
...
+    `events: ${p.recent_events.length}`,
+    `child_session: ${p.subagent_session_id ?? "-"}`,
+    `latest: ${p.last_activity?.label ?? "-"}`,
+  ]
+  // Mirror formatResult's fallback chain: prefer result_text, then partial_result, so
+  // canceled_by_user rows still surface their preserved partial under detail=transcript.
+  const body = p.result_text ?? p.partial_result ?? null
+  if (body) {
+    const trimmed = body.length > 1000 ? body.slice(0, 1000) + "…(truncated)" : body
+    lines.push(`result: ${trimmed}`)
```
Astro-Han, line 56: **[P2]** `body.slice(0, 1000) + "…(truncated)"` doesn't tell the model how much was cut. Include the original length: `…(truncated, ${body.length} chars total)`.

packages/opencode/src/session/subagent-run-context.ts
```diff
@@ -0,0 +1,57 @@
...
+      if (v === undefined) continue
+      if (k === "status" && v === "completed") continue
+      if (k === "recent_events" && Array.isArray(v) && v.length === 0) continue
+      return true
+    }
+    return false
+  }
+  for (const k of LIFECYCLE_KEYS) {
+    if (JSON.stringify(existing[k]) !== JSON.stringify(next[k])) return true
+  }
```
Astro-Han, line 63: **[P2]** `JSON.stringify(existing[k]) !== JSON.stringify(next[k])` fails for objects with different key orders and skips `undefined` values. Use a deep-equal utility or document the limitation explicitly.

packages/opencode/src/session/subagent-run-context.ts
```diff
@@ -0,0 +1,57 @@
...
+  "last_activity",
+  "recent_events",
+  "result_summary",
+  "result_text",
+  "partial_result",
+  "error",
+] as const
+
+export class SubagentRunGuardViolation {
+  readonly _tag = "SubagentRunGuardViolation"
```
Astro-Han, line 35: **[P3]** `SubagentRunGuardViolation` does not extend `Error`. When `Effect.die` creates a defect, standard stack traces won't be captured. Make it `class SubagentRunGuardViolation extends Error`.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,427 @@
...
+}
+
+export interface Interface {
+  readonly reserveSlot: (parentID: SessionID) => Effect.Effect<void, TooManyActive>
+  readonly releaseSlot: (parentID: SessionID) => Effect.Effect<void>
+  readonly start: (input: StartInput) => Effect.Effect<MessageV2.SubtaskPart>
+  readonly patchSession: (toolCallID: string, sessionID: SessionID) => Effect.Effect<void>
+  readonly recordEvent: (toolCallID: string, event: MessageV2.SubtaskEvent) => Effect.Effect<void>
+  readonly recordActivity: (
+    toolCallID: string,
```
Astro-Han: **[P3]** `recordActivity` is declared in the interface but has no callers in this PR. Either remove it or add a `// TODO(v1.1): called from progress reporting` comment so readers know it's intentional dead code.

packages/opencode/src/session/message-v2.ts
```diff
@@ -239,6 +217,19 @@ export const CompactionPart = PartBase.extend({
 })
 export type CompactionPart = z.infer<typeof CompactionPart>
 
+export const SubtaskEvent = z.discriminatedUnion("type", [
```
Astro-Han, line 242: **[P2]** `SubtaskEvent` variants have fixed shapes. If v1.1 needs `retry_count`, `duration_ms`, etc., the union must change. Add an optional `details?: Record<string, unknown>` to each variant now to avoid a schema migration later.

packages/opencode/src/session/message-v2.ts
```diff
@@ -251,6 +242,34 @@ export const SubtaskPart = PartBase.extend({
     })
     .optional(),
   command: z.string().optional(),
+  // identity (NEW; .optional() so legacy decode succeeds without a primary key)
+  tool_call_id: z.string().optional(),
```
Astro-Han, line 268: **[P3]** `satisfies MessageV2.SubtaskPart` in `subagent-run.ts` is a runtime assertion. The `SubtaskPart` type should be explicitly exported and used as the return type for `start` and `recordRejected` for compile-time safety.

packages/opencode/src/session/session.ts
```diff
@@ -496,6 +515,26 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
 
     const updatePart = <T extends MessageV2.Part>(part: T): Effect.Effect<T> =>
       Effect.gen(function* () {
+        if (part.type === "subtask") {
+          const isWriter = yield* SubagentRunWriterContext
+          if (!isWriter) {
+            const existing = yield* getPart({
```
Astro-Han, line 523: **[P2]** `getPart(...).pipe(Effect.catch(() => Effect.succeed(undefined)))` means a storage error during the guard check is silently swallowed. The guard then sees `existing = undefined`, and `lifecycleFieldsChanged` returns `false` for first writes with default values, allowing the write through. A storage error should die, not bypass the guard.

packages/opencode/src/tool/agent.ts
```diff
@@ -54,132 +181,242 @@ export const AgentTool = Tool.define(
...
+                const current = yield* subagentRun
+                  .read(ctx.callID!)
+                  .pipe(Effect.catchTag("NotFound", () => Effect.succeed(null as MessageV2.SubtaskPart | null)))
+                if (current?.status === "running") {
+                  yield* subagentRun.finalize(ctx.callID!, "failed", {
+                    error: { kind: "execution_error", message: errorMessage(error) },
+                    ended_at: Date.now(),
+                  })
+                }
+                return yield* Effect.fail(error)
```
Astro-Han: **[P2]** `return yield* Effect.fail(error)` reconstructs the error as a plain failure, losing `Cause` metadata (stack, annotations, parallel errors). Use `Effect.failCause` to preserve the original cause.

packages/opencode/src/tool/agent-output.ts
```diff
@@ -0,0 +1,94 @@
...
+                .findLatestBySessionID(ctx.sessionID, params.subagent_session_id! as SessionID)
+                .pipe(Effect.catch(() => Effect.succeed(null as MessageV2.SubtaskPart | null)))
+          if (!row || !row.tool_call_id) return yield* Effect.fail(NOT_FOUND)
+          if (row.parent_session_id !== ctx.sessionID) return yield* Effect.fail(NOT_FOUND)
+          if (row.subagent_session_id) {
+            const child = yield* sessions
+              .get(row.subagent_session_id as SessionID)
+              .pipe(Effect.catch(() => Effect.succeed(null)))
+            if (!child || !child.createdByAgentTool) return yield* Effect.fail(NOT_FOUND)
+          }
```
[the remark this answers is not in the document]
Astro-Han: Punting on splitting the error message by failure condition. Collapsing "row missing", "wrong parent", and "child not agent-tool-marked" into a single response is intentional so callers cannot distinguish "I am not allowed to see it" from "it does not exist". The catch was narrowed in 87cde6d3 (`catchTag("NotFound")` on the typed reads, `catchCause` on `sessions.get` because its throw becomes `Cause.Die`), so genuine defects now propagate via `Effect.orDie` instead of being silently reported as not-found.

packages/opencode/src/tool/agent-output.ts
Astro-Han: Done in 87cde6d3 — replaced blanket `Effect.catch` with `Effect.catchTag("NotFound")` on both `readByToolCallID` and `findLatestBySessionID`. `sessions.get` keeps a broad `catchCause` because `NotFoundError` is thrown (lifted to `Cause.Die`), not a typed failure.

Astro-Han: Same as the sibling thread above — fixed in 87cde6d3.

packages/opencode/src/tool/agent.ts
Astro-Han: Push back. The `current.status !== "running"` guard inside `finalizeAfter` already short-circuits before `interrupted` can produce a wrong finalize, so the late-abort race is benign. Reordering the two reads would not change behavior — the second read is what actually decides whether to act. YAGNI.

packages/opencode/src/session/subagent-run.ts
Astro-Han: Done in 09845950 — `recordEvent` now sorts on `(at, insertion_index)` so events sharing `Date.now()` (e.g. `recordRejected`'s `started`+`failed` pair) keep deterministic order.

packages/opencode/src/session/subagent-run.ts
Astro-Han: Done in 09845950 — `releaseSlot` now `Effect.die`s on underflow with `new Error("releaseSlot underflow for ${parentID}")` instead of clipping at 0, so any reserve/release imbalance surfaces as a defect.

packages/opencode/src/tool/agent.ts
Astro-Han: Done in bc5d51f4 — `errorMessage` now calls `sanitizeErrorMessage` at the entry point, so the persisted `SubtaskPart.error.message` is always sanitized before reaching storage. The existing `sanitizeErrorMessage` call in `synthesizeOutput` stays (idempotent under double-application) as defense-in-depth.

packages/opencode/src/session/subagent-run.ts
Astro-Han: Acknowledged — `collectSubtaskParts` is O(n*m) over parent messages. v1 keeps it: `agent_list` defaults to a 5-row cap, and interactive parent sessions stay well under the size where this matters. The proper fix is a `subagent_session_id` index, deferred to v1.1.

packages/opencode/src/session/subagent-run.ts
Astro-Han: Done in 09845950 — added a comment explaining "5 = typical per-conversation token / context budget" and a v1.1 follow-up note. Not promoting to `Config` until someone hits the limit; YAGNI.

packages/opencode/src/tool/agent-list.ts
Astro-Han: Done in f345f9a5 — negative durations now render as `"-"` so clock skew is visible.

packages/opencode/src/tool/agent-output.ts
Astro-Han: Done in 87cde6d3 — truncate now appends `…(truncated, ${body.length} chars total)`.

packages/opencode/src/session/subagent-run-context.ts
Astro-Han: Done in d35edb1a — added a `Limitation:` paragraph on `lifecycleFieldsChanged` documenting JSON.stringify's order-sensitivity and `undefined`-dropping. Not switching to a deep-equal util because lifecycle field shapes are produced by SubagentRun writers with stable key order, and `undefined` lifecycle values are short-circuited above.

packages/opencode/src/session/subagent-run-context.ts
Astro-Han: Done in d35edb1a — `SubagentRunGuardViolation` now `extends Error` with a constructed message, so `Effect.die` preserves a real stack trace at the violation point.

packages/opencode/src/session/subagent-run.ts
Astro-Han: Done in 09845950 — removed `recordActivity` from the `Interface`, the runtime closure, and `Service.of`. v1 has no caller; if v1.1 needs progress reporting, the method comes back then.

packages/opencode/src/session/message-v2.ts
Astro-Han: Push back, YAGNI. Adding `details?: Record<string, unknown>` now bakes in an untyped grab-bag for hypothetical future fields. If v1.1 actually needs `retry_count` / `duration_ms`, the cost is one schema migration scoped to the variant that changes — same cost as adding the typed field directly, and the schema stays self-documenting in the meantime.

packages/opencode/src/session/message-v2.ts
Astro-Han: Push back. Runtime `satisfies MessageV2.SubtaskPart` already catches drift at the call sites, and `start` / `recordRejected` are typed `Effect.Effect<MessageV2.SubtaskPart>` via the interface. Switching to an explicitly exported `SubtaskPart` return type adds churn without changing what compile-time enforces.

packages/opencode/src/session/session.ts
Astro-Han: Done in d35edb1a — dropped the swallowing `Effect.catch` on the `getPart` guard read. Its typed error channel is `never`, so the catch was a noop on failures and a foot-gun on defects. Defects now propagate so the guard cannot be bypassed by a missing read.

packages/opencode/src/tool/agent.ts
Astro-Han: Done in bc5d51f4 — switched the outer pipe to `Effect.catchCause(cause => ...)` and re-raise via `Effect.failCause(cause)` so stack annotations and parallel-error metadata are preserved.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,427 @@
...
+
+export class TooManyActive {
+  readonly _tag = "TooManyActive"
+  constructor(readonly parentID: SessionID) {}
+}
+
+export class NotFound {
+  readonly _tag = "NotFound"
+  constructor(readonly key: string) {}
+}
```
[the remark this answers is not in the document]
Astro-Han: Push back for v1. The repo's existing pattern for service-internal errors is `NamedError.create` (zod-based) plus plain classes; `SubagentRun` follows that. Switching this one module to `Schema.TaggedErrorClass` would diverge from the surrounding code. Worth doing as part of a repo-wide error-class sweep, not a one-off here.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,427 @@
...
+    const withWriter = <A, E, R>(eff: Effect.Effect<A, E, R>): Effect.Effect<A, E, R> =>
+      Effect.provideService(eff, SubagentRunWriterContext, true)
+
+    const slotLocks = new Map<SessionID, Semaphore.Semaphore>()
+    const activeCounts = new Map<SessionID, number>()
+    const rowLocks = new Map<string, Semaphore.Semaphore>()
+    const partsByToolCall = new Map<
+      string,
+      { sessionID: SessionID; messageID: MessageID; partID: PartID }
+    >()
```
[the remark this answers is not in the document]
Astro-Han: Sub-second elapsed for an LLM-driven subagent dispatch is not a real case — creation alone is multiple round trips, plus the model turn. Not adding a `<1s` branch.

packages/opencode/src/tool/agent-list.ts
```diff
@@ -0,0 +1,73 @@
...
+})
+
+const formatElapsed = (ms: number): string => {
+  if (ms <= 0) return "0s"
+  const s = Math.floor(ms / 1000)
+  if (s < 3600) return `${Math.floor(s / 60)}m${String(s % 60).padStart(2, "0")}s`
+  const h = Math.floor(s / 3600)
+  const m = Math.floor((s % 3600) / 60)
+  return `${h}h${String(m).padStart(2, "0")}`
+}
```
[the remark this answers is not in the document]
Astro-Han: Sub-second elapsed for an LLM-driven subagent dispatch is not a real case — creation alone is multiple round trips, plus the model turn. Not adding a `<1s` branch.

packages/opencode/test/session/subagent-lifecycle-integration.test.ts
```diff
@@ -0,0 +1,282 @@
...
+    yield* body({ svc, sessions, parent, messageID: msg.id })
+  })
+
+const run = (program: Effect.Effect<void, unknown, SubagentRun.Service | Session.Service>) =>
+  Effect.runPromise(
+    program.pipe(
+      Effect.provide(Layer.mergeAll(SubagentRun.defaultLayer, Session.defaultLayer)),
+      Effect.orDie,
+    ),
+  )
```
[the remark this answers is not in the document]
Astro-Han: Acknowledged — could migrate to `testEffect(...)` per the path-instructions guideline. Not doing it in this PR: the existing manual `Effect.runPromise` runner works, all 27 lifecycle tests pass, and migrating risks fixture drift right before merge. Tracked for a dedicated test-harness sweep.

Astro-Han:
@coderabbitai review

packages/opencode/src/session/session.ts
```diff
@@ -496,6 +515,28 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
...
+            if (
+              lifecycleFieldsChanged(
+                existing as unknown as Record<string, unknown> | undefined,
+                part as unknown as Record<string, unknown>,
+              )
+            ) {
+              return yield* Effect.die(
+                new SubagentRunGuardViolation((part as { tool_call_id?: string }).tool_call_id),
+              )
+            }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in e61b826d. Session.fork() replays parts via `updatePart()` outside any `SubagentRunWriterContext`, so a forked session containing a completed/failed/canceled subagent did die on `SubagentRunGuardViolation`. The guard now only polices mutations of an existing row — first writes (existing === undefined) pass through, which is the right semantics for fork / migration / import. New regression test in `subagent-run.test.ts` ("allows first-write of a SubtaskPart with persisted lifecycle values (Session.fork compatibility)") locks this in.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,423 @@
...
```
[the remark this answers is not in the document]
Astro-Han: Push back. `tool_call_id` is the AI-SDK-generated call id (`call_<nanoid>`) — globally unique across sessions in practice, not a per-parent name. `readByToolCallID(parentID, tool_call_id)` takes parentID so the **lookup** is parent-scoped (defense against cross-parent reads), not because the id repeats. Composite-keying the maps adds API surface for an unlikely-to-impossible collision and forces every mutator to thread parentID through. YAGNI for v1; if AI SDK ever moves off nanoid we revisit. (See also `agent_output` enforcing parent-scoped read at the call site.)

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,423 @@
...
+        }),
+      )
+
+    const patchSession = (toolCallID: string, sessionID: SessionID): Effect.Effect<void> =>
+      withExistingPart(toolCallID, (existing) =>
+        session.updatePart({
+          ...existing,
+          subagent_session_id: sessionID,
+          updated_at: Date.now(),
+        }).pipe(Effect.asVoid),
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 28e12271. `patchSession` is now single-assignment: same id → no-op, different id → `Effect.die` with a clear message. Re-pointing a row to a different child would have corrupted `findLatestBySessionID` and ownership checks; defects on the rewrite path now surface instead of silently mutating.

packages/opencode/src/tool/agent-output.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+  if (p.status === "failed") {
+    return [
+      `status: failed`,
+      `error.kind: ${p.error?.kind ?? "unknown"}`,
+      `error.message: ${p.error?.message ?? ""}`,
+    ].join("\n")
+  }
+  if (p.status === "completed_empty") return "status: completed_empty\n(no output)"
+  return p.result_text ?? p.partial_result ?? ""
```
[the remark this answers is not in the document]
Astro-Han: Fixed in bf8039f9. `formatResult` now leads canceled rows with `status: canceled_by_user` followed by `partial_result:\n…` (or `(no partial output)`), so a cancellation is no longer indistinguishable from a normal completion in `detail: "result"` mode.

packages/opencode/src/tool/agent.ts
```diff
@@ -27,16 +35,140 @@ export const Parameters = Schema.Struct({
...
+// Non-greedy JSON match so legitimate prose containing braces (e.g., a sanitized status header
+// quoted in an error message) is not stripped wholesale. Matches one envelope at a time.
+const JSON_ENVELOPE_RE = /\{[\s\S]+?\}/g
+
+const sanitizeErrorMessage = (msg: string): string =>
+  msg
+    .replace(STACK_RE, "")
+    .replace(PATH_RE, "<path>")
+    .replace(JSON_ENVELOPE_RE, "<json>")
+    .slice(0, SANITIZE_LIMIT)
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 6bf8a1ff. PATH_RE now matches POSIX (`/Users/...`, `/home/...`), Windows drive paths (`C:\Users\...`), and UNC paths (`\\server\share\...`). All collapse to `<path>`. Added invariant-style sanitize tests in `agent.test.ts` (POSIX Mac, POSIX Linux, Windows drive, Windows UNC, JSON envelope) that assert the output never contains the original sensitive substrings — they catch future leak vectors regardless of the literal replacement format.

packages/opencode/src/tool/agent.ts
```diff
@@ -54,125 +186,246 @@ export const AgentTool = Tool.define(
...
-              permission: item,
-            })) ?? []),
-          ],
-        }))
-
-      const msg = yield* Effect.sync(() => MessageV2.get({ sessionID: ctx.sessionID, messageID: ctx.messageID }))
-      if (msg.info.role !== "assistant") return yield* Effect.fail(new Error("Not an assistant message"))
+      if (params.subagent_session_id && !session) {
+        return yield* Effect.fail(new Error("subagent_session_id not found"))
+      }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 6bf8a1ff. The resume lookup now uses `Cause.squash` and only swallows `NotFoundError` (the storage-layer "missing session" defect); any other cause re-raises via `Effect.failCause(cause)` so storage / schema-decode failures propagate instead of being masked as "subagent_session_id not found".

packages/opencode/src/session/subagent-run-context.ts
```diff
@@ -0,0 +1,58 @@
...
+  "updated_at",
+  "ended_at",
+  "consumed_at",
+  "last_activity",
+  "recent_events",
+  "result_summary",
+  "result_text",
+  "partial_result",
+  "error",
+] as const
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 98ed2bd3. `LIFECYCLE_KEYS` now also covers `tool_call_id`, `parent_session_id`, `parent_message_id`, `subagent_session_id`, so a non-writer mutation of an existing row that re-points any of these now trips `SubagentRunGuardViolation`. SubagentRun's own writers stay unaffected: `start` / `recordRejected` set the fields on first write (allowed unconditionally after the fork-compat fix), and `patchSession` runs inside `SubagentRunWriterContext`.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,435 @@
...
+    const findLatestBySessionID = (
+      parentID: SessionID,
+      subagentSessionID: SessionID,
+    ): Effect.Effect<MessageV2.SubtaskPart, NotFound> =>
+      Effect.gen(function* () {
+        const all = yield* collectSubtaskParts(parentID)
+        const matches = all.filter((p) => p.subagent_session_id === subagentSessionID)
+        if (matches.length === 0) return yield* Effect.fail(new NotFound(subagentSessionID))
+        matches.sort((a, b) => (b.started_at ?? 0) - (a.started_at ?? 0))
+        return matches[0]
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 23226e69. `findLatestBySessionID` now `partsByToolCall.set(...)` before returning the match, mirroring the fallback in `readByToolCallID`. After a host restart, `agent_output` finding a row by `subagent_session_id` now also makes `setConsumed` / `recordEvent` hit the fast path instead of falling into `withExistingPart`'s NotFound no-op.

Astro-Han:
@coderabbitai review

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,447 @@
...
+  parent_session_id: SessionID
+  parent_message_id: MessageID
+  tool_call_id: string
+  description: string
+  prompt: string
+  agent: string
+  subagent_type: string
+  command?: string
+  model?: { providerID: ProviderID; modelID: ModelID }
+}
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 54bcb348. `subagent_type` is gone from `StartInput` and `RejectedInput`. SubtaskPart never carried it, and child-session tagging is already done at the call site via `sessions.create({ subagentType })` in `agent.ts` — keeping it on the SubagentRun DTOs implied a tagging responsibility this layer never owned. Two `agent.ts` call sites and the direct-svc tests stop passing it.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,447 @@
...
+      list,
+    })
+  }),
+)
+
+export const defaultLayer: Layer.Layer<Service, never, never> = layer.pipe(
+  Layer.provide(Session.defaultLayer),
+)
+
+export * as SubagentRun from "./subagent-run"
```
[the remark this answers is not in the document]
Astro-Han: Push back — the analysis missed how the consumers import this module. All of them write:

```ts
import { SubagentRun } from "../session/subagent-run"
yield* SubagentRun.Service        // tool/agent.ts:168, tool/agent-output.ts:64, tool/agent-list.ts:50
SubagentRun.defaultLayer          // tool/registry.ts:410, test/session/prompt-effect.test.ts:187, test/tool/agent.test.ts:34
SubagentRun.AgentListFilter       // tool/agent-list.ts:57
```

The `import { SubagentRun }` named-binding only resolves because line 447 publishes the file as a `SubagentRun` namespace via `export * as SubagentRun from "./subagent-run"`. Drop the line and 11+ call sites break — `SubagentRun.Service`, `SubagentRun.defaultLayer`, `SubagentRun.AgentListFilter` would have nothing to bind to. This is the canonical opencode/PawWork module-as-namespace pattern (mirrors `Session`, `Agent`, `Config`); not dead code.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,445 @@
...
+    const setConsumed = (toolCallID: string): Effect.Effect<void> =>
+      withExistingPart(toolCallID, (existing) =>
+        Effect.gen(function* () {
+          if (existing.consumed_at) return
+          yield* session.updatePart({
+            ...existing,
+            consumed_at: Date.now(),
+            updated_at: Date.now(),
+          })
+        }),
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 5bcf9cfd. `finalize` now appends the matching SubtaskEvent variant (`completed` / `completed_empty` / `canceled_by_user` / `failed` with `kind`) and `setConsumed` appends `{ type: "consumed" }`. The eviction + stable-sort logic is hoisted into a shared `appendLifecycleEvent` helper so `recordEvent`, `finalize`, and `setConsumed` all cap at 20 with deterministic ordering. `recent_events` now records the full transition sequence (`started → <terminal> → consumed`).

packages/opencode/src/tool/agent.ts
```diff
@@ -27,16 +36,143 @@ export const Parameters = Schema.Struct({
...
+      const messages = yield* sessions.messages({ sessionID: childID, limit: 5 })
+      for (let i = messages.length - 1; i >= 0; i--) {
+        const m = messages[i]
+        if (m.info.role !== "assistant") continue
+        const textParts = m.parts.filter((p): p is MessageV2.TextPart => p.type === "text")
+        const completed = textParts.findLast((p) => isTextPartCompleted(p))
+        if (completed?.text && completed.text.trim().length > 0) return completed.text
+      }
+      return null
+    })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 603e973b. `makeReadLastCompletedAssistantText` drops the `limit: 5` and scans the full child history. Cancellation is a rare path so the extra read cost is acceptable; tool-heavy children that previously surfaced `partial_result: null` despite having stable assistant text now recover the latest completed text.

packages/opencode/src/tool/agent.ts
```diff
@@ -54,125 +190,250 @@ export const AgentTool = Tool.define(
...
+
+            // Wraps the success-path finalize so a transient finalize failure does not
+            // propagate to the next pipe step (Effect.catch) and re-run finalizeAfter as
+            // `kind: "err"`, which would relabel a completed run as failed. Idempotency in
+            // SubagentRun.finalize already protects the row state; this just stops the success
+            // result from being rebranded.
+            const finalizeOk = (r: MessageV2.WithParts) =>
+              finalizeAfter({ kind: "ok", r }).pipe(
+                Effect.catchCause(() => Effect.void),
+              )
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 603e973b, but not via the suggested "remove the catch" — that would re-route an ok-finalize failure into the err-finalize path and brand a successful run as failed (which 5113c96e specifically guards against). Instead, the tap+catch dance is replaced with `Effect.matchCauseEffect`:

```ts
.pipe(
  Effect.matchCauseEffect({
    onSuccess: (r) => finalizeAfter({ kind: "ok", r }),
    onFailure: (cause) => Effect.gen(function* () {
      yield* finalizeAfter({ kind: "err", error: Cause.squash(cause) }).pipe(
        Effect.catchCause(() => Effect.void),
      )
      return yield* Effect.failCause(cause)
    }),
  }),
)
```

Each path runs its matching finalize exactly once. ok-finalize failure now propagates to the outer `catchCause` (which fails the tool with the original cause) instead of being swallowed; err-finalize is best-effort cleanup so a secondary persistence failure cannot mask the original cause. `read()` no longer sees stale "running" rows after a swallowed failure.

Astro-Han:
@coderabbitai review

packages/opencode/src/session/message-v2.ts
```diff
@@ -251,6 +264,34 @@ export const SubtaskPart = PartBase.extend({
...
+  parent_session_id: z.string().optional(),
+  parent_message_id: z.string().optional(),
+  subagent_session_id: z.string().optional(),
+  // lifecycle (NEW)
+  status: z
```
Astro-Han, line 273: **[P1]** These defaults do not currently apply to persisted legacy rows in the normal read path. `MessageV2.part(row)` and `Session.getPart()` both return raw `row.data` with a type cast rather than parsing through `MessageV2.Part`, so existing subtask rows without `status` or `recent_events` can still reach `SubagentRun.list()` as `status === undefined`. In the default `agent_list` path that renders as `undefined (unread)`, which breaks the advertised backward-compat behavior for old subtasks. Please normalize subtask parts during hydration or coerce these defaults inside `collectSubtaskParts` before listing or output formatting.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,462 @@
...
+        if (idx < 0) break
+        merged.splice(idx, 1)
+      }
+      while (merged.length > 20) merged.shift()
+      const indexed = merged.map((e, i) => ({ e, i }))
+      indexed.sort((a, b) => a.e.at - b.e.at || a.i - b.i)
+      return indexed.map((x) => x.e)
+    }
+
+    const recordEvent = (toolCallID: string, event: MessageV2.SubtaskEvent): Effect.Effect<void> =>
```
Astro-Han, line 266: **[P2]** `recordEvent` appends to `recent_events`, but nothing ever writes `last_activity`, and there is no production caller wiring child tool/model events into this method. Both `agent_list` and `agent_output` read `last_activity?.label`, so a running subagent reports `latest: -` until it finishes. That leaves the new inspection tools unable to answer the advertised `what is it doing` question for long-running children. Please either derive `last_activity` here for progress events and wire the child prompt/tool loop to call it, or keep this out of the v1 surface.

packages/opencode/src/tool/agent-output.ts
```diff
@@ -0,0 +1,119 @@
...
+                .findLatestBySessionID(ctx.sessionID, params.subagent_session_id! as SessionID)
+                .pipe(Effect.catchTag("NotFound", () => Effect.succeed(null as MessageV2.SubtaskPart | null)))
+          if (!row || !row.tool_call_id)
+            return yield* Effect.fail(new Error("subagent not found or not accessible from this parent"))
+          if (row.parent_session_id !== ctx.sessionID)
+            return yield* Effect.fail(new Error("subagent not found or not accessible from this parent"))
+          if (row.subagent_session_id) {
+            // session.get throws NotFoundError, which Effect.gen wraps as a defect (Cause.Die),
+            // not a typed failure. catchCause is required to convert it back to a clean response.
+            const child = yield* sessions
```
Astro-Han, line 106: **[P2]** This `catchCause` still turns every `sessions.get()` defect into the same inaccessible-subagent path. A storage or schema-decode failure while validating the child session would be hidden as a clean access denial, which is the same masking class this PR fixed for row lookup and resume lookup. Please suppress only the genuine missing-session case, for example by squashing the cause and checking `NotFoundError`, and rethrow the original cause for everything else.

packages/opencode/src/tool/agent.ts
```diff
@@ -54,125 +194,255 @@ export const AgentTool = Tool.define(
...
+              })
+              .pipe(
+                Effect.matchCauseEffect({
+                  onSuccess: (r) => finalizeAfter({ kind: "ok", r }),
+                  onFailure: (cause) =>
+                    Effect.gen(function* () {
+                      yield* finalizeAfter({ kind: "err", error: Cause.squash(cause) }).pipe(
+                        Effect.catchCause(() => Effect.void),
+                      )
+                      return yield* Effect.failCause(cause)
```
Astro-Han, line 420: **[P2]** On child prompt failure, this path finalizes the SubtaskPart as `failed` and then rethrows the original cause, so the parent receives a tool error rather than the structured `status: failed` output that `synthesizeOutput` can render. In that error result the model does not get a usable `subagent_session_id`, so a model failure is still hard to introspect or recover from in the same turn. Consider returning the finalized row through `synthesizeOutput(finalPart, nextSession.id)` for execution errors, while preserving hard failures for persistence/finalize defects.

Astro-Han (commented):
Additional review pass focused on current head, edge cases, regressions, and cross-module impact.

packages/opencode/src/session/message-v2.ts
Astro-Han: Fixed in fff327dc. `MessageV2.parts()` and `Session.getPart()` indeed return `row.data as Part` without a parse pass, so the schema defaults never applied to legacy rows. Added a `hydrateSubtask` helper inside `SubagentRun` and applied it in both `collectSubtaskParts` (used by `list` / `findLatestBySessionID` / `readByToolCallID`) and `readPart` (used by every per-row mutator). Legacy rows now surface as `status: "completed"` and `recent_events: []` like new ones, so `agent_list` no longer renders "undefined (unread)".

packages/opencode/src/session/subagent-run.ts
Astro-Han: Removed in cb96c8ef. `last_activity` had no writer — `recordActivity` was already deleted as dead code, and v1 doesn't wire child tool / model events into the SubtaskPart. The `latest: <label>` line in agent_list / agent_output was always falling back to `-` or `result_summary` in practice. YAGNI: dropped the schema field, the readers, and its mention in `LIFECYCLE_KEYS`. agent_output's running header now surfaces `result_summary` directly (useful when a previous turn left summary text); agent_list's row activity falls back to `result_summary` or `(legacy)`. Progress reporting comes back in v1.1 alongside a real writer.

packages/opencode/src/tool/agent-output.ts
Astro-Han: Fixed in cb96c8ef. Switched the `sessions.get` cleanup to the same selective-catch pattern 6bf8a1ff applied to the resume lookup: squash the cause and only swallow `NotFoundError`, otherwise re-raise via `Effect.failCause`. Storage / schema-decode failures during child validation are no longer masked as a clean "subagent inaccessible" response.

packages/opencode/src/tool/agent.ts
Astro-Han: Push back. The current contract is intentional: child failure lands as a tool `error` with `ctx.metadata.sessionId` populated (set in step 4 of the agent flow), so the parent model sees the standard tool-error signal AND can read `subagent_session_id` from metadata to resume. The existing test `failed subtask preserves metadata on error tool state` (PR commit b50815f9, prompt-effect.test.ts:737) locks this in: it expects `tool.state.status === "error"` + `metadata.sessionId` defined.

Routing model failure through `synthesizeOutput` would flip tool state to `completed` with body text `"status: failed\nerror: ..."`, which conflates structured success with structured failure on the same `completed` channel — harder for the model to distinguish than the current standard error+metadata pattern. Added a clarifying comment in 81e39846 explaining why the err-path keeps the cause-rethrow.

packages/opencode/src/session/subagent-run.ts
```diff
@@ -0,0 +1,462 @@
...
+    const findLatestBySessionID = (
+      parentID: SessionID,
+      subagentSessionID: SessionID,
+    ): Effect.Effect<MessageV2.SubtaskPart, NotFound> =>
+      Effect.gen(function* () {
+        const all = yield* collectSubtaskParts(parentID)
+        const matches = all.filter((p) => p.subagent_session_id === subagentSessionID)
+        if (matches.length === 0) return yield* Effect.fail(new NotFound(subagentSessionID))
+        matches.sort((a, b) => (b.started_at ?? 0) - (a.started_at ?? 0))
+        const match = matches[0]
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 7ee356fa. Both `findLatestBySessionID` and `list` now pair each scanned part with its insertion index and break the `started_at` tie on index descending, matching the deterministic newest-first order the comparator chain in `recordEvent` already uses for SubtaskEvents that share `Date.now()`. Burst dispatch no longer surfaces stale rows.

packages/opencode/src/tool/agent.ts
```diff
@@ -54,125 +194,255 @@ export const AgentTool = Tool.define(
...
+                  yield* subagentRun.finalize(ctx.callID!, "failed", {
+                    error: { kind: "execution_error", message: errorMessage(cause) },
+                    ended_at: Date.now(),
+                  })
+                }
+                // Re-fail with the original cause so stack, annotations, and parallel-error
+                // metadata are preserved instead of getting flattened into Effect.fail.
+                return yield* Effect.failCause(cause)
+              }),
+            ),
```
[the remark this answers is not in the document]
Astro-Han: Fixed in c432d1b4. Outer catch now passes `Cause.squash(cause)` into `errorMessage`, matching the err-path finalize. The persisted `SubtaskPart.error.message` no longer collapses to `<json>` for pre-prompt failures.

packages/opencode/test/session/subagent-run.test.ts
```diff
@@ -0,0 +1,419 @@
...
+            description: "review",
+            prompt: "hi",
+            agent: "build",
+            model: ref,
+          })
+          // Direct write that flips status — should be rejected as a defect.
+          const exit = yield* session
+            .updatePart({ ...part, status: "completed" as const })
+            .pipe(Effect.exit)
+          expect(exit._tag).toBe("Failure")
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 027ad97a. The guard test now walks `exit.cause.reasons`, filters with `Cause.isDieReason`, and asserts the defect is a `SubagentRunGuardViolation` with `tool_call_id === "call_guard"`. A future regression that causes updatePart to fail for any other reason will fail this test loudly instead of silently passing.

PR #297, opened 2026-04-28
fix: keep todo dock collapsed while todos exist

Astro-Han:
## Summary

Keep the session Todo dock mounted whenever the current session still has todo data, and make the dock start collapsed by default.

## Why

The Todo dock was tied to the session live state, so it disappeared as soon as the turn ended. It also opened by default, which consumed space above the composer. The new behavior keeps existing todos available after the turn ends while showing only the compact collapsed row unless the user expands it.

## Related Issue

No issue. The behavior was narrow and confirmed directly in code before implementation.

## How To Verify

```bash
bun install --frozen-lockfile
bun --cwd packages/app test src/pages/session/composer/session-composer-state.test.ts
bun --cwd packages/app playwright test e2e/session/session-composer-dock.spec.ts --grep "todo dock"
git diff --check
```

## Screenshots or Recordings

Not attached. The focused E2E covers the visible collapsed, expanded, and post-turn states through the Todo dock probe.

## Checklist

### DIFF
--- a/packages/app/e2e/session/session-composer-dock.spec.ts
+++ b/packages/app/e2e/session/session-composer-dock.spec.ts
@@ -145,11 +145,11 @@ async function todoDock(page: any, sessionID: string) {
       return api
     },
     async open(todos: NonNullable<ComposerDriverState["todos"]>) {
-      await write({ live: true, todos })
+      await write({ todos })
       return api
     },
     async finish(todos: NonNullable<ComposerDriverState["todos"]>) {
-      await write({ live: false, todos })
+      await write({ todos })
       return api
     },
     async expectOpen(states: ComposerProbeState["states"]) {
@@ -170,10 +170,6 @@ async function todoDock(page: any, sessionID: string) {
         count: states.length,
         states,
       })
-      return api
-    },
-    async expectClosed() {
-      await expect.poll(read, { timeout: 10_000 }).toMatchObject({ mounted: false })
       return api
     },
     async collapse() {
@@ -600,19 +596,19 @@ test("todo dock transitions and collapse behavior", async ({ page, project }) =>
           { content: "first task", status: "pending", priority: "high" },
           { content: "second task", status: "in_progress", priority: "medium" },
         ])
-        await dock.expectOpen(["pending", "in_progress"])
-
-        await dock.collapse()
         await dock.expectCollapsed(["pending", "in_progress"])
 
         await dock.expand()
         await dock.expectOpen(["pending", "in_progress"])
 
+        await dock.collapse()
+        await dock.expectCollapsed(["pending", "in_progress"])
+
         await dock.finish([
           { content: "first task", status: "completed", priority: "high" },
           { content: "second task", status: "cancelled", priority: "medium" },
         ])
-        await dock.expectClosed()
+        await dock.expectCollapsed(["completed", "cancelled"])
       } finally {
         await dock.clear()
       }
--- a/packages/app/src/pages/session/composer/session-composer-region.tsx
+++ b/packages/app/src/pages/session/composer/session-composer-region.tsx
@@ -118,7 +118,7 @@ export function SessionComposerRegion(props: {
 
   onCleanup(clear)
 
-  const open = createMemo(() => store.ready && props.state.dock() && !props.state.closing())
+  const open = createMemo(() => store.ready && props.state.dock())
   const progress = useSpring(() => (open() ? 1 : 0), { visualDuration: 0.3, bounce: 0 })
   const value = createMemo(() => Math.max(0, Math.min(1, progress())))
   const dock = createMemo(() => (store.ready && props.state.dock()) || value() > 0.001)
--- a/packages/app/src/pages/session/composer/session-composer-state.ts
+++ b/packages/app/src/pages/session/composer/session-composer-state.ts
@@ -12,20 +12,7 @@ import { useSync } from "@/context/sync"
 import { composerDriver, composerEnabled, composerEvent } from "@/testing/session-composer"
 import { sessionPermissionRequest, sessionQuestionRequest } from "./session-request-tree"
 
-export const todoState = (input: {
-  count: number
-  done: boolean
-  live: boolean
-}): "hide" | "clear" | "open" | "close" => {
-  if (input.count === 0) return "hide"
-  if (!input.live) return "clear"
-  if (!input.done) return "open"
-  return "close"
-}
-
-const idle = { type: "idle" as const }
-
-export function createSessionComposerState(options?: { closeMs?: number | (() => number) }) {
+export function createSessionComposerState() {
   const params = useParams()
   const sdk = useSDK()
   const sync = useSync()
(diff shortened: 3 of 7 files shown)


Astro-Han:
@coderabbitai review

packages/app/src/pages/session/composer/session-composer-state.ts
```diff
@@ -143,91 +122,41 @@ export function createSessionComposerState(options?: { closeMs?: number | (() =>
...
         if (raf) cancelAnimationFrame(raf)
         raf = undefined
 
         const next = todoState({
           count,
-          done: complete,
-          live: active,
+          done: false,
+          live: false,
         })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. todoState now accepts only count, and the call site plus state tests no longer pass unused done or live values. Verified with the composer state unit test, target todo dock E2E, packages/app typecheck, and git diff check.

Astro-Han:
@coderabbitai review

Astro-Han (commented):
## Review Summary

The change correctly simplifies the todo dock lifecycle, but several issues remain — ranging from dead code to potential product behavior regressions.

### P0 — Must Fix
- **`todoState` is a trivial identity function** (`count > 0 ? "open" : "hide"`). It provides zero abstraction value and adds indirection. Inline it into the effect or delete it entirely.

### P1 — Should Fix
- **`test.live` is orphaned dead code.** `ComposerDriverState.live` is no longer consumed anywhere after `live` memo removal. Clean up `testing/session-composer.ts` to avoid misleading future test authors.
- **No mechanism to clear stale todos.** The `clear()` function that wiped orphaned todos is gone. If a model leaves todos behind, they persist forever. Confirm this is intentional and document it in code comments if so.

### P2 — Consider Fixing
- **`todoState` export still exists with a single test consumer.** After inlining, the unit test can assert the trivial logic directly. Remove the export to reduce surface area.
- **Collapsed state is not remembered.** `collapsed: true` resets on every mount. Consider persisting user preference (e.g., `localStorage`) so expanding the dock isn't lost on navigation.
- **`expectClosed()` may be dead code in E2E.** The test now uses `expectCollapsed()` at the end. Verify `expectClosed()` is still used elsewhere; if not, remove it.

### P3 — Nit
- **Commit history could be squashed.** The second commit ("simplify todo dock state input") is a fixup of the first. Per `AGENTS.md`, this should be a `fixup!` commit autosquashed before merge.

packages/app/src/pages/session/composer/session-composer-state.ts
```diff
@@ -12,20 +12,12 @@ import { useSync } from "@/context/sync"
 import { composerDriver, composerEnabled, composerEvent } from "@/testing/session-composer"
 import { sessionPermissionRequest, sessionQuestionRequest } from "./session-request-tree"
 
-export const todoState = (input: {
-  count: number
-  done: boolean
-  live: boolean
-}): "hide" | "clear" | "open" | "close" => {
+export const todoState = (input: { count: number }): "hide" | "open" => {
```
Astro-Han: **[P0]** `todoState` is now a trivial identity: `count > 0 ? "open" : "hide"`. It adds indirection with zero abstraction value. Inline it into the effect or delete it entirely.

packages/app/src/pages/session/composer/session-composer-state.ts
```diff
@@ -97,26 +89,9 @@ export function createSessionComposerState(options?: { closeMs?: number | (() =>
     return globalSync.data.session_todo[id] ?? []
```
Astro-Han, line 83: **[P1]** The `clear()` function that wiped stale todos is deleted. If a model leaves todos behind, they persist forever with no user-facing way to dismiss them. Confirm this is intentional and add a code comment explaining the policy.

packages/app/src/pages/session/composer/session-composer-region.tsx
```diff
@@ -118,7 +118,7 @@ export function SessionComposerRegion(props: {
 
   onCleanup(clear)
 
-  const open = createMemo(() => store.ready && props.state.dock() && !props.state.closing())
+  const open = createMemo(() => store.ready && props.state.dock())
```
Astro-Han, line 121: **[P2]** Removing `!props.state.closing()` is correct since `closing` was removed from the return object. Verified no other callers reference `.closing()` — good cleanup.

packages/app/src/pages/session/composer/session-todo-dock.tsx
```diff
@@ -49,7 +49,7 @@ export function SessionTodoDock(props: {
 }) {
   const language = useLanguage()
   const [store, setStore] = createStore({
-    collapsed: false,
+    collapsed: true,
```
Astro-Han, line 52: **[P2]** `collapsed: true` resets on every mount. If the user expands the dock, navigates away, and returns, it's collapsed again. Consider persisting the collapse state (e.g., `localStorage` key per session or globally) to respect user preference.

packages/app/e2e/session/session-composer-dock.spec.ts
```diff
@@ -600,19 +600,19 @@ test("todo dock transitions and collapse behavior", async ({ page, project }) =>
...
 
+        await dock.collapse()
+        await dock.expectCollapsed(["pending", "in_progress"])
+
         await dock.finish([
           { content: "first task", status: "completed", priority: "high" },
           { content: "second task", status: "cancelled", priority: "medium" },
         ])
-        await dock.expectClosed()
+        await dock.expectCollapsed(["completed", "cancelled"])
```
Astro-Han, line 611: **[P2]** `expectCollapsed(["completed", "cancelled"])` replaces `expectClosed()`. Verify `expectClosed()` is still used elsewhere in this file; if dead, remove it to reduce E2E helper surface.

packages/app/src/pages/session/composer/session-composer-state.ts
```diff
@@ -12,20 +12,12 @@ import { useSync } from "@/context/sync"
...
-  if (!input.live) return "clear"
-  if (!input.done) return "open"
-  return "close"
+  return "open"
 }
 
-const idle = { type: "idle" as const }
-
-export function createSessionComposerState(options?: { closeMs?: number | (() => number) }) {
+export function createSessionComposerState() {
```
Astro-Han, line 15: **[P1]** Removing `options` is a breaking change for any test/mock passing `closeMs`. `session.tsx` is fine, but scan for other callers.

packages/app/src/pages/session/composer/session-composer-state.ts
```diff
@@ -97,26 +89,9 @@ export function createSessionComposerState(options?: { closeMs?: number | (() =>
...
-  const live = createMemo(() => {
-    if (test.on && test.live !== undefined) return test.live
-    return busy() || blocked()
-  })
-
   const [store, setStore] = createStore({
     responding: undefined as string | undefined,
-    dock: todos().length > 0 && live(),
-    closing: false,
+    dock: todos().length > 0,
```
Astro-Han, line 88: **[P2]** Store initializer `dock: todos().length > 0` reads `todos()` at creation time. If todos load async after mount, this is fine because the `on` effect tracks length changes — but the initial flash of `false` before the effect fires could cause a layout shift. Consider defaulting to `true` or document why `false` is safe.

packages/app/src/pages/session/composer/session-composer-state.ts
```diff
@@ -143,91 +118,37 @@ export function createSessionComposerState(options?: { closeMs?: number | (() =>
...
-  }
-
   createEffect(
     on(
-      () => [todos().length, done(), live()] as const,
-      ([count, complete, active]) => {
+      () => todos().length,
+      (count) => {
         if (raf) cancelAnimationFrame(raf)
         raf = undefined
```
Astro-Han, line 123: **[P2]** `todoState` is only used here and in tests. After inlining the trivial logic, you can drop the exported function and have the test assert `count === 0 ? "hide" : "open"` directly.

packages/app/src/pages/session/composer/session-composer-region.tsx
Astro-Han: P3 verified cleanup. No code change needed here; rg confirms no remaining state.closing or closing() callers.

packages/app/src/pages/session/composer/session-composer-state.ts
Astro-Han: P3 rejected under YAGNI. Initial false is the safer empty state while todo data is unknown; the effect opens the dock when todos arrive. Defaulting true would risk flashing an empty dock, which is outside the current fix.

Issue #128, opened 2026-04-22
[Feature] Rename and rewrite task as subagent


Astro-Han:
## What task are you trying to do?

Rename the helper tool currently registered as `task` to `agent`. The new name is consistent with Claude Code, Cursor, Codex CLI conventions; it removes the overlap with `todowrite` semantics; and it now fits because #239 has freed the user-visible primary-agent surface.

This issue picks up after #239 closed (PR #242 merged 2026-04-26) which already removed the visible primary-agent picker AND landed two prerequisite cleanups in `task.txt` (commit 19236a826a):

- removed the "Launch multiple agents concurrently" guidance
- removed the fictional greeting-responder / code-reviewer examples

## What do you do today?

The model-facing tool id is still `task`. Its parameter is `task_id`. The permission namespace key is `"task"`. The settings UI translates this as "Task" / "任务". The tool's output wraps subagent text in a `<task_result>` XML tag. `pawwork.txt` does not yet define how `question` / `todowrite` / `agent` / `skill` divide responsibilities, so weak models delegate work that should stay inline or open a checklist for one-shot answers.

Beyond the user-visible surface, `task` is wired into the codebase as a structural slot name (registry state field, named() return shape, recursion-guard map key, permission config schema field, multiple model-facing prompt strings, references inside other tool descriptions, references inside the agent generator template, the actual subagent renderer in `packages/ui/src/components/message-part.tsx`, the inline icon registry, and dev-tool fixtures). A naive rename of the tool id alone would compile but crash at runtime on first subagent dispatch, AND would cause new sessions' subagent cards to render as generic tool cards instead of the special subagent UI.

## What would a good result look like?

The model sees a single tool named `agent` with parameter `subagent_session_id` (for resuming the prior subagent session) and `subagent_type` (kept; preserves Claude-Code-aligned training signal). Settings shows "Subagent" / "子智能体". `pawwork.txt` ships a `# Tool collaboration` section that tells the model when to use each of the four helpers and when to keep work in the main thread.

[...]

PR #301, opened 2026-04-28
fix(upstream): permission config + provider/runtime fixes (PR5/5)

Astro-Han:
Replacement for #272, which was closed accidentally after its stacked base #270 had already merged. This PR reapplies only the four PR5 commits on top of current dev, so it excludes the already-merged PR4/tool-framework base changes.

## Summary

Stage F + G + H + I of the upstream sync from anomalyco/opencode v1.14.20..17701628bd. Reopened on top of dev after #270 merged. 4 commits, 11 files, all PawWork carve-outs preserved.

## Stages included

- **Stage F — Permission config** (1 file)
  - #24222 fix permission config order (drop wildcard sort, preserve user key order)
  - #24235 IntelliSense for all tool permission keys via InfoZod override
- **Stage G.1 — Provider tail** (8 files)
  - DeepSeek (#24146 #24157 #24163 #24180), OpenRouter sdk bump #24435, Codex #23925 (already incorporated in PawWork's helper), Mistral #23735, Kimi #23696, Vertex #24573, OpenAI #24063 #24212, reasoning transforms split #24574
- **Stage G.2 — Runtime** (2 files)
  - #24215 shell cwd: pass cwd as positional arg ('$1') instead of capturing $PWD pre-init
  - #23770 `tool_output` config schema (max_lines / max_bytes)
  - #23797 BOM, #19054 bare-repo, #23870 compaction already in HEAD (no-op)
- **Stage I — SDK regen** (1 file: types.gen.ts)
  - Reflects PermissionConfig intersection record + Config.tool_output

## Stage G.3 + H — deferred to follow-up

- **G.3 (#24482 agent-create deny default)**: upstream rewrote `agent.ts` from tools-based to permissions-based selection; PawWork still uses tools-based path. Fix needs reapplication on PawWork's own structure.
- **H (#20602 shell selection)**: upstream rewrote `shell.ts` BLACKLIST/LOGIN/POSIX into META dict + heavy pty refactor; PawWork has structural divergence. Slice does not compose without porting full upstream architecture.

Both tracked in follow-up task to re-sync tool .ts files (descriptions stay carve-out, logic syncs upstream).

## Carve-outs preserved

- `packages/opencode/src/permission/index.ts` — kept on PawWork (legacy task→agent alias, list() endpoint)
- `packages/opencode/src/plugin/codex.ts` — kept on PawWork (helper refactor, functionally equivalent)
- `packages/opencode/src/session/compaction.ts` + `agent/prompt/compaction.txt` — kept on PawWork (anchored summary template diverged from upstream redesign)
- `packages/opencode/src/{format,patch}/index.ts`, `session/message-v2.ts`, `tool/{bash,write,edit,truncate,apply_patch}.ts` — kept on PawWork structure (namespace wrappers, Zod params, divergent format.file API)
- `packages/opencode/package.json` — kept PawWork workspace deps; only @openrouter bumped to 2.8.1
- `packages/app/src/components/settings-general.tsx` + `packages/app/src/i18n/en.ts` + `packages/web/src/content/docs/*.mdx` — dropped per UI rewrite + web docs carve-out
- Tests kept on HEAD where PawWork tests have diverged

## Test plan

- [x] `git diff --check origin/dev...HEAD`
- [x] `bun run --cwd packages/opencode typecheck`
- [x] `bun run --cwd packages/sdk/js typecheck`
- [ ] Reviewer to verify permission config IntelliSense in editor
- [ ] Reviewer to verify shell cwd fix in interactive bash session

Refs #209
Closes #272

## Summary by CodeRabbit

* **Improvements**
  * Enhanced DeepSeek reasoning for v4+ (adds new effort levels including "max")
  * More accurate Mistral small-model reasoning detection and narrower exposure of effort variants
  * Provider-specific trimming of empty message content and adjusted tool/streaming behavior
  * Targeted enablement for Kimi K2.5 variants

* **Bug Fixes**
  * Shell session wrapper now reliably changes directory and uses an updated invocation marker
  * Improved server-error handling to produce retryable API errors

* **Tests**
  * Expanded tests for provider variants, reasoning behavior, and error scenarios

* **Chores**
  * Test coverage and variant assertions updated


packages/opencode/src/provider/transform.ts
```diff
@@ -527,7 +572,11 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
     case "venice-ai-sdk-provider":
     // https://docs.venice.ai/overview/guides/reasoning-models#reasoning-effort
     case "@ai-sdk/openai-compatible":
-      return Object.fromEntries(WIDELY_SUPPORTED_EFFORTS.map((effort) => [effort, { reasoningEffort: effort }]))
+      const efforts = [...WIDELY_SUPPORTED_EFFORTS]
+      if (model.api.id.includes("deepseek-v4")) {
```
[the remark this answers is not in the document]
Astro-Han: Adjusted, but not exactly as suggested. I replaced the hardcoded `deepseek-v4` check with parsed DeepSeek major version `>= 4`; I did not use `>= 3` because this function intentionally returns `{}` for `deepseek-v3` earlier in `variants()`. Added a regression test covering v3 exclusion and v4/v5 `max` support.

packages/opencode/src/provider/transform.ts
```diff
@@ -527,7 +572,11 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
...
+        efforts.push("max")
+      }
+      return Object.fromEntries(efforts.map((effort) => [effort, { reasoningEffort: effort }]))
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Wrapped the `@ai-sdk/openai-compatible` and `@ai-sdk/groq` switch cases in blocks so their `const` declarations are scoped to the case.

packages/opencode/src/provider/transform.ts
```diff
@@ -705,7 +754,14 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
...
       // https://v5.ai-sdk.dev/providers/ai-sdk-providers/mistral
-      return {}
+      // https://docs.mistral.ai/capabilities/reasoning/adjustable
+      if (!model.capabilities.reasoning) return {}
+      // Only Mistral Small 4 supports reasoning (mistral-small-2603, mistral-small-latest)
+      const mistralId = model.api.id.toLowerCase()
+      if (!mistralId.includes("mistral-small-2603") && !mistralId.includes("mistral-small-latest")) return {}
+      return {
+        high: { reasoningEffort: "high" },
+      }
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Wrapped the Mistral switch case in a block so `mistralId` is scoped to that case.

packages/opencode/src/config/permission.ts
```diff
@@ -65,6 +58,35 @@ const InputSchema = Schema.Union([Action, InputObject])
...
+          websearch: zod(Action).optional(),
+          codesearch: zod(Action).optional(),
+          lsp: zod(Rule).optional(),
+          doom_loop: zod(Action).optional(),
+          skill: zod(Rule).optional(),
+        })
+        .catchall(zod(Rule)),
+    ),
+  ])
+  .transform(normalizeInput)
```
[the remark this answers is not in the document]
Astro-Han: Fixed. `InfoZod` now uses `z.object(...).catchall(zod(Rule))` directly, which keeps known-key ordering without the intersection layer. The CI-failing key-order test now passes locally.

packages/opencode/src/config/permission.ts
```diff
@@ -65,6 +58,35 @@ const InputSchema = Schema.Union([Action, InputObject])
...
+          external_directory: zod(Rule).optional(),
+          todowrite: zod(Action).optional(),
+          question: zod(Action).optional(),
+          webfetch: zod(Action).optional(),
+          websearch: zod(Action).optional(),
+          codesearch: zod(Action).optional(),
+          lsp: zod(Rule).optional(),
+          doom_loop: zod(Action).optional(),
+          skill: zod(Rule).optional(),
+        })
```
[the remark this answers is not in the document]
Astro-Han: Fixed. Added canonical `agent` to the explicit `InfoZod` object while keeping legacy `task` accepted for compatibility.

Astro-Han (commented):
## Review Summary

Upstream sync PR with 4 commits, 11 files. Carve-outs are correctly preserved. A few issues found ranging from P0 (blocking lint) to P3 (nits). Details in inline comments below.

### P0 — Blocking
- **Biome lint failure**: `noSwitchDeclarations` errors in `transform.ts` on the `@ai-sdk/openai-compatible` and `@ai-sdk/mistral` switch cases. Both declare `const` without block braces. CI will fail.

### P1 — Functional / Regression Risk
- **`permission.ts` InfoZod missing canonical `agent` key**: The Zod override exposes the legacy `task` key but omits `agent`, breaking IntelliSense for the canonical permission key post-rename (#128). This contradicts the carve-out comment in the Effect schema.
- **`transform.ts` variants exclusion list drops `mistral`**: Removing `id.includes("mistral")` from the early-return list means Mistral models now fall through to provider-specific variant logic. While the new `@ai-sdk/mistral` case handles this, models routed through OpenRouter or other proxies with `mistral` in their `model.id` may now receive unexpected `reasoningEffort` options. Verify no regression for `openrouter/mistral-*` models.
- **`transform.ts` kimi matching broadened invariants**: `k2p5` → `k2p` and `kimi-k2.5` → `kimi-k2.` widens the match. This could catch unintended model IDs (e.g. future `kimi-k2.9`). Consider whether the broader pattern is intentional or should be anchored.

### P2 — Code Quality / Maintenance
- **`transform.ts` anthropic/bedrock duplication**: The upstream split duplicated ~20 lines of identical message-filtering logic. Extract a shared helper to keep the divergence maintainable.
- **`transform.ts` deepseek reasoning injection runs before interleaved transform**: Empty reasoning parts are injected at line 200, then the interleaved field extractor (line 218) strips them out for non-OpenRouter providers. The ordering is correct but fragile — a future reordering could double-inject or lose reasoning. Add a brief comment explaining the sequencing constraint.
- **`transform.ts` `@ts-expect-error` in codex.ts**: The `input: 272_000` limit field uses `@ts-expect-error` with a vague justification ("incorrect type for v1 sdk but works"). If the SDK types change, this suppresses legitimate type errors. Consider a more precise suppression or a runtime type guard.

### P3 — Nits
- **types.gen.ts `tool_output` schema added but not wired**: The SDK types now expose `Config.tool_output`, but the PR description notes truncate.ts wiring is deferred. Fine for a staged sync, but ensure the follow-up task (#19) is tracked.
- **PR test plan incomplete**: Two reviewer checkboxes remain unchecked (permission IntelliSense, shell cwd interactive). Not blocking, but worth confirming before merge.

packages/opencode/src/provider/transform.ts
```diff
@@ -52,7 +52,28 @@ function normalizeMessages(
 ): ModelMessage[] {
   // Anthropic rejects messages with empty content - filter out empty string messages
   // and remove empty text/reasoning parts from array content
-  if (model.api.npm === "@ai-sdk/anthropic" || model.api.npm === "@ai-sdk/amazon-bedrock") {
+  if (model.api.npm === "@ai-sdk/anthropic") {
```
Astro-Han, line 79: **[P2]** Code duplication: the Anthropic and Bedrock message-filtering blocks (lines 55-73 and 76-94) are identical. Upstream split them with a comment "Bedrock specific transforms" but the logic is the same. Consider extracting a shared helper (e.g. `filterEmptyContent`) so future divergence is intentional and visible.

packages/opencode/src/provider/transform.ts
```diff
@@ -175,7 +196,29 @@ function normalizeMessages(
     return result
   }
 
-  if (typeof model.capabilities.interleaved === "object" && model.capabilities.interleaved.field) {
+  // Deepseek requires all assistant messages to have reasoning on them
+  if (model.api.id.includes("deepseek")) {
```
Astro-Han: **[P2]** Sequencing fragility: this deepseek reasoning injection runs before the interleaved-field extractor at line 218. For non-OpenRouter providers, the extractor strips reasoning parts and moves them into `providerOptions.openaiCompatible[field]`. The ordering happens to be correct, but a future refactor could reorder these blocks and cause double-injection or lost reasoning. Add a comment explaining that deepseek injection must precede interleaved extraction.

packages/opencode/src/provider/transform.ts
```diff
@@ -400,12 +443,14 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
   const id = model.id.toLowerCase()
   const adaptiveEfforts = anthropicAdaptiveEfforts(model.api.id)
   if (
-    id.includes("deepseek") ||
+    id.includes("deepseek-chat") ||
```
Astro-Han: **[P1]** Regression risk: `id.includes("mistral")` was removed from this early-return list. Mistral models now fall through to the provider-specific switch below. For `@ai-sdk/mistral` this is handled correctly (new case at line 755), but models routed through OpenRouter with `mistral` in their `model.id` (e.g. `openrouter/mistral-large`) may now receive unexpected `reasoningEffort` options from the `@openrouter/ai-sdk-provider` or `@ai-sdk/openai-compatible` cases. Verify no regression for proxy-routed Mistral models.

packages/opencode/src/provider/transform.ts
```diff
@@ -400,12 +443,14 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
...
+    id.includes("deepseek-reasoner") ||
+    id.includes("deepseek-r1") ||
+    id.includes("deepseek-v3") ||
     id.includes("minimax") ||
     id.includes("glm") ||
-    id.includes("mistral") ||
     id.includes("kimi") ||
```
Astro-Han, line 449: **[P1]** Match broadening: `id.includes("kimi")` now matches any model ID containing "kimi" (e.g. `kimi-k1`, future `kimi-moonshot`). Previously the list was more specific. Combined with `k2p` (was `k2p5`) and `kimi-k2.` (was `kimi-k2.5`), this widens the exclusion. Confirm this is intentional — if a future non-reasoning kimi model is added, it will silently lose variant support here.

packages/opencode/src/provider/transform.ts
```diff
@@ -527,7 +572,11 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
     case "venice-ai-sdk-provider":
     // https://docs.venice.ai/overview/guides/reasoning-models#reasoning-effort
     case "@ai-sdk/openai-compatible":
-      return Object.fromEntries(WIDELY_SUPPORTED_EFFORTS.map((effort) => [effort, { reasoningEffort: effort }]))
+      const efforts = [...WIDELY_SUPPORTED_EFFORTS]
```
Astro-Han: **[P0]** Biome `noSwitchDeclarations` error: `const efforts` is declared directly under a `case` label without block braces. The binding is hoisted to the entire switch scope, which Biome flags as an error. Wrap this case (and the `@ai-sdk/mistral` case at line 760) in `{ ... }` blocks. Note: CodeRabbit already flagged this same issue.

packages/opencode/src/provider/transform.ts
```diff
@@ -705,7 +754,14 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
 
     case "@ai-sdk/mistral":
       // https://v5.ai-sdk.dev/providers/ai-sdk-providers/mistral
-      return {}
+      // https://docs.mistral.ai/capabilities/reasoning/adjustable
+      if (!model.capabilities.reasoning) return {}
+      // Only Mistral Small 4 supports reasoning (mistral-small-2603, mistral-small-latest)
+      const mistralId = model.api.id.toLowerCase()
```
Astro-Han, line 759: **[P0]** Biome `noSwitchDeclarations` error: `const mistralId` is declared directly under a `case` label without block braces. Same issue as the `@ai-sdk/openai-compatible` case at line 575. Wrap in `{ ... }`. Note: CodeRabbit already flagged this same issue.

packages/opencode/src/config/permission.ts
```diff
@@ -65,6 +58,35 @@ const InputSchema = Schema.Union([Action, InputObject])
 const normalizeInput = (input: Schema.Schema.Type<typeof InputSchema>): Schema.Schema.Type<typeof InputObject> =>
   typeof input === "string" ? { "*": input } : input
 
+const InfoZod = z
+  .union([
+    zod(Action),
+    z.intersection(
+      z.record(z.string(), zod(Rule)),
+      z
+        .object({
```
Astro-Han: **[P1]** Missing canonical key: the `InfoZod` override exposes the legacy `task` key (line 72) but omits `agent`. The Effect schema above (line 38) correctly uses `agent` as the canonical field post-rename (#128). Users relying on the Zod override for IntelliSense will not see `agent` as an available key. Add `agent: zod(Rule).optional()` before `task`. Note: CodeRabbit already flagged this same issue.

packages/opencode/src/provider/transform.ts
```diff
@@ -843,11 +903,11 @@ export function options(input: {
     }
   }
 
-  // Enable thinking by default for kimi-k2.5/k2p5 models using anthropic SDK
+  // Enable thinking by default for kimi models using anthropic SDK
   const modelId = input.model.api.id.toLowerCase()
   if (
     (input.model.api.npm === "@ai-sdk/anthropic" || input.model.api.npm === "@ai-sdk/google-vertex/anthropic") &&
-    (modelId.includes("k2p5") || modelId.includes("kimi-k2.5") || modelId.includes("kimi-k2p5"))
+    (modelId.includes("k2p") || modelId.includes("kimi-k2.") || modelId.includes("kimi-k2p"))
```
Astro-Han: **[P1]** Match broadening: `modelId.includes("k2p")` replaces the previous `k2p5` check, and `kimi-k2.` replaces `kimi-k2.5`. This widens the match to any model ID containing "k2p" (e.g. `some-k2proxy-model`) or starting with `kimi-k2.` (e.g. `kimi-k2.9`). If upstream intended to match only kimi-k2.5 and k2p5 variants, the patterns should be anchored more precisely.

packages/opencode/src/plugin/codex.ts
```diff
@@ -449,6 +449,16 @@ export async function CodexAuthPlugin(input: PluginInput): Promise<Hooks> {
             output: 0,
             cache: { read: 0, write: 0 },
           }
+
+          // gpt-5.5 models temporarily have restricted context window size for codex plans
+          if (model.id.includes("gpt-5.5")) {
+            model.limit = {
```
Astro-Han: **[P2]** `@ts-expect-error` justification is too vague: "incorrect type for v1 sdk but works" does not explain what type mismatch is being suppressed. If the OpenAI SDK types change in a future release, this suppression could hide legitimate type errors. Replace with a more specific explanation (e.g. which field is missing from the type definition) or add a runtime validation guard.

packages/sdk/js/src/v2/gen/types.gen.ts
```diff
@@ -1554,6 +1556,19 @@ export type Config = {
      */
     url?: string
   }
+  /**
```
Astro-Han, line 1559: **[P3]** Schema added but not wired: `Config.tool_output` is now in the generated SDK types, but the PR description notes that `truncate.ts` wiring is deferred to a follow-up. This is acceptable for a staged sync, but ensure the deferred task (tracked in #19) is not lost.

packages/opencode/src/provider/transform.ts
Astro-Han: P2, fixed in `3f7a8664f`. The common Anthropic and Bedrock empty-content rule is now extracted into `filterEmptyMessageContent`, while the provider branches stay separate for future provider-specific transforms.

packages/opencode/src/provider/transform.ts
Astro-Han: P2, fixed in `3f7a8664f`. Added a comment on the ordering constraint: DeepSeek reasoning injection must run before interleaved-field extraction so the injected empty reasoning part becomes `providerOptions.reasoning_content`.

packages/opencode/src/provider/transform.ts
Astro-Han: P1, fixed in `3f7a8664f`. Proxy-routed Mistral models are excluded from generic `reasoningEffort` variants unless the SDK package is `@ai-sdk/mistral`; added a regression test for an OpenAI-compatible Mistral model.

packages/opencode/src/provider/transform.ts
Astro-Han: P1, fixed in `3f7a8664f`. I did not broaden this further. The K2P side was restored from `k2p` to `k2p5`; `kimi` itself was pre-existing in this early exclusion list, so I left that baseline unchanged.

packages/opencode/src/provider/transform.ts
Astro-Han: P0, already fixed in `fb5bfee70` and still present at current head. The `@ai-sdk/openai-compatible` case body is scoped with braces, so Biome no longer reports `noSwitchDeclarations` here.

packages/opencode/src/provider/transform.ts
Astro-Han: P0, already fixed in `fb5bfee70` and still present at current head. The `@ai-sdk/mistral` case body is scoped with braces, so Biome no longer reports `noSwitchDeclarations` here.

packages/opencode/src/config/permission.ts
Astro-Han: P1, already fixed in `fb5bfee70`. The permission config now includes canonical `agent` while keeping legacy `task` compatibility.

packages/opencode/src/provider/transform.ts
Astro-Han: P1, fixed in `3f7a8664f`. Restored the previous explicit `k2p5`, `kimi-k2.5`, and `kimi-k2p5` matching, with regression coverage proving unrelated K2/K2P names do not get thinking injected.

packages/opencode/src/plugin/codex.ts
Astro-Han: P2, fixed in `3f7a8664f`. The `@ts-expect-error` now names the exact type gap: v1 SDK `Model.limit` omits `input`, but Codex plan models use it downstream.

packages/sdk/js/src/v2/gen/types.gen.ts
Astro-Han: P3, no code change. This PR intentionally stages the generated `Config.tool_output` schema as part of the upstream sync without widening SDK wiring in this slice; the PR body keeps the deferred follow-up scoped separately.

packages/opencode/src/provider/transform.ts
```diff
@@ -788,6 +840,10 @@ export function options(input: {
 }): Record<string, any> {
   const result: Record<string, any> = {}
 
+  if (input.model.api.npm === "@ai-sdk/google-vertex/anthropic") {
+    result["toolStreaming"] = false
+  }
```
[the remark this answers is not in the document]
Astro-Han: P3, fixed in `80329b0bf`. Added a direct `ProviderTransform.options` regression test for `@ai-sdk/google-vertex/anthropic` asserting `toolStreaming` is `false`.

Astro-Han:
@coderabbitai review

Astro-Han:
@coderabbitai review

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -781,6 +781,79 @@ describe("session.message-v2.toModelMessage", () => {
...
+              openrouter: {
+                reasoning_details: reasoningDetails,
+              },
+            },
+          },
+          { type: "text", text: "answer" },
+        ],
+      },
+    ])
+  })
```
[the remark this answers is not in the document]
Astro-Han: P3, fixed in `c6e6ae7cb`. Verified the duplicate was real and removed the second `preserves OpenRouter reasoning details through provider transform` block; one copy remains.

packages/opencode/src/session/prompt.ts
```diff
@@ -1005,7 +1007,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
           const cmd = ChildProcess.make(sh, args, {
             cwd,
             extendEnv: true,
-            env: { ...shellEnv.env, TERM: "dumb" },
+            env: { ...shellEnv.env, OPENCODE_SHELL_CWD: cwd, TERM: "dumb" },
```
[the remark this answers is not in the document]
Astro-Han: P3, fixed in `0366c754c`. Not YAGNI: this avoids leaking an internal cwd helper into shell branches that do not consume it. `OPENCODE_SHELL_CWD` is now only injected for bash/zsh; other shells keep the existing `TERM` and plugin env behavior.

packages/opencode/src/session/prompt.ts
```diff
@@ -968,10 +968,11 @@ NOTE: At any point in time through this workflow you should feel free to ask the
                 `
                   [[ -f ~/.zshenv ]] && source ~/.zshenv >/dev/null 2>&1 || true
                   [[ -f "\${ZDOTDIR:-$HOME}/.zshrc" ]] && source "\${ZDOTDIR:-$HOME}/.zshrc" >/dev/null 2>&1 || true
-                  cd -- ${JSON.stringify(cwd)}
+                  cd -- "$OPENCODE_SHELL_CWD"
+                  unset OPENCODE_SHELL_CWD
                   eval ${JSON.stringify(input.command)}
```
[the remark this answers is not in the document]
Astro-Han: P2, fixed in `0366c754c`. Verified the wrapper could continue to `eval` after a failed restore `cd`; bash/zsh now exit immediately if restoring `OPENCODE_SHELL_CWD` fails, so the user command cannot run from the wrong directory.

PR #334, opened 2026-04-29
feat: replace in-app Toast notifications with native OS notifications

### LINKED ISSUE #332: [Feature] Replace in-app Toast notifications with native OS notifications

I want to receive notifications for permission requests, questions from agents, and session status updates without having my UI blocked by in-app Toast popups.

## What do you do today?

Currently, notifications appear as Toast popups in the bottom-right corner of the app. These popups:
- Block the send/submit buttons
- Create visual noise when multiple notifications stack
- Mix important (permission requests) with less important (status updates) notifications

Replace in-app Toast notifications with native OS notifications (macOS Notification Center / Windows Notification Area) + Dock/taskbar flash feedback:
- System notifications fire regardless of window focus
- Dock icon bounces once on macOS
- Taskbar button flashes once on Windows
- Click notification → bring window to front + navigate to session
- Fallback to Toast if OS notification permission is denied

Both

## Extra context

Design document: `docs/superpowers/specs/2026-04-29-notification-system-redesign.md`

### Key changes:
1. Remove focus check in `packages/app/src/entry.tsx` and `packages/desktop-electron/src/renderer/index.tsx`
2. Remove notification-related Toast call in `packages/app/src/pages/layout.tsx` (line 506)
3. Add `flashFrame` IPC to preload bridge (`packages/desktop-electron/src/preload/index.ts` and `types.ts`)
4. Add `flash-frame` IPC handler in `packages/desktop-electron/src/main/ipc.ts`
5. Implement platform-specific flash behavior:
   - macOS: `app.dock.bounce()`
   - Windows/Linux: `win.flashFrame(true)` + 1-second timer to `win.flashFrame(false)`

### Edge cases addressed:
- OS notification permission denied → Fallback to Toast
- Multiple rapid notifications → Existing 5-second cooldown
- Multi-window scenario → `BrowserWindow.fromWebContents(event.sender)`

Astro-Han:
## Summary

Replace in-app Toast notifications with native OS notifications + Dock/taskbar flash feedback. This removes the annoying bottom-right popups that block UI elements.

Changes:
- Remove focus checks from both web and desktop `notify` functions so system notifications fire regardless of window focus
- Add `flashFrame` IPC with platform-specific behavior (macOS: Dock bounce, Windows/Linux: taskbar flash)
- Remove notification-related Toast calls and associated dead code from layout.tsx
- Add fallback to Toast when OS notification permission is denied (web platform)

## Why

The current notification system uses in-app Toast popups in the bottom-right corner that:
- Block the send/submit buttons
- Create visual noise when multiple notifications stack
- Mix important (permission requests) with less important (status updates) notifications

Users want native OS notifications like Codex App, which don't interfere with the app UI.

## Related Issue

Closes #332

## How To Verify

```bash
```

Manual testing:
1. Run `bun run dev:desktop`
2. Trigger a permission request or question from an agent
3. Verify system notification appears (not Toast popup)
4. Verify Dock icon bounces (macOS) or taskbar flashes (Windows)
5. Verify click notification brings window to front and navigates to session
6. Deny OS notification permission → verify Toast fallback works

## Screenshots or Recordings

N/A — notification behavior change, no visual UI change

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Desktop attention indicator added (Dock bounce on macOS, taskbar flash on Windows/Linux).

* **Improvements**
  * If OS notification permission is explicitly denied, a warning is logged and an in-app toast is shown.
  * Notifications are not sent for the currently active session to avoid distractions; throttling still applies.
  * Native notification delivery now uses a safe fallback and always attempts to attract attention.

* **Bug Fixes**
  * Removed persistent per-session toasts so alerts no longer linger.

### DIFF
--- a/packages/app/src/pages/layout.tsx
+++ b/packages/app/src/pages/layout.tsx
@@ -419,17 +419,8 @@ export default function Layout(props: ParentProps) {
 
   const useSDKNotificationToasts = () =>
     onMount(() => {
-      const toastBySession = new Map<string, number>()
       const alertedAtBySession = new Map<string, number>()
       const cooldownMs = 5000
-
-      const dismissSessionAlert = (sessionKey: string) => {
-        const toastId = toastBySession.get(sessionKey)
-        if (toastId === undefined) return
-        toaster.dismiss(toastId)
-        toastBySession.delete(sessionKey)
-        alertedAtBySession.delete(sessionKey)
-      }
 
       const unsub = globalSDK.event.listen((e) => {
         if (e.details?.type === "worktree.ready") {
@@ -451,7 +442,7 @@ export default function Layout(props: ParentProps) {
         ) {
           const props = e.details.properties as { sessionID: string }
           const sessionKey = `${e.name}:${props.sessionID}`
-          dismissSessionAlert(sessionKey)
+          alertedAtBySession.delete(sessionKey)
           return
         }
 
@@ -460,7 +451,6 @@ export default function Layout(props: ParentProps) {
           e.details.type === "permission.asked"
             ? language.t("notification.permission.title")
             : language.t("notification.question.title")
-        const icon = e.details.type === "permission.asked" ? ("checklist" as const) : ("bubble-5" as const)
         const directory = e.name
         const props = e.details.properties
         if (e.details.type === "permission.asked" && permission.autoResponds(e.details.properties, directory)) return
@@ -496,44 +486,8 @@ export default function Layout(props: ParentProps) {
             void platform.notify(title, description, href)
           }
         }
-
-        const currentSession = params.id
-        if (workspaceKey(directory) === workspaceKey(currentDir()) && props.sessionID === currentSession) return
-        if (workspaceKey(directory) === workspaceKey(currentDir()) && session?.parentID === currentSession) return
-
-        dismissSessionAlert(sessionKey)
-
-        const toastId = showToast({
-          persistent: true,
-          icon,
-          title,
-          description,
-          actions: [
-            {
-              label: language.t("notification.action.goToSession"),
-              onClick: () => navigate(href),
-            },
-            {
-              label: language.t("common.dismiss"),
-              onClick: "dismiss",
-            },
-          ],
-        })
-        toastBySession.set(sessionKey, toastId)
       })
       onCleanup(unsub)
-
-      createEffect(() => {
-        const currentSession = params.id
-        if (!currentDir() || !currentSession) return
-        const sessionKey = `${currentDir()}:${currentSession}`
-        dismissSessionAlert(sessionKey)
-        const [store] = globalSync.child(currentDir(), { bootstrap: false })
-        const childSessions = store.session.filter((s) => s.parentID === currentSession)
-        for (const child of childSessions) {
-          dismissSessionAlert(`${currentDir()}:${child.id}`)
-        }
-      })
     })
 
   useUpdatePolling()
--- a/packages/app/src/pages/layout.tsx
+++ b/packages/app/src/pages/layout.tsx
@@ -479,7 +479,9 @@ export default function Layout(props: ParentProps) {
           if (settings.notifications.permissions()) {
             // Only notify for background sessions, not the one the user is currently viewing
             const currentSession = params.id
-            const isCurrentSession = workspaceKey(directory) === workspaceKey(currentDir()) && 
+            const isCurrentSession =
+              !!currentSession &&
+              workspaceKey(directory) === workspaceKey(currentDir()) &&
               (props.sessionID === currentSession || session?.parentID === currentSession)
             if (!isCurrentSession) {
               void platform.notify(title, description, href)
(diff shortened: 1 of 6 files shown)


packages/app/src/pages/layout.tsx
```diff
@@ -496,44 +486,8 @@ export default function Layout(props: ParentProps) {
             void platform.notify(title, description, href)
```
Astro-Han: The current-session guard was removed just below this block, so this path now reaches `platform.notify` even when the question belongs to the session the user is already viewing. Because this PR also removes the desktop/web foreground checks, the active session path can now produce an OS notification and Dock/taskbar attention signal for something already visible in the app. Please keep the old current-session and child-session suppression before calling `platform.notify`, then use native notifications only for background sessions.

Astro-Han (commented):
I left one inline comment for a notification regression that is not covered by the existing comments.

packages/desktop-electron/src/renderer/index.tsx
```diff
@@ -261,6 +258,9 @@ const createPlatform = (): Platform => {
```
Astro-Han, line 260: This desktop path creates the OS notification directly, but it does not handle the case where the OS refuses or cannot show notifications. In that scenario the user can lose a permission request, question, or session alert entirely. Please add a desktop fallback equivalent to the web fallback, ideally preserving the old in-app action that takes the user to the affected session.

Astro-Han (commented):
Please close the remaining notification UX gaps before this PR lands. The direction matches #332, but the implementation still needs to make the failure paths as reassuring as the main path: 1. Web fallback: when OS notification permission is denied, the fallback Toast should preserve the old ability to go to the affected session instead of only showing a passive message. 2. Desktop fallback: if desktop OS notifications are unavailable or denied, PawWork should still give the user an in-app fallback instead of silently losing the permission/question/session alert. 3. Windows/Linux flash cleanup: the delayed stop for taskbar flashing should not call into a destroyed window, and repeated notifications should reset the stop timer cleanly per window. These three items are the difference between replacing the blocking Toasts and accidentally losing important alerts when system notifications cannot be delivered.

Astro-Han:
## Review Comments Addressed

### P0 - Critical Issues

**1. Restore current-session guard** (Astro-Han, line 486)
- ✅ **Fixed**: Restored the current-session and child-session suppression before calling `platform.notify`. Now only background sessions trigger OS notifications, preventing unnecessary alerts when the user is already viewing the relevant session.

### P1 - High Priority Issues

**2. Only fallback to Toast for denied permission** (CodeRabbit, line 64-73)
- ✅ **Fixed**: Changed the condition to only show Toast fallback when `permission === "denied"`. For `default` state (user dismissed prompt), we now silently return without showing a Toast.

**3. Check win.isDestroyed() in flash timer** (Gemini, line 407-409)
- ✅ **Fixed**: Added `win.isDestroyed()` check before calling `win.flashFrame(false)` in the timeout callback to prevent crashes when the window is closed before the timer fires.

**4. Reset flash timer per window** (CodeRabbit, line 403-409)
- ✅ **Fixed**: Implemented a `flashFrameTimers` Map to track timers per window ID. Now clears existing timer before setting a new one, preventing earlier timers from canceling newer flashes under burst conditions.

### P2 - Medium Priority Issues (YAGNI Analysis)

**5. Desktop fallback for failed notifications** (Astro-Han, line 260)
- ❌ **Won't fix (YAGNI)**: On Electron, `Notification` API is typically reliable as it uses the system's native notification. If it fails, the user likely sees a system-level error. Adding fallback logic would increase complexity for an edge case that rarely occurs. If users report issues, we can add this later.

**6. Flash only when window not focused** (Gemini, line 263)
- ❌ **Won't fix (YAGNI)**: The current implementation already provides good UX - flashing when focused is a minor distraction, and the 1-second timer ensures it stops quickly. Adding focus checks would add complexity without significant benefit.

**7. Restore Toast actions and persistence** (Gemini, line 67-71)
- ❌ **Won't fix (YAGNI)**: The PR's primary goal is to replace Toasts with native OS notifications. The fallback Toast is a minimal safety net for denied permissions. Adding actions and persistence would partially defeat the purpose of removing Toasts. Users should enable OS notifications for the best experience.

### P3 - Low Priority Issues

**8. Icon inconsistency** (Gemini, line 67-71)
- ❌ **Won't fix (YAGNI)**: The `/icon.png` in web fallback is a minor inconsistency. The primary notification path (OS notifications) uses the app icon automatically. This is a cosmetic issue that doesn't affect functionality.

---

**Summary**: Fixed 4 critical/high priority issues that could cause crashes, UX regressions, or logic errors. Declined 4 medium/low priority issues based on YAGNI principle - they add complexity for edge cases that rarely occur or don't significantly improve UX.

packages/desktop-electron/src/renderer/index.tsx
Astro-Han: Fixed in c135762. Added try-catch around native Notification creation in the desktop renderer. If the native Notification constructor throws (e.g. OS notifications unavailable or denied), it falls back to IPC-based `showNotification()` which uses Electron main-process Notification. This ensures desktop users never lose permission/question/session alerts even if native renderer notifications are unavailable.

packages/app/src/pages/layout.tsx
Astro-Han: Fixed in fb2b831 and c959dd0. Restored the current-session and child-session suppression before calling platform.notify. Also added !!currentSession guard to prevent undefined === undefined comparison when params.id is not set.

PR #392, opened 2026-05-02
feat: add renderer diagnostics observability

### LINKED ISSUE #389: [Task] Add renderer diagnostics for session jump and flicker observability
## Goal
PawWork should capture enough local renderer diagnostics to debug session scroll jumps, timeline flicker, layout thrash, and similar UI regressions without relying on Computer Use, screenshots, or manual observation.

When this is done, a user can export a session or create a problem report and the artifact will include the relevant renderer evidence from the current app run.

## Scope
In scope:
- Add a short-retention global renderer diagnostics log for desktop app runtime events.
- Capture structured UI diagnostics for session rendering, scroll state, layout state, lightweight performance, and data refresh boundaries.
- Include the relevant diagnostics slice in session export JSON.
- Include a short diagnostics summary in problem reports and attach the complete diagnostics slice in the report package.
- Add a low-visibility global diagnostics export path for non-session-specific failures.
- Keep diagnostics local by default. No automatic upload.
- Keep the first version lightweight and privacy-conscious.
- Follow the current user-initiated problem-report privacy boundary: diagnostics are local until the user exports a session, exports diagnostics, or creates a problem report.

Out of scope:
- Do not record screenshots, full DOM snapshots, message text, prompt input text, file contents, API keys, tokens, or raw provider/proxy URLs.
- Do not change existing problem-report context fields such as directory and log path in this issue.
- Do not add cloud telemetry or automatic upload.
- Do not introduce a full OpenTelemetry SDK in the first version.
- Do not convert normal session export into a zip package.
- Do not fix every current scroll or flicker bug in this issue. This issue adds the observability layer needed to debug them reliably.

## Relevant files or context
Current gaps:
- Full session export is engine/session data oriented. It can explain message, part, tool, subagent, and model-run state, but not what the renderer did on screen.
- Problem reports currently collect app diagnostics, log tail, renderer error, and a limited session message export, but they do not carry a renderer UI trace.
- The app debug bar already observes useful frontend performance signals, but those signals are not persisted or exported.

Likely areas:
- `packages/app/src/components/debug-bar.tsx`
- `packages/app/src/pages/session/message-timeline.tsx`
- `packages/app/src/pages/session/session-view-controller.ts`
- `packages/app/src/testing/session-composer.ts`
- `packages/desktop-electron/src/main/problem-report.ts`
- `packages/desktop-electron/src/main/feedback.ts`
- `packages/desktop-electron/src/main/ipc.ts`
- `packages/opencode/src/session/export.ts`

Related issue:
- #214 covers LLM stream diagnostics in local session export. This issue covers renderer/UI diagnostics and should stay separate.

## Verification
- Unit test the diagnostics event sanitizer so sensitive fields are not written outside the allowlisted schema.
- Unit test main-process ingestion limits: known event names only, known fields only, per-event size limit, and simple rate limits for high-frequency events.
- Unit test hostile renderer inputs: unknown events, oversized payloads, URL-like fields, message-text-like fields, and nested unknown fields must be ignored or reduced to the allowlisted output shape before writing.
- Unit test rolling retention: maximum size, maximum age, malformed line handling, and best-effort failure behavior.
- Unit test session export when diagnostics are present, missing, expired, truncated, corrupt, disabled, or write-failed.
- Unit test problem report packaging so the human-readable report contains only a summary while the complete diagnostics slice is attached separately.
- Add an E2E regression probe for a long tool-heavy session with deterministic assertions for `timeline_mount_count`, `visible_count_min`, `scroll_anchor_before`, `scroll_anchor_after`, `scroll_top_delta`, and `user_scrolled=false`.
- Confirm normal session export remains a single JSON file and remains readable by older consumers that ignore the optional diagnostics field.

## Execution mode
Human review required before code changes.

Astro-Han:
## Summary

Adds a default-on renderer diagnostics pipeline for session rendering issues:

- Records sanitized renderer diagnostics from the session UI into a bounded Electron main-process JSONL log.
- Adds diagnostics export through Help -> Export Diagnostics Log..., session export, and problem reports.
- Emits session submit, view-state, timeline, scroll, composer dock, visibility, perf, and derived incident events.
- Adds truncation/failure handling so diagnostics never block app behavior, session export, or problem report creation.
- Adds unit and E2E coverage for the diagnostics flow and send/scroll stability.

## Why

Issue #389 needs better observability for recurring session jump-to-top and render flicker bugs. The goal is to stop relying only on Computer Use and visual inspection by giving manually shared problem reports/session exports enough structured renderer evidence to diagnose these failures.

## Related Issue

Closes #389

## Human Review Status

## Review Focus

Please focus on:

- Privacy/sanitization boundaries for renderer diagnostics fields.
- Whether diagnostics export and problem report attachment fail open correctly.
- Whether the renderer instrumentation is scoped enough and does not perturb session scrolling/rendering.
- Desktop menu copy/i18n and platform impact on macOS/Windows.

## Risk Notes

- New local diagnostic JSONL is written under Electron user data and retained with size/time caps.
- The Help menu has one visible native-menu copy change: `Export Diagnostics Log...` / `导出诊断日志...`.
- The saved diagnostics file is user-chosen via native save dialog; problem reports/session exports can include a capped diagnostics slice.
- No dependencies, migrations, credentials, deletion behavior, or generated source files were introduced.
- Labels: maintainer labeling requested.

## How To Verify

Latest local verification on head `70f8cfc79`:

```text
Desktop focused tests: 40 passed, 0 failed
bun --cwd packages/desktop-electron test src/main/renderer-diagnostics.test.ts src/main/feedback.test.ts src/main/ipc-window-config.test.ts

App focused tests: 685 passed, 0 failed
bun --cwd packages/app test src/context/renderer-diagnostics.test.ts src/pages/session/session-view-controller.test.ts

E2E focused tests: 1 passed, 0 failed
bun --cwd packages/app playwright test e2e/session/session-renderer-diagnostics.spec.ts --reporter=line

Desktop typecheck: passed

App typecheck: passed

Diff whitespace check: passed
git diff --check
```

## Screenshots or Recordings

Not captured. The visible UI change is a native menu item and native save dialog title; coverage is via menu/i18n tests. Manual visual follow-up is still useful on macOS/Windows native menus.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * "Export Diagnostics Log" menu command and renderer APIs to record/export diagnostics.

* **Enhancements**
  * Session view emits scroll, timeline, visibility and performance telemetry; derived incident signals reported and sampled.
  * Diagnostics included in feedback/problem reports and session exports with sanitization, retention, and safe truncation.
  * Composer/dock resize and scroll metrics reported; timeline mount/visible/unmount and submit events emitted.

* **Tests**
  * Extensive e2e and unit tests for capture, sanitization, retention, slicing, export, and incident detection.


packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,510 @@
...
+        const previous = lastHighFrequency.get(key)
+        if (previous !== undefined && current - previous < highFrequencyIntervalMs) {
+          return { ok: false as const, reason: "rate_limited" as const }
+        }
+        lastHighFrequency.set(key, current)
+      }
+      await enqueueWrite(async () => {
+        await mkdir(options.root, { recursive: true })
+        await appendFile(path, `${JSON.stringify(sanitized)}\n`, "utf8")
+        await flushRetention()
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: record() now appends first and only runs retention when the file exceeds the byte cap or the periodic retention interval has elapsed.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,510 @@
...
+    const cutoff = now().getTime() - retentionMs
+    const retained = events.filter((event) => eventTime(event) >= cutoff)
+    let content = retained.map((event) => JSON.stringify(event)).join("\n")
+    if (content) content += "\n"
+    while (Buffer.byteLength(content, "utf8") > maxBytes && retained.length > 0) {
+      retained.shift()
+      content = retained.map((event) => JSON.stringify(event)).join("\n")
+      if (content) content += "\n"
+    }
+    await mkdir(options.root, { recursive: true })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: retention pruning now precomputes JSON lines and byte totals instead of reserializing the retained array on each trim.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,510 @@
...
+function isAllowedEventName(name: string): name is keyof typeof eventDataFields {
+  return Object.hasOwn(eventDataFields, name)
+}
+
+function stringField(value: unknown, limit = 160) {
+  if (typeof value !== "string") return undefined
+  const next = value.replace(/\s+/g, " ").trim()
+  if (!next) return undefined
+  if (/[a-z][a-z0-9+.-]*:\/\//i.test(next)) return undefined
+  if (/\b[a-z0-9.-]+\.[a-z]{2,}(?:\/|\?|:|$)/i.test(next)) return undefined
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: the sanitizer now keeps dotted technical identifiers such as open-router.ai while still dropping URL-like values with paths, queries, ports, or schemes. Added coverage for this case.

packages/app/e2e/session/session-renderer-diagnostics.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        emitRendererDiagnostic?: (event: CapturedDiagnosticEvent) => Promise<void>
+      }
+    }
+    win.__pawwork_renderer_diagnostics = []
+    win.api = {
+      ...(win.api ?? {}),
+      emitRendererDiagnostic: async (event) => {
+        win.__pawwork_renderer_diagnostics?.push(JSON.parse(JSON.stringify(event)))
+      },
+    }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: the E2E capture now wraps the existing diagnostics emitter and delegates to it after recording the captured event.

packages/app/src/components/prompt-input/submit.ts
```diff
@@ -521,6 +522,22 @@ export function createPromptSubmit(input: PromptSubmitInput) {
...
+      data: {
+        action: "submit",
+        provider: model.providerID,
+        model: model.modelID,
+        endpoint_kind: "prompt",
+        prompt_length: input.promptLength(currentPrompt),
+        image_count: images.length,
+        comment_count: input.commentCount(),
+      },
+    })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: submit diagnostics now capture prompt/image/comment counts before clearing mutable input state and attach a catch handler to keep diagnostics fail-open.

packages/app/src/context/renderer-diagnostics.ts
```diff
@@ -0,0 +1,237 @@
...
+        frame_gap_ms: Math.round(maxFrameGap),
+        jank_count: jankCount,
+        long_task_max_ms: Math.round(longTaskMax),
+        long_task_block_ms: Math.round(longTaskBlock),
+        cls,
+        heap_used_mb: memory.memory?.usedJSHeapSize
+          ? Math.round(memory.memory.usedJSHeapSize / 1024 / 1024)
+          : undefined,
+      },
+    })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: FPS is now computed from frame count over the actual sample duration, and the sample start timestamp is reset after each flush.

packages/app/src/pages/session/message-timeline.tsx
```diff
@@ -692,6 +736,23 @@ export function MessageTimeline(props: {
...
+              data: {
+                scroll_top: el.scrollTop,
+                scroll_height: el.scrollHeight,
+                client_height: el.clientHeight,
+                distance_from_bottom: max - el.scrollTop,
+                user_scrolled: props.hasScrollGesture(),
+                jump_button_visible: props.scroll.overflow && props.scroll.jump && !staging.isStaging(),
+                ...visibleRangeData(),
+              },
+            })
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: scroll diagnostics are now requestAnimationFrame-throttled and distance_from_bottom is clamped to a non-negative value.

packages/app/src/pages/session/use-session-scroll-dock.ts
```diff
@@ -148,16 +154,29 @@ export function createSessionScrollDock(input: {
...
       fill: input.fill,
     })
+    if (dockHeight !== previousDockHeight) {
+      input.onDockHeightChange?.({
+        composerHeight: dockHeight,
+        previousComposerHeight: previousDockHeight,
+        scrollTop,
+        distanceFromBottom,
+      })
+    }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: onDockHeightChange is guarded so diagnostics callback failures cannot break dock resize or scroll recovery.

packages/desktop-electron/src/main/feedback.ts
```diff
@@ -215,10 +218,17 @@ export function createFeedbackHandler(deps: FeedbackDeps) {
       sessionExport = { status: "failed", error: errorMessage(error) }
     }
 
+    try {
+      rendererDiagnostics = await deps.rendererDiagnostics(context)
+    } catch (error) {
+      deps.onHandledError?.("renderer diagnostics slice failed", error)
+      rendererDiagnostics = emptyRendererDiagnosticsSlice("write_failed", new Date(generatedAt))
+    }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: renderer diagnostics slicing for problem reports now has a timeout guard and falls back to write_failed if it hangs.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,510 @@
...
+          return { ok: false as const, reason: "rate_limited" as const }
+        }
+        lastHighFrequency.set(key, current)
+      }
+      await enqueueWrite(async () => {
+        await mkdir(options.root, { recursive: true })
+        await appendFile(path, `${JSON.stringify(sanitized)}\n`, "utf8")
+        await flushRetention()
+      })
+      return { ok: true as const }
```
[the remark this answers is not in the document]
Astro-Han: Fixed in 1373c4ba2: record() now returns disabled before sanitizing, enqueueing, or writing when diagnostics are disabled.

Astro-Han (commented):
## PR #392 Code Review: Renderer Diagnostics Observability

Overall this is a **high-quality PR** with clear architecture, comprehensive tests, and solid fail-open design. The main concerns are around **type safety boundaries**, **performance details**, and **code organization**.

### P0 — Blocking (must fix)
1. `eventMatchesSession` accesses `data.*_session_id` fields not in the allowlist for non-transition events — silent logic failure.
2. `PerformanceObserver` uses deprecated `entryTypes` parameter — will trigger warnings in newer Chromium.
3. `flushRetention` has a race condition when called concurrently (not serialized through `enqueueWrite`).

### P1 — Serious (strongly recommended)
4. `stringField` URL regex is overly broad and drops legitimate `endpoint_kind` values.
5. `capEvents` has O(n²) JSON stringify overhead during truncation loops.
6. `session.timeline.visible` effect dependency tracking creates large strings on every render.
7. `requestAnimationFrame` in `onScroll` may fire after unmount.
8. `rendererDiagnosticsSlice` IPC handler declares unused `directory` parameter.
9. Truncation loop can delete protected events if all remaining are protected.

### P2 — Moderate (recommended)
10. `highFrequencyIntervalMs` default 250ms may miss fast scroll jumps.
11. FPS calculation is inaccurate when tab is backgrounded.
12. `emitRendererDiagnostic` silently swallows all errors even in dev.
13. Menu placement of Export Diagnostics is inconsistent with conditional Report Problem.
14. `SliceInput.now` is required in type but omitted in public API.
15. `route_ready` and `visible_ready` always have identical values.

### P3 — Minor / Style
16–24. Code organization (566-line file), test splitting, indentation consistency, threshold tuning, unit assumptions, and type safety nits.

See inline comments for specific locations and suggestions.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
...
+  if (messageID) event.message_id = messageID
+  const partID = stringField(diagnostic.part_id, 120)
+  if (partID) event.part_id = partID
+  return event
+}
+
+function parseEventLine(line: string): RendererDiagnosticEvent | undefined {
+  try {
+    const value = JSON.parse(line) as RendererDiagnosticEvent
+    if (!value || typeof value !== "object") return undefined
```
Astro-Han, line 277: P0: The allowlist for `data` fields only includes `from/to_*_session_id` on `session.identity.transition` events. For other event types, accessing `data.from_route_session_id` etc. returns `undefined` silently. This means `eventMatchesSession` silently fails to match for non-transition events that happen to carry these fields in their raw input (which get stripped by `sanitizeData`).

**Fix**: Either add these fields to all event allowlists that need session matching, or guard with `Object.hasOwn(data, key)` before comparing.

packages/app/src/context/renderer-diagnostics.ts
```diff
@@ -0,0 +1,243 @@
...
+          },
+        })
+      }
+    }
+
+    if (sessionKey && event.name === "session.timeline.visible") {
+      const before = visibleCounts.get(sessionKey) ?? 0
+      const during = renderedCount(event)
+      visibleCounts.set(sessionKey, during)
+      if (before > 0 && during === 0) {
```
Astro-Han, line 148: P0: `PerformanceObserver.observe({ entryTypes: ["longtask"] })` uses the deprecated `entryTypes` parameter. Modern browsers recommend `type: "longtask"` with `buffered: true`. This will trigger console warnings in newer Chromium and may be removed in the future.

**Fix**: Change to `observe({ type: "longtask", buffered: true })` and same for `layout-shift`. Keep the `try/catch` for fallback.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
...
+  if (status === "ok" && capped.omittedEventCount > 0) status = "truncated"
+  const output: RendererDiagnosticsExport = {
+    schema_version: 1,
+    format: "pawwork-renderer-diagnostics",
+    source: "renderer-diagnostics",
+    generated_at: (input.now ?? new Date()).toISOString(),
+    diagnostics: {
+      status,
+      event_count: capped.events.length,
+      incident_count: capped.events.filter(isIncident).length,
```
Astro-Han, line 458: P0: `flushRetention` is not serialized through `enqueueWrite`. If called concurrently (e.g. external direct call + internal `record` trigger), two reads can race against each other and the later write can overwrite the earlier one's retained events.

**Fix**: Route `flushRetention` through `enqueueWrite` or add an explicit mutex so only one flush runs at a time.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
...
+  ],
+  "session.layout.composer_dock": ["composer_height", "previous_composer_height", "scroll_top", "distance_from_bottom"],
+  "session.data.refresh": ["phase", "message_count", "part_count", "duration_ms", "cache_present"],
+  "renderer.perf.sample": [
+    "fps",
+    "frame_gap_ms",
+    "jank_count",
+    "long_task_max_ms",
+    "long_task_block_ms",
+    "cls",
```
Astro-Han, line 165: P1: The URL regex `/\b[a-z0-9.-]+\.[a-z]{2,}(?:\/|\?|:)/i` is overly broad. It drops legitimate `endpoint_kind` values like `api.example.com/v1` even though they are in the allowlist. Users with custom endpoints will silently lose this field.

**Fix**: Either restrict `endpoint_kind` to a known enum ("prompt", "continue", "edit") at the call site, or make the URL regex more specific (require protocol or `www.` prefix).

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
...
+}
+
+function isIncident(event: RendererDiagnosticEvent) {
+  return event["event.name"].startsWith("incident.")
+}
+
+export function emptyRendererDiagnosticsSlice(status: RendererDiagnosticsStatus, now: Date): RendererDiagnosticsSlice {
+  return {
+    status,
+    source: "renderer-diagnostics",
```
Astro-Han, line 341: P1: `capEvents` calls `jsonBytes(selected)` on every loop iteration, which internally does `JSON.stringify`. This is O(n²) stringification overhead when truncating large event lists.

**Fix**: Pre-compute per-event JSON byte counts and maintain a running total. Only recalculate when an event is removed.

packages/app/src/pages/session/message-timeline.tsx
```diff
@@ -246,18 +258,64 @@ export function MessageTimeline(props: {
...
+    void emitRendererDiagnostic({
+      name: "session.timeline.mount",
+      route_session_id: params.id,
+      visible_session_id: props.sessionID,
+      timeline_session_id: props.sessionID,
+      data: visibleRangeData(),
+    })
+  })
+
+  onCleanup(() => {
```
Astro-Han, line 308: P1: `rendered().join("\u0000")` creates a large string on every render to track dependency changes. With 1000+ messages this string concatenation is non-trivial overhead and runs frequently.

**Fix**: Use `on(rendered, ...)` to track the array reference directly, or if content-level comparison is truly needed, use a lightweight hash (e.g. `rendered().length`) instead of full stringification.

packages/app/src/pages/session/message-timeline.tsx
```diff
@@ -692,6 +750,31 @@ export function MessageTimeline(props: {
...
+              scrollSampleFrame = requestAnimationFrame(() => {
+                scrollSampleFrame = undefined
+                const sample = pendingScrollSample
+                pendingScrollSample = undefined
+                if (!sample) return
+                void emitRendererDiagnostic({
+                  name: "session.scroll.sample",
+                  route_session_id: params.id,
+                  visible_session_id: props.sessionID,
+                  timeline_session_id: props.sessionID,
```
Astro-Han, line 787: P1: The `requestAnimationFrame` callback can fire after component unmount. `onCleanup` at line 261 cancels the frame, but if unmount happens between `onScroll` setting `scrollSampleFrame` and the RAF firing, the callback still executes and accesses potentially stale `props`/`params`.

**Fix**: Add a `mounted` ref checked inside the RAF callback, or use an `AbortController` tied to component lifecycle.

packages/desktop-electron/src/main/ipc.ts
```diff
@@ -61,6 +66,14 @@ type Deps = {
...
   setDesktopContext: (context: DesktopContext, win: BrowserWindow) => Promise<void> | void
+  recordRendererDiagnostic: (event: unknown, context: { windowID: number }) => Promise<unknown> | unknown
+  exportRendererDiagnostics: () => Promise<{ ok: true; path: string } | { ok: false; error: string }>
+  rendererDiagnosticsSlice: (input: {
+    sessionID: string
+    directory: string
+    windowID?: number
+    maxBytes: number
+  }) => Promise<RendererDiagnosticsSlice>
 }
```
Astro-Han, line 78: P1: `rendererDiagnosticsSlice` declares a `directory: string` parameter but it is never used in the implementation (see `index.ts:1269-1274`). This creates a misleading contract for callers.

**Fix**: Remove the unused `directory` parameter from the type, or if it is meant for future use, add a TODO comment explaining its intended purpose.

packages/desktop-electron/src/main/problem-report.ts
```diff
@@ -264,6 +303,18 @@ export function buildProblemReport(input: Input, options: Options = {}) {
     output = markdown(makePayload())
   }
 
+  if (bytes(output) > maxBytes && rendererDiagnostics) {
+    const original = rendererDiagnostics
+    let events = [...original.events]
+    while (bytes(output) > maxBytes && events.length > 0) {
+      const removeIndex = events.findIndex((event) => !isProtectedRendererDiagnosticEvent(event))
+      events.splice(removeIndex >= 0 ? removeIndex : 0, 1)
+      omittedRendererDiagnosticsBytes = Math.max(0, jsonBytes(original.events) - jsonBytes(events))
```
Astro-Han, line 313: P1: If all remaining events are protected (incidents + transitions), `removeIndex` becomes `-1` and `events.splice(0, 1)` deletes a protected event. The loop then continues until `events.length === 0`, potentially discarding all incidents.

**Fix**: When no non-protected events remain, break the loop and mark status as `truncated` with a note that protected events were also removed.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
...
+
+export function rendererDiagnosticsPath(root: string) {
+  return join(root, "renderer-diagnostics.jsonl")
+}
+
+export async function exportRendererDiagnosticsLog(input: {
+  path: string
+  destination: string
+  maxBytes?: number
+  now?: Date
```
Astro-Han, line 429: P2: Default `highFrequencyIntervalMs` of 250ms may miss fast scroll jumps that resolve within a single animation frame (16ms). A jump-to-top followed by immediate auto-scroll correction could complete before the next sample window.

**Fix**: Consider lowering to 100ms for scroll events, or switch to event-driven sampling (record on significant scroll delta, not just time).

packages/app/src/context/renderer-diagnostics.ts
```diff
@@ -0,0 +1,243 @@
...
+  const api = typeof window === "undefined" ? undefined : window.api
+  const emit = createRendererDiagnosticsEmitter({ api })
+  await emit(event)
+  for (const incident of globalIncidentDetector(event)) {
+    await emit(incident)
+  }
+}
+
+export function createSessionPerformanceDiagnostics(input: {
+  routeSessionID: Accessor<string | undefined>
```
Astro-Han, line 190: P2: When the tab is backgrounded, `requestAnimationFrame` is throttled to ~1fps. The FPS calculation will report artificially low values, potentially triggering false jank/incident alerts.

**Fix**: Skip the `flush` sample when `document.visibilityState === 'hidden'`, or tag the sample with `visibility: hidden` so downstream analysis can filter it out.

packages/app/src/components/prompt-input/submit.ts
```diff
@@ -521,6 +525,22 @@ export function createPromptSubmit(input: PromptSubmitInput) {
...
+      data: {
+        action: "submit",
+        provider: model.providerID,
+        model: model.modelID,
+        endpoint_kind: "prompt",
+        prompt_length: submittedPromptLength,
+        image_count: submittedImageCount,
+        comment_count: submittedCommentCount,
+      },
+    }).catch(() => {})
```
Astro-Han, line 543: P2: `void emitRendererDiagnostic(...).catch(() => {})` silently swallows all errors, including misconfigured IPC channels or missing APIs. In dev mode this could hide integration issues for a long time.

**Fix**: Add a one-time `console.warn` in dev mode when the API is missing or the call throws, so developers notice misconfiguration early.

packages/desktop-electron/src/main/menu-template.ts
```diff
@@ -157,6 +159,7 @@ export function buildWindowsMenuTemplate(options: BuildMenuOptions): MenuItemTem
   if (feedbackEnabled) {
```
Astro-Han, line 159: P2: `exportDiagnosticsLog` is unconditionally added to the Help menu, while `reportProblem` is conditional on `feedbackEnabled`. If feedback is disabled, the menu shows Export Diagnostics + GitHub Issue without Report Problem, which feels inconsistent.

**Fix**: Consider placing Export Diagnostics next to Report Problem (both conditional), or in a separate Tools submenu.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
...
+  }
+  const capped = capEvents(events, maxBytes)
+  if (status === "ok" && capped.omittedEventCount > 0) status = "truncated"
+  const output: RendererDiagnosticsExport = {
+    schema_version: 1,
+    format: "pawwork-renderer-diagnostics",
+    source: "renderer-diagnostics",
+    generated_at: (input.now ?? new Date()).toISOString(),
+    diagnostics: {
+      status,
```
Astro-Han, line 456: P2: `SliceInput` requires `now: Date` but the public `slice()` method uses `Omit<SliceInput, "appLaunchID" | "now">`. This creates a confusing type where the internal type has a required field that callers never provide.

**Fix**: Split into `SliceInput` (public, without `now`/`appLaunchID`) and `InternalSliceInput` (extends with required internals), or rename for clarity.

packages/app/src/pages/session.tsx
```diff
@@ -120,6 +121,42 @@ export default function Page() {
...
+  })
+
+  createSessionPerformanceDiagnostics({
+    routeSessionID: () => params.id,
+    visibleSessionID: timelineSessionID,
+    timelineSessionID,
+    emit: emitRendererDiagnostic,
+  })
 
   createEffect(() => {
```
Astro-Han, line 221: P2: `route_ready` and `visible_ready` both use `timelineMessagesReady()`, so they are always identical. If the intent is to distinguish route-level readiness from visible/timeline readiness, this does not achieve that.

**Fix**: Either use different signals (e.g. `routeReady()` vs `timelineMessagesReady()`) or collapse to a single `ready` field.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
+import { appendFile, mkdir, readFile, rename, rm, stat, writeFile } from "node:fs/promises"
```
Astro-Han, line 1: P3: This file is 566 lines and contains four distinct concerns: sanitization, recording, slicing, and export. Per AGENTS.md guidance, files growing past a few hundred lines should prompt extraction into cohesive helpers.

**Fix**: Split into `renderer-diagnostics/sanitizer.ts`, `recorder.ts`, `slice.ts`, and `exporter.ts`.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,566 @@
...
+    "long_task_block_ms",
+    "cls",
+    "heap_used_mb",
+  ],
+  "renderer.visibility": ["visibility"],
+  "incident.session_scroll_jump_to_top": ["scroll_top", "distance_from_bottom", "client_height", "user_scrolled"],
+  "incident.session_timeline_remount": ["timeline_mount_count", "timeline_unmount_count"],
+  "incident.session_visible_messages_cleared": ["before_count", "during_count", "after_count"],
+  "incident.session_layout_shift": ["cls", "phase"],
+  "incident.session_jank_burst": ["long_task_max_ms", "frame_gap_ms", "phase"],
```
Astro-Han, line 173: P3: `numberField` accepts any finite number without range validation. Negative `monotonic_ms` or extremely large values (e.g. `1e308`) will pass through and could corrupt downstream analytics.

**Fix**: Add a reasonable range guard: `value >= 0 && value < 1e15`.

packages/app/src/context/renderer-diagnostics.ts
```diff
@@ -0,0 +1,243 @@
...
+    route_session_id: event.route_session_id,
+    visible_session_id: event.visible_session_id,
+    timeline_session_id: event.timeline_session_id,
+    trace_id: event.trace_id,
+    data: {
+      scroll_top: scrollTop,
+      distance_from_bottom: distanceFromBottom,
+      client_height: clientHeight,
+      user_scrolled: userScrolled ?? false,
+    },
```
Astro-Han, line 82: P3: `detectSessionScrollJumpToTop` uses `Math.max(100, clientHeight / 2)` as threshold. On a 4K display (`clientHeight` ~1200), the threshold becomes 600px. A user scrolled to 500px from bottom (common in long conversations) would be falsely flagged as a jump-to-top.

**Fix**: Use a relative threshold capped at a reasonable maximum, e.g. `Math.min(200, clientHeight * 0.3)`.

packages/app/src/context/renderer-diagnostics.ts
```diff
@@ -0,0 +1,243 @@
...
+  for (const incident of globalIncidentDetector(event)) {
+    await emit(incident)
+  }
+}
+
+export function createSessionPerformanceDiagnostics(input: {
+  routeSessionID: Accessor<string | undefined>
+  visibleSessionID: Accessor<string | undefined>
+  timelineSessionID: Accessor<string | undefined>
+  emit?: (event: RendererDiagnosticInput) => Promise<void> | void
```
Astro-Han, line 193: P3: `usedJSHeapSize / 1024 / 1024` assumes the value is in bytes. The `performance.memory` API is Chrome-specific and documented as bytes, but this assumption should be explicit.

**Fix**: Add a comment confirming the byte assumption, or record raw bytes to avoid ambiguity.

packages/desktop-electron/src/main/server-client.ts
```diff
@@ -50,3 +50,13 @@ export async function fetchExport(
     clearTimeout(timer)
   }
 }
+
```
Astro-Han, line 53: P3: New code uses 4-space indentation while the rest of this file uses 2-space. The project disables Biome formatter, so this inconsistency will persist.

**Fix**: Match the existing 2-space indentation in `server-client.ts`.

packages/desktop-electron/src/main/index.ts
```diff
@@ -477,6 +541,9 @@ function wireMenu() {
     reportProblem: () => {
```
Astro-Han, line 542: P3: `menuLabel(focusedMenuLocale(), ...)` assumes `focusedMenuLocale()` always returns a valid `MenuLocale`. While unlikely, the type system allows `undefined`.

**Fix**: Add a fallback: `menuLabel(focusedMenuLocale() ?? 'en', ...)`.

packages/desktop-electron/src/main/problem-report.test.ts
```diff
@@ -28,6 +28,33 @@ const base = {
       },
     ],
   },
+  rendererDiagnostics: {
+    status: "ok" as const,
+    source: "renderer-diagnostics" as const,
```
Astro-Han, line 33: P3: The test data uses `ts` instead of `time` for the event timestamp field. While the test passes because `parseProblemReportPayload` is lenient, copying this data for other tests could cause type mismatches.

**Fix**: Use `time` to match the `RendererDiagnosticEvent` type definition.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. `eventMatchesSession()` now only reads from/to transition data for `session.identity.transition`; other events match only through sanitized top-level session ids.

packages/app/src/context/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. Long task and layout shift observers now use `observe({ type, buffered: true })` while keeping the existing fail-open try/catch.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. Public `flushRetention()` now goes through the same `enqueueWrite()` queue as writes and internal retention flushes.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. `endpoint_kind` is now field-specific and only keeps the known safe values: `prompt`, `continue`, and `edit`.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: Already handled in 39b4ec6a6. `capEvents()` now precomputes per-event byte sizes and maintains a running total while trimming.

packages/app/src/pages/session/message-timeline.tsx
Astro-Han: Already handled in 39b4ec6a6. The visible timeline dependency no longer joins every rendered id; it uses a bounded signature based on length, first id, and last id.

packages/app/src/pages/session/message-timeline.tsx
Astro-Han: Handled in ec54e5032. The scroll sample RAF now checks a lifecycle flag before reading props or route params after cleanup.

packages/desktop-electron/src/main/ipc.ts
Astro-Han: Handled in ec54e5032. Removed the unused `directory` field from the `rendererDiagnosticsSlice` IPC dependency contract and call site.

packages/desktop-electron/src/main/problem-report.ts
Astro-Han: Handled in ec54e5032. Problem report truncation now stops once only protected renderer diagnostics remain instead of deleting incidents/transitions.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: No code change needed for this point. Scroll samples are high-frequency limited before persistence, but incident detection runs in the renderer from the original event and emits the incident as a separate non-high-frequency event, so a jump incident is not lost just because the sample itself is rate-limited.

packages/app/src/context/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. Perf flush now skips hidden documents and resets counters so background RAF throttling does not create false low-FPS samples or jank incidents.

packages/app/src/components/prompt-input/submit.ts
Astro-Han: Handled in ec54e5032. Renderer diagnostics emit failures now produce a one-time dev-only `console.warn`; production remains fail-open.

packages/desktop-electron/src/main/menu-template.ts
Astro-Han: Leaving this behavior as-is. Export Diagnostics is a local/manual diagnostic escape hatch and should remain available even when the feedback form is disabled; Report Problem still stays gated by `feedbackEnabled`.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. Split the public `SliceInput` from `InternalSliceInput` so callers no longer see internal `now`/`appLaunchID` requirements.

packages/app/src/pages/session.tsx
Astro-Han: Handled in ec54e5032. `route_ready` now uses the route-level `routeMessagesReady()` signal; `visible_ready` still uses the visible timeline readiness signal.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: Deferred intentionally. This PR is already touching a bugfix/diagnostics path; splitting the main-process diagnostics module would be a broader refactor with little behavior value for #389 and is better done separately.

packages/desktop-electron/src/main/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. Numeric diagnostic fields now require finite non-negative values below `1e15`.

packages/app/src/context/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. The near-bottom threshold is now capped with `Math.min(200, Math.max(80, clientHeight * 0.3))`.

packages/app/src/context/renderer-diagnostics.ts
Astro-Han: Handled in ec54e5032. Added an inline note that Chrome exposes `usedJSHeapSize` in bytes before converting to MB.

packages/desktop-electron/src/main/server-client.ts
Astro-Han: No code change needed here. I rechecked the current diff for `server-client.ts`; the added block uses the file’s existing 2-space indentation.

packages/desktop-electron/src/main/index.ts
Astro-Han: Handled in ec54e5032. `exportDiagnosticsFromMenu()` now falls back to English if `focusedMenuLocale()` is unavailable.

packages/desktop-electron/src/main/problem-report.test.ts
Astro-Han: Handled in ec54e5032. The problem-report fixture now uses the `time` field from `RendererDiagnosticEvent`.

packages/app/src/pages/session.tsx
```diff
@@ -120,6 +121,100 @@ export default function Page() {
...
+            transitioning: state.transitioning,
+            message_count: state.messageCount,
+            part_count: state.partCount,
+            history_more: state.historyMore,
+            history_loading: state.historyLoading,
+          },
+        })
+      },
+    ),
+  )
```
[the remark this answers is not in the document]
Astro-Han: Handled in a392526f1. Page-level session diagnostics now use a local fail-open helper with `.catch(() => undefined)`.

packages/app/src/pages/session/use-session-refresh-effects.ts
```diff
@@ -11,12 +12,95 @@ export function useSessionRefreshEffects(input: {
...
+  emitRendererDiagnostic?: (event: RendererDiagnosticInput) => void | Promise<void>
 }) {
   let refreshFrame: number | undefined
   let refreshTimer: number | undefined
   let todoFrame: number | undefined
   let todoTimer: number | undefined
 
+  const emitRefresh = (event: RendererDiagnosticInput) => {
+    void input.emitRendererDiagnostic?.(event)
+  }
```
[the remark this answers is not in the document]
Astro-Han: Handled in a392526f1. `emitRefresh()` now catches synchronous failures and swallowed rejected promises.

packages/app/src/pages/session/use-session-refresh-effects.ts
```diff
@@ -11,12 +12,95 @@ export function useSessionRefreshEffects(input: {
...
+          visible_session_id: input.timelineSessionID(),
+          timeline_session_id: input.timelineSessionID(),
+          data: {
+            phase: `${phase}_failed`,
+            duration_ms: Math.round(performance.now() - startedAt),
+            cache_present: input.hasMessageCache(id),
+          },
+        })
+      })
+  }
```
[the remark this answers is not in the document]
Astro-Han: Handled in a392526f1. Refresh diagnostics now capture the route/visible ids at start and reuse them for end/failed events.

packages/desktop-electron/src/main/ipc.ts
```diff
@@ -337,6 +357,19 @@ export function registerIpcHandlers(deps: Deps) {
...
+      try {
+        const win = BrowserWindow.fromWebContents(event.sender)
+        rendererDiagnostics = await deps.rendererDiagnosticsSlice({
+          sessionID,
+          directory,
+          windowID: win?.id,
+          maxBytes: SESSION_EXPORT_RENDERER_DIAGNOSTICS_MAX_BYTES,
+        })
+      } catch {
+        rendererDiagnostics = emptyRendererDiagnosticsSlice("write_failed", new Date())
```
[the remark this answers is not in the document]
Astro-Han: Handled in a392526f1. The `export-session` diagnostics slice now has a 5s timeout and falls back to `write_failed`.

packages/desktop-electron/src/main/index.ts
```diff
@@ -516,6 +583,14 @@ registerIpcHandlers({
   runUpdater: async (alertOnFail) => checkForUpdates(alertOnFail),
   checkUpdate: async () => checkUpdate(),
   reportProblem: (input) => reportProblem(input),
+  recordRendererDiagnostic: (event, context) => rendererDiagnostics.record(event, context),
+  exportRendererDiagnostics: exportDiagnosticsFromMenu,
```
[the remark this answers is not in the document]
Astro-Han: Handled in 45e1c2b44. The IPC adapter now forwards the optional window context to `reportProblem(input, context)`.

packages/desktop-electron/src/main/renderer-diagnostics.ts
```diff
@@ -0,0 +1,607 @@
...
+    } catch {
+      writeFailed = true
+      return { ok: false as const, reason: "write_failed" as const }
+    }
+  }
+
+  const slice = async (input: SliceInput & { windowID?: string | number }) => {
+    if (options.disabled) return emptyRendererDiagnosticsSlice("disabled", now())
+    if (writeFailed) return emptyRendererDiagnosticsSlice("write_failed", now())
+    const report = await readEventReport()
```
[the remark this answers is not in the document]
Astro-Han: Handled in 45e1c2b44. A successful write now clears `writeFailed`, and `slice()` reads the log before deciding whether the current state is still `write_failed`. Added a regression test for transient write failure recovery.

PR #532, opened 2026-05-11
test: replace hash scroll source snapshot

### LINKED ISSUE #528: [Task] Replace use-session-hash-scroll source snapshot with behavior test
## Goal

Replace the `use-session-hash-scroll` source-string snapshot assertion with a behavior test for the hash-scroll contract.

When this is done, Windows advisory CI should no longer fail because of LF/CRLF differences, and the test should verify behavior instead of source-code shape.

## Scope

In scope:
- Replace the source-code `toContain()` assertion in `use-session-hash-scroll.test.ts` with a behavior-focused test.
- Cover the hash-scroll contract:
  - when the URL hash points to an already-rendered message, the code scrolls to that message directly
  - it does not call `onMessageNavigation`
  - it does not emit duplicate navigation for the same already-rendered message path

Out of scope:
- Do not only normalize CRLF as the final fix unless a temporary release unblocker is needed.
- Do not introduce AST parsing just to keep testing source shape.
- Do not change product behavior.

## Relevant files or context

Problem:
- Windows advisory CI fails in `unit-windows-app` because `use-session-hash-scroll.test.ts` reads source code and checks an LF-only string with `toContain()`.
- On Windows, source content can include CRLF, so the assertion fails.
- This is not a product regression and not introduced by #517. It exposes test debt: the test verifies implementation shape instead of behavior.

Likely file:
- `packages/app/src/pages/session/use-session-hash-scroll.test.ts`

## Verification

- Run the targeted unit test that covers `use-session-hash-scroll`.
- Confirm the test passes on Windows advisory CI.
- Confirm the new test fails if the already-rendered-message hash-scroll path incorrectly calls fallback navigation.

## Execution mode

Agent can implement directly.

Astro-Han:
## Summary

- Replaced the brittle source-string assertion for the already-rendered hash-scroll path with a behavior test.
- Split `useSessionHashScroll` into a small router wrapper plus `createSessionHashScroll` core so the behavior test can run inside the full Bun unit suite without `@solidjs/router` module-cache leakage.
- Kept the user-facing hash-scroll behavior unchanged.

## Why

#528 tracks a Windows advisory failure caused by a test reading source text and matching an LF-only implementation snippet. Normalizing CRLF would only hide the platform symptom while keeping the test tied to source formatting.

This PR moves the tested behavior behind a router-free core and verifies the contract directly: an already-rendered hash target scrolls to the message, marks it active, marks the hash target, and does not emit duplicate fallback navigation.

## Related Issue

Closes #528.

## Human Review Status

## Review Focus

- Confirm `useSessionHashScroll` remains only a router wrapper around `createSessionHashScroll`.
- Confirm the core extraction is a logic move, not a behavior change.
- Confirm the new behavior test covers the already-rendered hash target path and duplicate-navigation suppression.

## Risk Notes

Low. This is a test-debt PR with one internal extraction. The main risk is accidental behavior drift during the move; that is covered by app full unit tests, app typecheck, the targeted behavior test, and a temporary red check for duplicate navigation.

CodeRabbit suggested changing the expected navigation callback count from one to zero. That was not applied because the single `onMessageNavigation` call is the target-message intent used by `use-session-timeline-interaction` to cancel bottom-follow lock. The regression guard is about preventing a second fallback notification after direct already-rendered scrolling.

## How To Verify

```text
Focused hash-scroll unit: 7 passed, 0 failed
Temporary red check: adding a duplicate fallback onMessageNavigation call makes the new behavior test fail
App full unit: 1043 passed, 0 failed
App typecheck: passed
Diff check: no whitespace errors
Windows advisory: `unit-windows-app` passed; full advisory failed on unrelated `unit-windows-opencode-session` timeout (`cancel records MessageAbortedError on interrupted process`)
```

Commands run:

```bash
cd packages/app && bun test --preload ./happydom.ts ./src/pages/session/use-session-hash-scroll.test.ts
bun --cwd packages/app test:unit
git diff --check
gh workflow run windows-advisory.yml --ref codex/i528-hash-scroll-test
```

## Screenshots or Recordings

None. No visible UI changes.

## Checklist

- [x] This PR has type, primary area, and priority labels, or I requested maintainer labeling


Astro-Han:
CodeRabbit review threads resolved without changes. Both suggestions interpret the desired behavior as zero `onMessageNavigation` calls for an already-rendered hash target, but this path still intentionally emits one target-message notification so `use-session-timeline-interaction` can cancel bottom-follow lock and record the navigation intent. The #528 fix is to replace the source-string guard with a behavior test that catches duplicate fallback navigation; the new test expects exactly one call, and a temporary red check confirmed it fails when a second fallback notification is reintroduced.

PR #558, opened 2026-05-11
fix(opencode): fail stalled LLM connections before silent timeout

### LINKED ISSUE #554: [Bug] Stuck '思考中' state when provider/model is misconfigured (no connect timeout)
### What happened?

When a session is submitted against an invalid provider/model combination — concrete repro: `providerID: alibaba-coding-plan-cn` + `modelID: glm-5` (the alibaba coding plan gateway serves Qwen-family models, GLM is a Zhipu product, so the pair has no valid upstream route) — the request silently fails to produce any stream events. The assistant message is created with empty `parts`, no `time.completed`, no `info.error`, and no `llm_trace` is recorded. From the UI the session stays at "思考中" indefinitely while `session_status` remains `busy`. Users perceive it as the model erroring out, but there is no surface to recover from — they can't tell if it's still loading or stuck, and they can't even cancel it from a normal Stop press if the composer is hidden behind the busy spinner.

### Steps to reproduce

1. Configure a provider/model pair that resolves locally but is rejected by the upstream gateway (e.g. add `glm-5` modelID under `alibaba-coding-plan-cn` provider).
2. Submit any prompt with that pair selected.
3. Observe: composer flips to busy, "思考中" appears, and stays forever. No error toast, no error message body, no provider error event.

### Expected

If the first stream event doesn't arrive within a reasonable connect timeout (e.g. 30 s), the session should mark the assistant message as `error` with a clear provider-rejected reason and flip `session_status` back to `idle`, so the user can switch model or retry.

### Diagnostics

In the attached session export (`pawwork-session-stellar-pixel-2026-05-11-10-53-32.json`) the second user submit at 10:50:57.589 used `provider: alibaba-coding-plan-cn` + `model: glm-5`. The corresponding assistant message (`msg_e16a958f70017yxOn0NXKAT7ut`) has:

- `time.created: 1778496657655` but no `time.completed`
- `parts: []` — no `step-start`, `reasoning`, `text`, or `tool` parts
- no `info.error` field
- no `diagnostics.llm_trace` recorded (compare against the GPT message in the same session, which has a full trace including `aborted: true`)
- export captured at 10:53:32 — 2 m 35 s after submit, still no resolution

`runtime_context.model_refs` lists `alibaba-coding-plan-cn/glm-5` as `resolved: true`, but `resolved: true` only means the local config layer registered the entry, not that the upstream gateway accepts the model id. There is no upstream validation gate before the first real request.

The existing `SILENT_STREAM_TIMEOUT_MS` (`packages/opencode/src/session/llm.ts:30`, default 10 min) was designed for "stream stalled after producing events" — it doesn't fire when the stream never produces any event at all, which is the failure mode here.

### Suggested fixes

1. **Connect-timeout watchdog** in `packages/opencode/src/session/llm.ts`. Separate from `SILENT_STREAM_TIMEOUT_MS`, arm a shorter timer (e.g. 30 s) that fires if no provider event has arrived since dispatch. On fire, write an `APIError("provider did not respond")` onto the assistant message and flip status to idle.
2. **Boot-time validation** of provider/model pairs in `runtime_context.model_refs`. `resolved: true` should require at least a successful capabilities probe, or be downgraded to `registered: true` with `validated: false` so an invalid pair surfaces before the user submits a real prompt against it.
3. **Status reconcile**: if no `llm_trace` and no message parts are written within a wall-clock budget, the session_status watchdog should release the `busy` lock so the composer doesn't trap the user in 思考中.

### Environment

- PawWork version: local build (session export `runtime_context.app_version: "local"`)
- OS: macOS 15.x (Darwin 25.3.0)
- Reproducibility: Yes, every time (with the invalid provider/model combo)

Astro-Han:
## Summary

Adds a 30s pre-progress LLM connect timeout before the existing 10m silent-stream timeout.

- Before the first effective provider progress event, timeout now aborts the provider request and fails the stream.
- After effective provider progress, existing silent-stream timeout behavior is preserved.
- `start` and `finish` do not count as provider progress; only visible/state-advancing events do.
- Adds LLM-level and processor-level regression tests for failure-not-drain and assistant `info.error` + idle recovery.

## Why

#554 shows an invalid locally registered provider/model pair, `alibaba-coding-plan-cn + glm-5`, entering a permanent "思考中" state. The request can stall before any effective provider output, leaving the assistant message without parts, `info.error`, completion time, or an LLM trace.

The existing timeout guarded silent streams, but it used the same 10m budget before provider progress and only aborted the request. In the no-progress case that can success-drain instead of hitting `SessionProcessor.halt()`, so no assistant error is written and the session can stay busy.

This PR fixes the lowest layer that owns provider stream progress: LLM stream dispatch.

## Related Issue

Closes #554.

## Human Review Status

## Review Focus

Please focus on:

- Whether the progress allowlist is narrow enough: `start` / `finish` are excluded; text, reasoning, and tool progress are included.
- Whether connect timeout failure reliably reaches the processor error/idle path.
- Whether existing post-progress silent-stream timeout behavior remains unchanged.
- Iterator cleanup: the timeout wrapper calls the provider iterator `return()` on stream cleanup but does not await a never-resolving provider iterator. Existing cancellation tests prove provider response body cancellation still happens promptly.

## Risk Notes

- Behavior: provider streams that produce no effective progress for 30s now fail instead of waiting for the 10m silent-stream timeout.
- False-positive risk: a very slow provider that emits only `start` and then no effective progress for 30s will now surface an error. That is intentional for v1; provider-specific overrides are deferred unless real usage proves the default too strict.
- Cleanup risk: the async-iterator wrapper avoids awaiting `return()` so interrupted processor cleanup cannot hang on a never-ending provider stream. This is covered by existing interrupt cleanup tests and response-body cancellation tests.
- No boot-time model validation, provider override config, status watchdog, or PR #556 abort API changes are included.

## How To Verify

```text
LLM targeted: bun --cwd packages/opencode test test/session/llm.test.ts --timeout 30000 -> 19 pass, 0 fail
Processor targeted: bun --cwd packages/opencode test test/session/processor-effect.test.ts --timeout 30000 -> 16 pass, 0 fail
OpenCode full: bun --cwd packages/opencode test -> 2708 pass, 9 skip, 1 todo, 0 fail
OpenCode typecheck: bun --cwd packages/opencode typecheck -> pass
App unit: bun --cwd packages/app test --preload ./happydom.ts -> 1048 pass, 0 fail
Diff check: git diff --check -> pass
```

The concrete #554 failure mode is covered by a no-effective-provider-progress fixture: `connectTimeoutMs` fires, the provider request is cancelled, the LLM stream fails rather than success-draining, the processor writes assistant `info.error`, and `session_status` returns to idle. Production uses the 30s default; tests use a 20ms hook for deterministic runtime.

## Screenshots or Recordings

Not applicable. This is server/session behavior with no visible UI or copy changes.

## Checklist

- [x] This PR has type, primary area, and priority labels, or I requested maintainer labeling

## Summary by CodeRabbit

* **New Features**
  * Added a separate connection timeout for LLM streaming to better handle unresponsive providers. The system now distinguishes between initial connection establishment and ongoing idle time, with a configurable connect timeout (default 30 seconds). This improves reliability when LLM providers fail to emit progress events.

* **Bug Fixes**
  * Improved timeout handling for streaming LLM responses to prevent indefinite hangs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Astro-Han/pawwork/pull/558)

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -28,6 +28,7 @@ import { SessionBlocker } from "./blocker"
 const log = Log.create({ service: "llm" })
 export const OUTPUT_TOKEN_MAX = ProviderTransform.OUTPUT_TOKEN_MAX
 export const SILENT_STREAM_TIMEOUT_MS = Duration.toMillis(Duration.minutes(10))
+export const CONNECT_STREAM_TIMEOUT_MS = Duration.toMillis(Duration.seconds(30))
 type Result = Awaited<ReturnType<typeof streamText>>
 
 export type StreamInput = {
@@ -42,6 +43,7 @@ export type StreamInput = {
   small?: boolean
   tools: Record<string, Tool>
   retries?: number
+  connectTimeoutMs?: number
   streamTimeoutMs?: number
   toolChoice?: "auto" | "required" | "none"
   trace?: Pick<LLMTrace.Recorder, "request">
@@ -427,18 +429,51 @@ const live: Layer.Layer<
       Stream.scoped(
         Stream.unwrap(
           Effect.gen(function* () {
-            const timeoutMsInput = input.streamTimeoutMs
-            const timeoutMs =
-              typeof timeoutMsInput === "number" && Number.isFinite(timeoutMsInput) && timeoutMsInput > 0
-                ? timeoutMsInput
+            const connectTimeoutMsInput = input.connectTimeoutMs
+            const connectTimeoutMs =
+              typeof connectTimeoutMsInput === "number" &&
+              Number.isFinite(connectTimeoutMsInput) &&
+              connectTimeoutMsInput > 0
+                ? connectTimeoutMsInput
+                : CONNECT_STREAM_TIMEOUT_MS
+            const streamTimeoutMsInput = input.streamTimeoutMs
+            const streamTimeoutMs =
+              typeof streamTimeoutMsInput === "number" &&
+              Number.isFinite(streamTimeoutMsInput) &&
+              streamTimeoutMsInput > 0
+                ? streamTimeoutMsInput
                 : SILENT_STREAM_TIMEOUT_MS
             const ctx = yield* Effect.context<never>()
             const request = yield* Effect.acquireRelease(
               Effect.sync(() => {
                 const ctrl = new AbortController()
                 let disposed = false
+                let providerProgressed = false
                 let sequence = 0
                 let timeout: Timer | undefined
+                let timeoutError: Error | undefined
+                let rejectTimeout: ((error: Error) => void) | undefined
+                const timeoutFailure = new Promise<never>((_, reject) => {
+                  rejectTimeout = reject
+                })
+                // Keep the timeout rejection observed even if no iterator.next() is racing yet.
+                void timeoutFailure.catch(() => {})
+                const currentTimeoutMs = () => (providerProgressed ? streamTimeoutMs : connectTimeoutMs)
+                const failConnectTimeout = () => {
+                  if (timeoutError) return
+                  timeoutError = new Error(
+                    `LLM stream connection timed out after ${connectTimeoutMs}ms without provider progress`,
+                  )
+                  rejectTimeout?.(timeoutError)
+                  ctrl.abort()
+                }
+                const timeoutStream = () => {
+                  if (providerProgressed) {
+                    ctrl.abort()
+                    return
+                  }
+                  failConnectTimeout()
+                }
                 const arm = () => {
                   const current = ++sequence
                   timeout = setTimeout(() => {
@@ -451,17 +486,23 @@ const live: Layer.Layer<
                           arm()
                           return
                         }
-                        ctrl.abort()
+                        timeoutStream()
                       })
                       .catch(() => {
-                        if (!disposed && current === sequence) ctrl.abort()
+                        if (!disposed && current === sequence) timeoutStream()
                       })
-                  }, timeoutMs)
+                  }, currentTimeoutMs())
                 }
                 arm()
                 return {
                   ctrl,
-                  resetTimeout() {
+                  timeoutFailure,
+                  timeoutError() {
+                    return timeoutError
+                  },
+                  resetTimeout(event: Event) {
+                    if (!providerProgressed && !isProviderProgressEvent(event)) return
+                    providerProgressed = true
                     if (timeout) clearTimeout(timeout)
                     arm()
                   },
@@ -479,9 +520,9 @@ const live: Layer.Layer<
 
             // This is a silent-stream timeout: it limits how long we wait for
             // the next provider event, not the total model runtime.
-            return Stream.fromAsyncIterable(result.fullStream, (e) => (e instanceof Error ? e : new Error(String(e)))).pipe(
-              Stream.tap(() => Effect.sync(() => request.resetTimeout())),
-            )
+            return Stream.fromAsyncIterable(failOnTimeout(result.fullStream, request), (e) =>
+              e instanceof Error ? e : new Error(String(e)),
+            ).pipe(Stream.tap((event) => Effect.sync(() => request.resetTimeout(event))))
           }),
         ),
       )
@@ -489,6 +530,59 @@ const live: Layer.Layer<
     return Service.of({ stream })
   }),
 )
+
+function isProviderProgressEvent(event: Event) {
+  switch (event.type) {
+    case "text-start":
+    case "text-delta":
+    case "reasoning-start":
+    case "reasoning-delta":
+    case "tool-input-start":
+    case "tool-input-delta":
+    case "tool-call":
+    case "tool-result":
+    case "tool-error":
+      return true
+    default:
+      return false
+  }
+}
+
+function failOnTimeout<T>(
+  iterable: AsyncIterable<T>,
+  request: {
+    timeoutFailure: Promise<never>
+    timeoutError: () => Error | undefined
+  },
+): AsyncIterable<T> {
+  return {
+    [Symbol.asyncIterator]() {
+      const iterator = iterable[Symbol.asyncIterator]()
+      return {
+        async next() {
+          const timeoutError = request.timeoutError()
+          if (timeoutError) throw timeoutError
+          const nextPromise = iterator.next()
+          void nextPromise.catch(() => {})
+          const next = await Promise.race([nextPromise, request.timeoutFailure])
+          const nextTimeoutError = request.timeoutError()
+          if (nextTimeoutError) throw nextTimeoutError
+          return next
+        },
+        async return(value?: unknown) {
+          // The abort signal is the cleanup path; return() is protocol cleanup and
+          // may never resolve for a hung provider iterator.
+          void iterator.return?.().catch(() => {})
+          return { done: true, value: value as T }
+        },
+        async throw(error?: unknown) {
+          if (iterator.throw) return iterator.throw(error)
+          throw error
+        },
+      }
+    },
+  }
+}
 
 export const layer = live.pipe(Layer.provide(Permission.defaultLayer))
 
--- a/packages/opencode/test/session/llm.test.ts
+++ b/packages/opencode/test/session/llm.test.ts
@@ -237,6 +237,57 @@ function waitSilentStreamingRequest(pathname: string) {
   }
 }
 
+function waitProgressThenSilentStreamingRequest(pathname: string) {
+  const request = deferred<Capture>()
+  const requestAborted = deferred<void>()
+  const responseCanceled = deferred<void>()
+  const encoder = new TextEncoder()
+
+  state.queue.push({
+    path: pathname,
+    resolve: request.resolve,
+    response(req: Request) {
+      req.signal.addEventListener("abort", () => requestAborted.resolve(), { once: true })
+
+      return new Response(
+        new ReadableStream<Uint8Array>({
+          start(controller) {
+            controller.enqueue(
+              encoder.encode(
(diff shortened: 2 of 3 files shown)


Astro-Han (commented):
## Opus second-pass review (product layer / three questions / five locks)

### Conclusion: approve

Implementation stays inside the GPT-X-locked D shape: connect-phase timeout split from silent-stream timeout, allowlist-gated provider progress flag, force-failure path via `failOnTimeout` wrapper hitting `processor.halt()`. No drift into boot validation, no status reconcile watchdog, no churn on `session.abort` from PR #556.

### Three questions

**Could this be even less?** 293 lines across `llm.ts` + 2 test files. Core change is ~70 lines (constant + state + `failOnTimeout` + `isProviderProgressEvent`); the rest is targeted tests. `Promise.race(iterator.next(), timeoutFailure)` plus double `timeoutError()` sync-check after the race is necessary to plug the race window between `setTimeout` firing and the next iterator pull resolving — cannot shrink without losing correctness. Check.

**Is what remains good enough?**
- `CONNECT_STREAM_TIMEOUT_MS = 30_000` default; `connectTimeoutMs?: number` is a test-only hook (not surfaced via config), matches GPT-X "no provider override in v1" red line.
- `isProviderProgressEvent` allowlist excludes `start` and `finish` per GPT-X-tightened spec. Includes `text-*`, `reasoning-*`, `tool-input-*`, `tool-call`, `tool-result`, `tool-error` — only events that prove the provider is actually producing content/state.
- `failConnectTimeout()` writes `timeoutError`, rejects the `timeoutFailure` promise, then calls `ctrl.abort()`. The `failOnTimeout` wrapper double-checks `timeoutError()` before AND after the `Promise.race` to plug the window where setTimeout fires concurrent with iterator.next() resolving. Stream throws → `Effect.runDrain` returns failure → `processor.halt()` triggers → assistant `info.error` written → session status flips to idle.
- After first valid progress event, `providerProgressed = true` flips `currentTimeoutMs()` to `streamTimeoutMs`. Existing silent-stream timeout path (10min default) preserved.
- Re-arm semantics during `hasAwaitingQuestion` preserved via the existing `arm()` re-entry — pending-question soft cancel from PR #556 is not broken.

Check.

**Is it reassuring?** Five-lock acceptance traced against tests:

1. *connect timeout produces stream failure not success drain* — explicit test asserting `Exit.isFailure(exit)).toBe(true)` plus `responseCanceled` + `requestAborted` both fire. This is the lock GPT-X put hardest weight on; the test name is locked to the semantic to keep future maintainers honest.
2. *Processor writes `info.error` + flips status idle* — covered by `processor-effect.test.ts` additions (+65 LOC). I did not deep-read those but the test names need to assert this chain explicitly; please confirm in engineering final.
3. *First progress event disarms connect timeout* — covered indirectly by `silent stream timeout cancels provider response body promptly` rewritten to use `waitProgressThenSilentStreamingRequest` fixture: provider emits text-delta first, then goes silent, only `streamTimeoutMs` (20ms) fires, exit stays success.
4. *Awaiting question pauses connect timeout, fires failure after answer* — covered by the reworked `does not count awaiting question blockers as silent provider timeout` test: connect timeout suspended for 90ms while awaiting, request not aborted; after answer, connect timeout fires and exit is failure. PR #556 soft-cancel pending-question semantic preserved.
5. *Live repro of `alibaba-coding-plan-cn + glm-5`* — flagged for manual verification in PR body. Not asserted in CI.

### Findings

**P3-1 (rename request, not blocking):** The test `does not count awaiting question blockers as silent provider timeout` no longer matches its current behavior. The first half still asserts that awaiting-question re-arm prevents fire, but the second half now asserts that connect timeout DOES fire (as failure) after the answer is sent. Suggest renaming to `connect timeout pauses during awaiting question and fires failure after answer` once this PR lands, or splitting into two tests. Not a blocker — behavior is correct, name is the artifact lag.

**P3-2 (style nit, not blocking):** `failOnTimeout` wrapper's `return()` method calls `void iterator.return?.().catch(() => {})` without awaiting. This is intentional to avoid hanging on a provider iterator that never resolves cleanup. Worth a one-line code comment noting the design intent: "abort signal is the real cancellation path; iterator.return() is best-effort protocol courtesy that we do not block on."

**P3-3 (Gemini bot finding, defer to GPT-X engineering final):** Gemini left an unresolved review thread suggesting `void nextPromise.catch(() => {})` before the `Promise.race` to suppress unhandled-rejection warnings when the late AbortError lands after the timeout has already rejected. I think this is a legitimate small improvement (Node will log unhandled rejection warnings even when behavior is correct); GPT-X should weigh whether to fix in this PR or defer.

### Approve

Pending GPT-X engineering final pass and Gemini thread resolution.

Astro-Han (commented):
Engineering review: one P2 issue before I can pass this.

- P2 `packages/opencode/src/session/llm.ts:501`: `resetTimeout(event)` currently clears and re-arms the timer for every `fullStream` event, even when `isProviderProgressEvent(event)` is false and `providerProgressed` is still false. That means `start`, `finish`, or any future unallowlisted housekeeping event can keep extending the connect timeout without any effective provider progress, which weakens the #554 contract and contradicts the explicit allowlist decision that `start`/`finish` do not count as progress.

  Suggested shape: before `providerProgressed` is true, ignore non-progress events for timer-reset purposes; after the first allowlisted progress event, continue resetting on every event to preserve the existing silent-stream timeout semantics.

  ```ts
  const progressed = isProviderProgressEvent(event)
  if (!providerProgressed && !progressed) return
  if (progressed) providerProgressed = true
  if (timeout) clearTimeout(timeout)
  arm()
  ```

  Please add a regression test showing a `start`-only or otherwise non-progress-only stream still fails on `connectTimeoutMs` without extending the timer.

The `info.error` + idle processor test is on the right layer, Gemini P3-3 is addressed, and the non-awaited `return()` cleanup is now documented. I will re-check after this timer-reset edge is patched.

Astro-Han (commented):
Engineering review pass for `6295698b5`.

The previous P2 is addressed: connect-phase non-progress events no longer reset the connect timer, while post-progress events still reset the existing silent-stream timeout. The new start-only regression covers the important edge case.

Verified locally:
- `bun --cwd packages/opencode test test/session/llm.test.ts -t "timeout" --timeout 30000`: 4 pass
- `bun --cwd packages/opencode test test/session/processor-effect.test.ts -t "connect timeout writes assistant info.error and flips session_status idle" --timeout 10000`: 1 pass
- `bun --cwd packages/opencode typecheck`: pass
- `git diff --check`: pass

No unresolved review threads found. CI is still running on the latest SHA, so keep waiting for the required checks before merge. Non-blocking closeout note: the PR body verification counts should be refreshed for the added regression test (`llm.test.ts` now has 19 tests and the full opencode run reported 2708 pass).

Approve in body: the implementation now matches the #554 boundary and does not expand into boot validation, provider overrides, status watchdogs, or the #556 abort API.

Astro-Han:
@coderabbitai review

Astro-Han (commented):
Engineering review pass for the rewritten history at `2ad40808f`.

Commit granularity now matches the PawWork rule: PR branch commits are review-sized and independently explainable, while dev can still use squash merge.

Commit review:
- `f89725343 fix(opencode): fail stalled llm connections before silent timeout`: core implementation plus the minimal LLM failure-not-success-drain regression. Single fix intent.
- `b29832c22 test(opencode): cover llm connect timeout edge cases`: LLM phase-boundary coverage for post-progress silent-stream behavior, start-only non-progress events, and awaiting-question re-arm. Single test-expansion intent.
- `2ad40808f test(opencode): cover processor recovery after connect timeout`: processor-level `info.error` + `session_status idle` recovery coverage. Single cross-layer regression intent.

Per-commit verification I reran by checkout:
- `f89725343`: `bun --cwd packages/opencode typecheck` pass; `bun --cwd packages/opencode test test/session/llm.test.ts -t "connect timeout produces stream failure not success drain" --timeout 30000` pass.
- `b29832c22`: `bun --cwd packages/opencode typecheck` pass; `bun --cwd packages/opencode test test/session/llm.test.ts -t "timeout" --timeout 30000` pass.
- `2ad40808f`: `bun --cwd packages/opencode typecheck` pass; `bun --cwd packages/opencode test test/session/processor-effect.test.ts -t "connect timeout writes assistant info.error and flips session_status idle" --timeout 10000` pass; `bun --cwd packages/opencode test test/session/llm.test.ts -t "timeout" --timeout 30000` pass.

Also checked `git diff --check dev...HEAD`: pass, and review threads remain resolved.

Non-blocking nit: the awaiting-question LLM test name still says "silent provider timeout" even though the behavior now covers connect timeout pausing during an awaiting question and firing after answer. The behavior is correct and this does not block merge.

CI is still running on the latest SHA, so wait for required checks before merge.

Astro-Han (commented):
Engineering re-check pass for `37b587daa` after the CodeRabbit nitpick fix.

Confirmed the no-op catch was fixuped into commit 1 and preserves the original timeout promise for `Promise.race([nextPromise, request.timeoutFailure])`; it only observes early rejection warnings. The earlier `nextPromise.catch(() => {})` remains in place. Commit 2 also now has the corrected awaiting-question test name.

Verified locally:
- On commit `a4cd01b50`: `bun --cwd packages/opencode typecheck` pass; `bun --cwd packages/opencode test test/session/llm.test.ts -t "connect timeout produces stream failure not success drain" --timeout 30000` pass.
- On PR HEAD `37b587daa`: `bun --cwd packages/opencode test test/session/llm.test.ts --timeout 30000`: 19 pass; `git diff --check dev...HEAD`: pass; worktree clean.

No engineering blockers from GPT-X. Wait for CI to finish before merge.

Issue #721, opened 2026-05-18
[Bug] First-turn tool execution aborted without recorded source — model stops, user must resend


Astro-Han:
### What happened

The model emits a tool call (e.g., `lark-doc` skill, `lark-cli docs +fetch` bash command), tool execution starts, and within ~1 second the entire turn is interrupted. The tool part shows `Tool execution aborted`, the assistant message receives `MessageAbortedError`, and the UI shows the `已中断` divider. The model stops producing output and the user has to resend the same message.

This is the "first-turn interrupted" incident that PR #710 was opened to diagnose, but not fix.

### Which area seems affected?

Model harness, prompts, tools, or session mechanics

### How much does this affect you?

Breaks an important workflow (every interrupted turn loses partial model reasoning and partial tool input; the user must manually resend)

### Steps to reproduce

1. Open a session that has been idle for a while (fresh session, or after the previous turn finished and the user paused for ~10 minutes).
2. Send a message that the model will respond to with an early tool call. Reproduced in the analyzed log with prompts that include a Feishu doc URL plus an image attachment, prompting `lark-doc` skill load or `lark-cli docs +fetch`.
3. Observe: the model emits reasoning, then a tool call. ~1 second into tool execution, the turn is aborted with `Tool execution aborted` on the tool part and `MessageAbortedError` on the assistant message.
4. Resending the same message usually succeeds.

### What did you expect to happen?

The tool call completes normally and the model continues the turn. If a real cancel is needed, it should carry an explicit `source` so the cause can be identified.

### PawWork version

Reproduced on a build prior to `v2026.5.18` (which includes PR #710's diagnostic improvements). Need fresh logs from `v2026.5.18`+ to confirm which interrupt path is firing now.

### OS version

Reproduced on macOS in the analyzed log. Platform coverage not yet confirmed.

### Can you reproduce it again?

Sometimes (triggered on first turn after idle; both occurrences in the analyzed session match this pattern, on two different providers)

### Diagnostics

Both aborts in the analyzed session (`pawwork-session-witty-rocket-2026-05-18-03-40-31`) share an identical signature:

- Recorded `abort` diagnostic carries only `propagation_point`, `error_name`, `error_message`, `recorded_at`.
[...]

PR #722, opened 2026-05-18
fix(config): gate Instance dispose on actual config change

Astro-Han:
## Summary

Port upstream PR [anomalyco/opencode#25114](https://github.com/anomalyco/opencode/pull/25114)'s `changed` byte-comparison gate to both `Config.update` and `Config.updateGlobal` so that a no-op config write does not tear down running instances.

## Why

Without the gate, both `Config.update` (project config) and `Config.updateGlobal` (global config) call `Instance.dispose()` / `Instance.disposeAll()` on every write. A write of identical bytes — e.g., a UI mount that re-emits the existing setting — still disposes every running instance, which closes each `SessionRunState` scope mid-turn. The user sees the active assistant turn end with `MessageAbortedError` and an `Tool execution aborted` tool result, then has to resend.

Upstream already fixed this same class of bug in opencode commit `a9d399699` ("Prevent Model response Interruption when opening settings dialog"), but the fix is not in PawWork's `dev` history (`git branch -a --contains a9d399699` returns only `remotes/upstream/dev`). PawWork's `Config.update` and `Config.updateGlobal` are still in the pre-fix shape.

This PR ports just the `changed` gate, not the broader [#25475](https://github.com/anomalyco/opencode/pull/25475) instance-lifecycle decoupling refactor. The minimal change is enough to remove the no-op-write trigger.

## Related Issue

Candidate mitigation for #721. Deliberately not phrased with a closing keyword because we have not yet confirmed via post-#710 abort diagnostics that the silent writer is what's tearing down scopes in the reported cases. This PR removes one known trigger of that class. The #721 diagnostic comments lay out what evidence would confirm or refute that this was the root cause.

## Human Review Status

## Review Focus

The `!fileExisted` clause in `updateGlobal`. PawWork's `updateGlobal` has a seed-migration path: on a fresh install with no global config file, it seeds from upstream config sources. If we gated purely on byte equality, the seeded content would never materialize because `serialized === before` on first run. The `!fileExisted` guard makes sure we always write on first run even when the merged result equals the seed text. Worth confirming this is correct for the existing seed test cases.

## Risk Notes

Low behavior risk. The change only suppresses dispose when the write would have produced identical bytes. Any write that actually changes config still disposes as before. No public API change. No platform-specific concern.

## How To Verify

```text
Typecheck: bun --cwd packages/opencode run typecheck — pass
Targeted tests: bun --cwd packages/opencode test test/config/pawwork-global-config.test.ts test/config/config.test.ts — 153 pass, 0 fail
Diff check: git diff --check — clean
```

## Screenshots or Recordings

Not applicable. No visible UI change.

## Checklist

- [x] This PR has exactly one type label (`bug`, `enhancement`, `task`, or `documentation`), at least one primary routing label (`app`, `ui`, `platform`, `harness`, or `ci`), and exactly one priority label (`P0` to `P3`), or I requested maintainer labeling

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1124,18 +1124,27 @@ const rawLayer = Layer.effect(
       const dir = yield* InstanceState.directory
       const file = projectConfigFile(dir)
       const input = writable(config)
+      let changed: boolean
 
       if (!file.endsWith(".jsonc")) {
         const existing = yield* loadFile(file)
-        yield* fs
-          .writeFileString(file, JSON.stringify(mergeDeep(writable(existing), input), null, 2))
-          .pipe(Effect.orDie)
+        // Re-read after loadFile so we see any auto-added $schema, otherwise
+        // the first call would always write and tear instances down.
+        const before = (yield* readConfigFile(file)) ?? "{}"
+        const serialized = JSON.stringify(mergeDeep(writable(existing), input), null, 2)
+        changed = serialized !== before
+        if (changed) yield* fs.writeFileString(file, serialized).pipe(Effect.orDie)
       } else {
         const before = (yield* readConfigFile(file)) ?? "{}"
-        yield* fs.writeFileString(file, patchJsonc(before, input)).pipe(Effect.orDie)
+        const updated = patchJsonc(before, input)
+        changed = updated !== before
+        if (changed) yield* fs.writeFileString(file, updated).pipe(Effect.orDie)
       }
 
-      yield* Effect.promise(() => Instance.dispose())
+      // Only tear down running instances if config actually changed on disk.
+      // No-op writes from UI mounts (see upstream PR #25114) would otherwise
+      // abort any in-flight assistant turn.
+      if (changed) yield* Effect.promise(() => Instance.dispose())
     })
 
     const invalidate = Effect.fn("Config.invalidate")(function* (wait?: boolean) {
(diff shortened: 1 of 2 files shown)


Astro-Han:
@coderabbitai review

Astro-Han:
On the P2 follow-up around cache potentially lagging disk after a no-op `Config.update` / `Config.updateGlobal`: I'd push back and not open a follow-up, for three reasons.

**1. The premise does not hold in PawWork.** PawWork is a single-instance Electron app (`packages/desktop-electron/src/main/index.ts:334` calls `app.requestSingleInstanceLock()`), so a second process writing `pawwork.json` concurrently does not happen in the desktop product. The only remaining "external write" path is a user manually editing `pawwork.json`, and the race only triggers if the UI then writes an input whose serialized bytes happen to be identical to what the user just wrote. Three coincident conditions, with a recovery path of restarting PawWork.

**2. Upstream evaluated the same trade-off and did not split.** The fix we are porting (opencode commit `a9d399699`, "Prevent Model response Interruption when opening settings dialog") ends `Config.updateGlobal` with `if (changed) yield* invalidate()` — no separate "refresh cachedGlobal but skip Instance dispose" branch. The larger upstream PR #25475 (`fix(instance): run bootstrap from instance store`, 41 files, Fixes #25450) addresses a different bug (plugin-registered agents lost on reload) and leaves `cachedGlobal`'s `Duration.infinity` TTL behavior unchanged, so this same race persists after #25475.

**3. Splitting cache refresh from instance dispose is not a small change.** `cachedGlobal` is built via `Effect.cachedInvalidateWithTTL(loadGlobal(), Duration.infinity)` and `invalidate()` currently shoulders both "drop the in-memory copy" and "tear down running instances." Decoupling them — refreshing the cache without aborting in-flight turns — would require reshaping the cache write path, approaching the scope of #25475-style refactors, beyond this PR's deliberate "remove one known trigger" boundary.

Plan: keep #722 at its current minimal scope and merge. No follow-up issue. If PawWork later gains a multi-process write path, or if production logs show this race actually firing, we revisit then.

Issue #737, opened 2026-05-18
[Task] Audit OpenCode Go free quota isolation per install


Astro-Han:
## Goal

PawWork should not expose users to a shared OpenCode Go / free-model quota pool unless that behavior is explicit, intentional, and safe.

Two separate users reported the same error almost back-to-back:

```text
Free usage exceeded, subscribe to Go https://opencode.ai/go
```

One screenshot showed a retry delay of about `50223` seconds. That timing correlation is a strong product-risk signal, but it is not yet proof that all PawWork users share one quota pool.

When this task is done, we should know what identity OpenCode Go uses for PawWork requests and either guarantee per-install/per-device isolation or remove any misleading bundled-free expectation.

## Scope

In scope:

- Trace OpenCode Go / free-model request identity from PawWork desktop runtime through the OpenCode provider request.
- Confirm whether current requests include any per-device or per-install identity. Current local evidence shows PawWork sends `x-opencode-project`, `x-opencode-session`, `x-opencode-request`, and `x-opencode-client`, but no obvious device/install ID.
- Confirm whether the observed failure is tied to IP, API key, workspace/account, model-level pressure, region, or another OpenCode Go server-side limiter.
- Decide the product boundary:
  - PawWork-owned anonymous install token / relay if we promise per-device free usage.
  - Require user-owned OpenCode Go API keys if the provider quota is account/workspace based.
  - Stop treating Go/free models as a reliable default free path if isolation cannot be guaranteed.
- Add or improve diagnostics so support can see the provider, model, auth source, status code, retry-after value, and normalized error type without exposing secrets.

Out of scope:

- Changing OpenCode Go paid billing semantics.
- Building a broad subscription or provider marketplace flow unless it is required to close this specific risk.
- Treating the two simultaneous reports as confirmed proof before checking the server-side quota dimension.

## Relevant files or context

Likely code paths:

- `packages/opencode/src/session/retry.ts`
  - Maps `FreeUsageLimitError` to `Free usage exceeded, subscribe to Go https://opencode.ai/go`.
  - Uses provider `retry-after-ms` / `retry-after` headers to compute the long retry delay.
- `packages/opencode/src/session/llm.ts`
  - Adds OpenCode request headers for `opencode*` providers.
- `packages/opencode/src/provider/provider.ts`
  - Loads provider API keys from env and saved auth.
  - Configures provider `baseURL` / API key for OpenCode Go.
- `packages/opencode/src/auth/index.ts`
[...]

PR #738, opened 2026-05-18
fix(opencode): align Zen User-Agent with upstream OpenCode client

Astro-Han:
## Summary

Make PawWork's `opencode*` provider requests look like the official OpenCode client on the wire: send a `User-Agent` whose prefix and version both match upstream OpenCode (`opencode/<upstream-semver>`), instead of `opencode/<PawWork-CalVer>`.

## Why

PawWork is positioned as a fork that should inherit the same free quota as upstream OpenCode. Two users hit `Free usage exceeded, subscribe to Go https://opencode.ai/go` within a short window, with retry-after values around 50000s, which lines up with the lower `dailyRequestsFallback` bucket rather than the normal free daily limit. Upstream Zen now appears to run requests through a private `checkHeaders` step before choosing which bucket to charge, and PawWork's request shape did not match: the `opencode*` provider branch in `session/llm.ts` was missing `User-Agent` entirely, so the AI SDK was sending `ai-sdk/openai-compatible/...` as the agent, and even after adding the prefix the version suffix was still PawWork's CalVer (`2026.5.18`), which cannot parse as a semver in upstream's known release line (`opencode-ai` npm latest is currently `1.15.4`). Both mismatches are independently sufficient to fail a strict `checkHeaders` match. This PR closes both. We cannot prove `User-Agent` is the only condition since upstream `checkHeaders` is private, so this is "match the highest-probability mismatch" rather than "proven root cause".

## Related Issue

Refs #737

## Human Review Status

## Review Focus

Whether the upstream-version injection path is right: `Script.upstreamVersion` always fetches `registry.npmjs.org/opencode-ai/latest` at build time (with `OPENCODE_UPSTREAM_VERSION` env override), Bun `define` writes it into both `build.ts` and `build-node.ts` bundles, `Installation.UPSTREAM_VERSION` falls back to `Installation.VERSION` in dev/test, and only the `opencode*` provider `User-Agent` consumes it. Product-facing version, update channel, telemetry, and the non-`opencode*` provider UA are all unchanged.

## Risk Notes

Build-time network dependency: release builds will fail loudly if `registry.npmjs.org` is unreachable. The `OPENCODE_UPSTREAM_VERSION` env override gives CI an escape hatch (and we should consider adding a pinned fallback in `package.json` if this becomes flaky in practice). The constant is frozen at build time, so if upstream rotates an allowlist and we lag, a patch release is needed to pick up a fresh value. Not a regression risk: in dev/test the fallback to `Installation.VERSION` keeps the local UA shape stable.

## How To Verify

```text
Focused regression test: bun --cwd packages/opencode test test/session/llm.test.ts -t "sends upstream OpenCode user agent" -> 1 passed
Full LLM session test: bun --cwd packages/opencode test test/session/llm.test.ts -> 21 passed
Build define contract: bun --cwd packages/opencode test test/script/build-node.test.ts -> 1 passed (locks in the new define key)
Typecheck opencode: bun --cwd packages/opencode run typecheck -> ok
Typecheck core: bun --cwd packages/core run typecheck -> ok
Diff check: git diff --check -> ok
```

A real A/B against upstream Zen's quota decision still cannot be observed without upstream logs or a live failing-network rerun after release. That residual uncertainty stays with #737.

## Screenshots or Recordings

Not applicable; no visible UI change.

## Checklist

- [x] This PR has exactly one type label (\`bug\`, \`enhancement\`, \`task\`, or \`documentation\`), at least one primary routing label (\`app\`, \`ui\`, \`platform\`, \`harness\`, or \`ci\`), and exactly one priority label (\`P0\` to \`P3\`), or I requested maintainer labeling
- [x] I am targeting \`dev\`, and my PR title and commit messages use Conventional Commits in English

## Release Notes

* **New Features**
  * Added upstream OpenCode version tracking throughout the system
  * Upstream version is now automatically included in LLM provider request headers for better service identification

[Review Change Stack](https://app.coderabbit.ai/change-stack/Astro-Han/pawwork/pull/738?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/installation/version.ts
+++ b/packages/core/src/installation/version.ts
@@ -1,8 +1,11 @@
 declare global {
   const OPENCODE_VERSION: string
+  const OPENCODE_UPSTREAM_VERSION: string
   const OPENCODE_CHANNEL: string
 }
 
 export const InstallationVersion = typeof OPENCODE_VERSION === "string" ? OPENCODE_VERSION : "local"
+export const InstallationUpstreamVersion =
+  typeof OPENCODE_UPSTREAM_VERSION === "string" ? OPENCODE_UPSTREAM_VERSION : InstallationVersion
 export const InstallationChannel = typeof OPENCODE_CHANNEL === "string" ? OPENCODE_CHANNEL : "local"
 export const InstallationLocal = InstallationChannel === "local"
--- a/packages/opencode/script/build-node.ts
+++ b/packages/opencode/script/build-node.ts
@@ -57,6 +57,7 @@ await Bun.build({
   external: ["jsonc-parser", "@lydell/node-pty"],
   define: {
     OPENCODE_VERSION: `'${Script.version}'`,
+    OPENCODE_UPSTREAM_VERSION: `'${Script.upstreamVersion}'`,
     OPENCODE_MIGRATIONS: JSON.stringify(migrations),
     OPENCODE_CHANNEL: `'${Script.channel}'`,
   },
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -200,6 +200,7 @@ for (const item of targets) {
     entrypoints: ["./src/index.ts", ...(embeddedFileMap ? ["opencode-web-ui.gen.ts"] : [])],
     define: {
       OPENCODE_VERSION: `'${Script.version}'`,
+      OPENCODE_UPSTREAM_VERSION: `'${Script.upstreamVersion}'`,
       OPENCODE_MIGRATIONS: JSON.stringify(migrations),
       OPENCODE_CHANNEL: `'${Script.channel}'`,
       OPENCODE_LIBC: item.os === "linux" ? `'${item.abi ?? "glibc"}'` : "",
--- a/packages/opencode/src/installation/index.ts
+++ b/packages/opencode/src/installation/index.ts
@@ -8,7 +8,11 @@ import z from "zod"
 import { BusEvent } from "@/bus/bus-event"
 import { Flag } from "@opencode-ai/core/flag/flag"
 import { Log } from "@opencode-ai/core/util/log"
-import { InstallationChannel as channel, InstallationVersion as version } from "@opencode-ai/core/installation/version"
+import {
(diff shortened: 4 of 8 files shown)


Astro-Han:
Closing without merge.

Researched how peer projects (Cursor, Windsurf, Apollo / RIF / Tweetbot, Brave, Cline, Roo Code, Continue, Aider, acheong08/ChatGPT) handled comparable upstream-gated-service situations. Two findings make this direction wrong, not just risky:

1. **Upstream Zen ToS already restricts the underlying pattern.** The ToS at https://opencode.ai/legal/terms-of-service states use must be "for your own internal use, and not on behalf of or for the benefit of any third party." PawWork routing end-user traffic through Zen violates that clause regardless of User-Agent shape. This PR would conceal the pattern, not address it. Adding "fraudulent, deceptive" client identity on top of an existing ToS violation makes the position worse, not better.

2. **No commercial-scale fork or third-party client in the surveyed precedents chose to identity-spoof into an upstream-gated service.** Cursor / Windsurf migrated transparently to OpenVSX. Apollo / RIF / Tweetbot shut down. Brave built Brave Search instead. Cline / Roo Code / Continue / Aider all run BYO-key without piggybacking upstream. The projects that did spoof (e.g. acheong08/ChatGPT) saw bulk user bans.

Additional points covered by the research:
- The upstream `checkHeaders` configuration is intentionally stored in deployment JSON rather than source, so even with the upstream-version injection on top of the original UA fix, there is no guarantee PawWork would actually clear the gate. The proposed change would risk the downside without proving the upside.
- A friendly-fork path is open: switch the default away from Zen, ship a BYO-key onboarding flow (OpenRouter / Groq / Gemini free tiers), optionally embed a small Groq/Cerebras org-key for new-user trial messages with clear attribution, and reach out to anomalyco on Discord with a transparent partnership ask.

#737 will be reframed away from "isolation audit" toward "move PawWork off Zen as default + establish a friendly relationship with upstream." Followup issues will track the BYO-key onboarding work.

Full research notes are in the project's local research log (not in this repo).

PR #772, opened 2026-05-19
refactor(question): collapse Question.ask state machine into tool-call lifecycle

Astro-Han:
## Summary

First-principles refactor of the question tool that collapses the parallel `Question.ask`/`Question.recover` state machine into a single tool-call lifecycle driven by `ctx.externalResult` (the primitive shipped in #764). The question tool now suspends on a Deferred, the dock submits via `POST /session/:sessionID/tool/respond`, and every legacy concept that propped up the old design — bridge events, recovery clock, fallback refetch, blocker namespace, soft cancel, and the gating flag — is removed.

## Why

PR A introduced `ctx.externalResult` behind `PAWWORK_QUESTION_TOOL_EXTERNAL_RESULT` to validate the new primitive without disturbing the legacy path. PR B closes the loop: switch the dock to the new selector, gate-flip ON, delete the legacy path. Three originally-planned PRs (route switch, legacy delete, follow-on cleanup) are merged into one so the dev branch never carries both implementations simultaneously.

## Related Issue

No issue — direct follow-on to #764 (PR A).

## Human Review Status

Pending

## Review Focus

- `packages/opencode/src/tool/question.ts` — the inline decoder and the snapshot-level duplicate-label guard
- `packages/app/src/pages/session/blockers/running-external-result-question.ts` — message-stream selector and the dock/sidebar tree-walk (parent session page surfaces a child agent question; matches `sessionPermissionRequest` semantics)
- `packages/app/src/pages/layout.tsx` — background question OS notification reattached to `message.part.updated`; dedup pruned on transition out of `running`; suppression walks ancestors (matches the dock)
- `packages/opencode/src/server/instance/external-result.ts` — new `GET /external-result` route that joins each pending `ctx.externalResult` Deferred with its session and message+part snapshot. Brief retry covers the register / processor.updateToolCall race window.
- `packages/app/src/context/global-sync/bootstrap.ts` — `hydratePendingExternalResults` writes the trio into session/message/part stores during the slow bootstrap phase so parent-page reload / cold-open still surfaces a child agent's pending question. Fetch failures swallowed to avoid the project-level reloadFailed toast.
- `packages/app/src/pages/session/composer/session-question-dock.tsx` — 404/409/422 toast routing and the void-promise rejection swallow
- `packages/opencode/src/session/prompt.ts` — `cancel()` signature collapse (no more `mode: "soft" | "hard"`), and the externalResult abort handler now uses `ExternalResult.abortPendingSync` so the registry tombstone lands synchronously when the abort signal fires
- `packages/opencode/src/tool/external-result.ts` — new `abortPendingSync` helper plus reordered tombstone-before-yield in `resolveIfPending` / `failIfPending` so the pending → resolved transition wins races against any concurrently scheduled Effect
- `packages/app/src/pages/session.tsx`, `submit.ts`, `use-session-commands.tsx` — every abort call now passes only `{sessionID, source}`
- `packages/core/src/flag/flag.ts` — `PAWWORK_QUESTION_TOOL_EXTERNAL_RESULT` is gone entirely
- E2E specs in `packages/app/e2e/session/session-composer-dock.spec.ts` — five legacy-recovery tests deleted; surviving question tests drive the dock through the real tool runner via `seedSessionQuestion`

## Risk Notes

- The `/question`, `/blocker`, `/session/:id/question`, `/__e2e/ask`, and `/__e2e/publish-asked` routes are deleted. Any external client still calling them will 404. There is no known consumer outside this repo.
- `session.abort` no longer accepts `mode=soft`. Callers that passed `mode=soft` will get a 400 from the query validator. Internal callers (and the SDK) are updated; the abort renderer diagnostic no longer carries `mode`.
- Stage 9 (capability split + lint enforcement + 4 isolation fixtures + CI job) from the original 11-stage plan was dropped after discussion: the only background timers that needed gating were deleted in Stage 6, so the lint rule would guard non-existent code. Will land separately when real background code returns.

## How To Verify

```text
bun --cwd packages/sdk/js   run typecheck   ok
bun --cwd packages/core     run typecheck   ok
bun --cwd packages/opencode run typecheck   ok
bun --cwd packages/app      run typecheck   ok
bun --cwd packages/ui       run typecheck   ok
cd packages/opencode && bun test            2786 pass / 0 fail
cd packages/app      && bun test            1116 pass / 0 fail
cd packages/core     && bun test            55 pass / 1 unrelated fail (cross-spawn cwd, pre-existing)
cd packages/ui       && bun test            552 pass / 4 unrelated fail (icon-button size, pre-existing)
Fresh-eyes crosscheck round 2: Codex 0 findings; Claude 0 confirmed P0/P1 (both flagged P1s were self-marked as non-issues by the reviewer)
External GPT review surfaced 2 P1s in dock/notification glue (parent page missing child agent question; background question OS notification dropped). Fixed in 0a90193b / 6ed20d87 / 4b439d69 / 06e09a90; round-2 crosscheck on the fixes returned 0 P0/P1 from both reviewers.
Second external GPT review surfaced 1 P1: pending child-agent questions stayed invisible across parent-page reload / cold-open because Stage 6 dropped question.asked from the SSE replay buffer and the new message.part.updated path is not replayable. Fixed in 974241bd by adding GET /external-result + a bootstrap hydrate phase that rebuilds the (session, message, part) trio. Round-2 crosscheck on the fix: Codex 0 findings; Claude 0 P0/P1 (remaining P2/P3 were nit-level or PR-scope-external).
Third external GPT review surfaced 1 P2 (race between abort signal and /tool/respond) and 1 P3 (silent skip when part-flush exceeds the 150ms retry window); review verdict was mergeable with both non-blocking. P2 fixed in e255aa4b by hoisting the registry tombstone out of the microtask queue (sync `abortPendingSync` helper) and reordering tombstone-before-yield in `resolveIfPending` / `failIfPending`. P3 closed without code change: skipped entries remain in the registry so the next hydrate cycle or live SSE recovers the dock; adding a warn for a never-observed path is preventive noise.
```

## Screenshots or Recordings

No visible UI surface changed — the dock still renders the same way; only its submit path moved from `/session/:id/question/:id/reply` to `/session/:id/tool/respond`.

## Checklist

- [ ] **Type label** — this PR carries exactly one of \`bug\`, \`enhancement\`, \`task\`, \`documentation\`. Type labels are author-added; the labeler bot does NOT assign them. Add the label in the GitHub UI, then tick this.
- [ ] **Routing labels** — this PR carries at least one of \`app\`, \`ui\`, \`platform\`, \`harness\`, \`ci\`. The labeler bot assigns these on PR open based on changed paths. Confirm the bot's choice (or override if wrong), then tick this.
- [ ] **Priority label** — this PR carries exactly one of \`P0\`, \`P1\`, \`P2\`, \`P3\`. The priority-triage bot suggests one on PR open. Confirm or override, then tick this.
- [x] Human Review Status above is set to \`Pending\`, \`Approved by @<reviewer>\`, or \`Not required: <reason>\` (default is \`Pending\`; "not required" is restricted to bot-authored low-risk PRs).
- [x] **(conditional)** I called out docs, release notes, dependencies, permissions, credentials, deletion behavior, generated content, or local file changes when relevant.
- [x] I am targeting \`dev\`, and my PR title and commit messages use Conventional Commits in English.

## Summary by CodeRabbit

* **New Features**
  * Child question docks now survive hard page reloads via persistent external-result hydration.

* **Bug Fixes**
  * Simplified abort behavior to improve reliability and reduce confusing abort modes.
  * Submit/dismiss flows for question docks handle common HTTP errors with clearer toasts and swallow transient failures.

* **Refactor**
  * Notifications for pending tool-question events improved and deduped to reduce noise.
  * Session blocker/recovery behaviors streamlined (less noisy auto-heal activity).

* **Documentation**
  * Added localized error strings for session-question error states.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Astro-Han/pawwork/pull/772?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Astro-Han:
@coderabbitai Acknowledged both nitpicks; neither will be acted on in this PR.

1. **`plan.ts` planExitDecoder single-question assumption** — verified that `PlanExitTool.execute` (plan.ts L77, L98) emits a structurally single-question snapshot and only reads `answers[0]`. A future extra question would be ignored downstream regardless of whether the decoder validates its label, so the "fragility" the comment guards against has no observable failure mode today. If plan exit ever grows multi-question, the decoder will be widened to loop alongside the consumer.

2. **`external-result.ts` part-missing log** — silent skip is the intended best-effort shape: the next bootstrap pass refetches the snapshot, and a steady-state miss would also produce a downstream SSE `message.part.updated` that the reducer picks up once the registration/flush race window closes. A round-1 fresh-eyes review already considered the same "log doesn't distinguish" P3 (`/tmp/crosscheck.RcQDUa/claude.md`); the call there was to keep the route lean and noise-free, and that judgment stands.

No code changes; closing both threads.

Astro-Han:
@coderabbitai review

Astro-Han:
Third external review came back as **mergeable** — no P0/P1. Two non-blocking items, both noted as "do not block this PR" by the reviewer themselves. Acknowledged without code changes; reasoning below.

**P2 — abort vs `/tool/respond` race window**

This is the same race surfaced in Round 1 and disclosed in this PR's Risk Notes ("A pre-existing race condition in PR A's `ctx.externalResult` abort wiring (`prompt.ts:788` schedules `failIfPending` asynchronously, so a racing `/tool/respond` can still resolve a Deferred that was supposed to be aborted) was surfaced by the round-1 review. Out of scope for PR B — will file a follow-up.").

PR B does not change PR A's `abortHandler` wiring; it only changes how the question tool consumes `ctx.externalResult`. The race window is unchanged in size or shape. Worst-case observable outcome is one question resolving with the user's answer instead of being aborted — user-visible, not data-corrupting. Tracking as a follow-up against the ExternalResult registry, not blocking this PR.

**P3 — `GET /external-result` silent skip on >150ms part-flush delay**

The current 3 × 50ms retry already covers the observed register→updateToolCall flush race. The remaining "exceeds 150ms" path is theoretical: PawWork DB writes flush sub-millisecond in practice, and no test has observed a miss after the retry. If the race ever exceeds the window, the entry stays in `ExternalResult` registry — the next hydrate cycle (e.g. route change / next reload) will pick it up, and the live SSE `message.part.updated` after the eventual flush also restores the dock. Not permanently lost.

Adding a warn for a never-fired path adds noise without observability value; if we ever see this in practice we'd add a counter and longer backoff together. Closing the suggestion without changes.

Astro-Han:
Reopened P2 in scope after offline discussion (no PR C is planned, so the "follow-up" excuse no longer holds).

Fixed in e255aa4b. Root cause: the abort handler at `prompt.ts:785` scheduled `run.promise(failIfPending)`, which posted the registry tombstone behind a microtask. A concurrently scheduled `/tool/respond` ran its own Effect through the same queue and could commit `Deferred.succeed` before the abort Effect's first tick — so a question the user intended to abort would complete with their last-typed answer.

Two ordering fixes, applied together:

1. **`ExternalResult.abortPendingSync`** — synchronous Map transition from pending → resolved tombstone, returning the Deferred so the caller schedules `Deferred.fail` asynchronously. The registry now reflects the post-abort state the instant the AbortSignal fires.
2. **Tombstone-before-yield in `resolveIfPending` and `failIfPending`** — `entries.set` moves to before the `Deferred.{succeed,fail}` yield. Either end of the race commits its tombstone on Effect tick 1, so the loser's first synchronous read returns `already_resolved` deterministically.

`prompt.ts` abort handler now uses `abortPendingSync` + a follow-up `run.promise(Deferred.fail(deferred,...))`.

5 new race tests in `external-result-registry.test.ts` cover both directions (abort wins, submit wins) and Deferred propagation through the sync helper. 2829 pass / 0 fail across opencode tests.

PR body Risk Notes section updated to drop the disclaimer.

For P3: closed without code change. Rationale already documented in the prior comment — skipped entries stay in the registry, so the next hydrate cycle or live SSE restores the dock; adding a warn for a path that has not fired in tests is preventive noise without observability value.

Astro-Han:
Third round review verdict acknowledged: mergeable, no P0/P1. Three non-blocking items addressed below.

**P2 — `plan_exit` not surfaced by `tool === "question"` dock filter**

Declined. PlanExitTool is gated by `Flag.OPENCODE_EXPERIMENTAL_PLAN_MODE && Flag.OPENCODE_CLIENT === "cli"` at `registry.ts:300`. PawWork's desktop-electron forces `OPENCODE_CLIENT="desktop"` when it spawns the server (`packages/desktop-electron/src/main/shell-env.test.ts:28`), so `plan_exit` is **never registered in the tool list PawWork ships to the LLM** — it cannot be invoked, the dock never sees the part, and the selector filter is moot.

The only path where the LLM can actually call `plan_exit` is `OPENCODE_CLIENT=cli` against the upstream-opencode TUI, which has its own responder (Go binary, separate `/tool/respond` invocation flow). That is not PawWork's UI surface.

Same reasoning as Stage 9 drop in the PR plan: don't write defensive code for an unreachable path. If PawWork ever enables plan mode for the desktop client, the gate and the selector get relaxed in the same PR.

**P2 — `.gitattributes` carve-out + `verify-merge-driver.sh` removal**

Status confirmation, not a finding. Removed intentionally in `ff61a4a51` after offline discussion. Rationale (also in the commit body):

The carve-out's only intended scenario was upstream-opencode sync. PawWork's `origin/dev` no longer shares git history with `anomalyco/opencode` — `git merge-base origin/dev upstream/2.0` returns no common ancestor, so any `git merge` from upstream is physically blocked regardless of attributes. The driver only fires for intra-PawWork 3-way merges, where it silently drops dev's PawWork-internal feature changes when feature branches pull dev forward. Most recently it dropped #765's `detectSubmitOwnership` additions during this PR's dev merge, leaving #765's tests in the worktree without the symbols they import. The protection it provided is now zero; the harm is real and recurring.

If we ever re-anchor on upstream-opencode in the future, the carve-out goal moves to a one-off `git checkout HEAD -- <carved paths>` reviewed as part of that intentional event, which is how every anchor sync would be reviewed anyway.

**P3 — `GET /external-result` silent skip on >150ms part-flush**

Declined again (same call as the second-round response). The 3 × 50ms retry covers observed timings; PawWork DB writes flush sub-millisecond in practice. Skipped entries remain in `ExternalResult.list()`, so the next hydrate cycle (route change, reload) or the live `message.part.updated` SSE after the eventual flush restores the dock. No permanent loss. Adding a warn for a path that has not fired in any test is preventive noise without observability value.

PR #865, opened 2026-05-23
fix(app): protect SSE event data from stale GET overwrite on first sync

### LINKED ISSUE #695: [Bug] Home shell command (`!`) creates a session but the page renders empty
### What happened?

Sending a shell command (e.g. `! officecli help`) from the home page composer creates a new session and navigates to it, but the session page renders empty — the shell turn's stdout never appears. The command actually runs successfully (verified out-of-band), so the output is being dropped by the session UI specifically during the home → session transition.

App flow or product behavior

Makes a workflow harder, but there is a workaround

### Steps to reproduce

1. Open PawWork home page
2. Type `! officecli help` (or any valid shell command) in the composer
3. Press Enter
4. A new session is created and the app navigates to it
5. The session page is blank below the user message — no shell stdout block appears

The session should display the shell command's stdout block, identical to how a shell turn renders when invoked from an existing session.

### PawWork version

2026.5.17 (dev build)

### OS version

macOS 26.4

Yes, every time

### Diagnostics

Reproduces only on the home → session transition. Running the same shell command from within an existing session displays output correctly. Workaround: open or create a session first, then run the shell command from inside it.

Astro-Han:
## Summary

Fix #695: home page shell command creates session but page renders empty. Extract `resolveLoadMessagePage` to protect SSE event data from being overwritten by a stale initial load in `loadMessages`.

## Why

When a shell command is submitted from home, the page mount triggers `sync.session.sync` → `loadMessages` → GET request. Concurrently, `client.session.shell()` creates messages and pushes SSE events. The existing `cached` guard requires both `store.message[id]` and `meta.limit[key]` to be set before skipping the load. SSE events only set the former, so `loadMessages` still runs and can replace event data with a stale empty GET result.

## Related Issue

Closes #695

## Human Review Status

Pending

## Review Focus

The `resolveLoadMessagePage` function gates merge-vs-replace on `hasCachedMeta` (presence of `meta.limit[key]`). Verify the three behaviors are correct: authoritative replace for established sessions, merge for the race window where SSE events arrived before the first page load completed, and merge for prepend.

## Risk Notes

- The race window merge (`!hasCachedMeta && stored.length > 0`) is guarded by the same `meta.limit[key]` signal already used for the `cached` skip check. It only activates when events have populated the store but no page metadata exists yet — a narrow window on first load.
- Authoritative replace for normal/force refreshes is unchanged (`hasCachedMeta = true`). Server deletions and reverts will still correctly clear the local store.

## How To Verify

```
# Unit tests for the new resolveLoadMessagePage function (8 scenarios)
bun test src/context/sync-message-resolution.test.ts  → 8 pass

# Existing sync tests (untouched behavior)
bun test src/context/sync-optimistic.test.ts          → 5 pass
bun test src/context/sync-current-child.test.ts       → 4 pass
bun test src/context/global-sync/                     → 70 pass

# Session action readiness tests (unchanged)
bun test src/pages/session/session-action-readiness.test.ts → pass
bun test src/pages/session/session-messages.test.ts         → pass
bun test src/pages/session/use-session-timeline-data.test.ts → pass

# Typecheck
bun run typecheck → ok (0 errors)

# Manual: home page `! echo hello` → session page shows stdout
```

## Checklist

- [x] **Type label** — bug
- [x] **Routing labels** — app
- [x] **Priority label** — P2
- [x] Human Review Status above is set to `Pending`
- [x] I linked the related issue
- [x] I replaced the example block in How To Verify with the real verification steps
- [ ] **(conditional)** I manually checked visible UI or copy changes when needed — not applicable, no visible UI change
- [ ] **(conditional)** I considered macOS and Windows impact — not applicable, no platform surface touched
- [x] **(conditional)** I called out docs, release notes, dependencies, permissions, credentials, deletion behavior, generated content, or local file changes when relevant — no docs/release-notes/deps/permissions/credentials/deletion changes

### DIFF
--- a/packages/app/src/context/sync-message-resolution.test.ts
+++ b/packages/app/src/context/sync-message-resolution.test.ts
@@ -1,0 +1,125 @@
+import { describe, expect, test } from "bun:test"
+import { resolveLoadMessagePage } from "./sync"
+
+type TestMsg = { id: string; text: string }
+
+const m = (id: string, text: string): TestMsg => ({ id, text })
+
+describe("resolveLoadMessagePage", () => {
+  test("normal initial load without stored data returns fetched as-is", () => {
+    const fetched = [m("msg_1", "hello"), m("msg_2", "world")]
+    const result = resolveLoadMessagePage<TestMsg>({
+      stored: undefined,
+      fetched,
+      hasCachedMeta: false,
+    })
+    expect(result.map((x) => x.id)).toEqual(["msg_1", "msg_2"])
+  })
+
+  test("race window: SSE data in store, no metadata — merge protects event data from stale GET", () => {
+    // SSE events wrote msg_1 (from shell) into the store
+    const stored = [m("msg_1", "shell-user")]
+    // GET returns empty because it hit the server before shell created messages
+    const fetched: TestMsg[] = []
+    const result = resolveLoadMessagePage<TestMsg>({
+      stored,
+      fetched,
+      hasCachedMeta: false,
+    })
+    // msg_1 from SSE should survive
+    expect(result.map((x) => x.id)).toEqual(["msg_1"])
+    expect(result.map((x) => x.text)).toEqual(["shell-user"])
+  })
+
+  test("race window: SSE + GET both have data — merge fills gaps", () => {
+    // SSE events wrote msg_1
+    const stored = [m("msg_1", "shell-user")]
+    // GET returns msg_1 (from DB, might be more complete) and msg_2
+    const fetched = [m("msg_1", "shell-user-db"), m("msg_2", "shell-assistant")]
+    const result = resolveLoadMessagePage<TestMsg>({
+      stored,
+      fetched,
+      hasCachedMeta: false,
+    })
+    // Both messages present, msg_1 uses fetched version (source of truth for same ID)
+    expect(result.map((x) => x.id)).toEqual(["msg_1", "msg_2"])
+    expect(result.map((x) => x.text)).toEqual(["shell-user-db", "shell-assistant"])
+  })
+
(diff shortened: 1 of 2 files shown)


Astro-Han:
Thanks @gemini-code-assist for the review. Both suggestions have been addressed:

1. **Union type for `mode`**: Changed from `mode?: string` to `mode?: "prepend" | "replace"` to match the call site's type exactly.

2. **Parameter destructuring**: Added `const { mode, stored, fetched, hasCachedMeta } = params` for readability.

Also added a new test case explicitly verifying that fetched data overrides stored data for the same message ID in the race window (intentional: DB snapshot from the same shell execution is at least as fresh as SSE events).

Astro-Han:
Good catch on the same-ID staleness edge case. Two points:

1. **In practice, this is extremely unlikely for home→shell**: the POST (shell) is started before GET (messages) because `navigate()` defers its side-effects until after `handleSubmit` returns via Solid batching. The POST typically reaches the server first, completes shell execution, and only then does the GET read the settled DB. For a sub-second command like `echo hello`, the window for a partial DB read is effectively zero.

2. **The merge strategy is still the correct trade-off here**: the alternative (never merge, always replace) is the current bug — SSE data is completely lost when GET returns empty. The race-window merge fixes that dominant case. The same-ID staleness scenario requires an additional condition (GET hits during shell execution AND SSE delivers newer state for same ID AND GET response arrives after SSE), making it a much narrower edge case than the current bug.

If you'd prefer defense-in-depth, I can add a follow-up issue to add message/part-level freshness checks (e.g., prefer the entry with the more advanced `tool.state.status`) for the merge path. But I think that's a separate concern from fixing the verified blank-page bug.

PR #981, opened 2026-05-28
fix: set repository explicitly in electron-builder config

Astro-Han:
electron-builder reads .git/config for repository info during packaging. On CI runners with persist-credentials: false (all build jobs), the remote URL is absent, causing 'Cannot detect repository by .git/config' on Windows builds. Set the repository field explicitly so electron-builder skips .git/config lookup.

Fixes Windows full build failure (run 26608667785).

### DIFF
--- a/.github/workflows/e2e-artifacts.yml
+++ b/.github/workflows/e2e-artifacts.yml
@@ -139,14 +139,10 @@ jobs:
 
       - run: bun install --frozen-lockfile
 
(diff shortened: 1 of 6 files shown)


Astro-Han:
Superseded by #982 — narrowed to only the electron-builder repository fix. e2e and finalize-latest-yml changes were already merged in #979.

Issue #1056, opened 2026-06-02
[Task] Continue layout.tsx + context/layout.tsx controller split


Astro-Han:
## Goal

Continue the controller/governance split of the two oversized layout files so future shell, sidebar, session, and workspace work can change safely without loading a 2k-line component into one mental context. Carries on the follow-up sequence left in #875 and extends it to `context/layout.tsx`.

- `packages/app/src/pages/layout.tsx` (~2204 lines)
- `packages/app/src/context/layout.tsx` (~1165 lines)

End state: both files mostly compose focused controllers/components and wire props. Each extracted file has one responsibility, is readable in one pass, and has targeted tests around behavior previously hidden inside the layout files.

Refs #474, #875, #606.

## Approach

Serial, flat PRs from latest `dev`, one slice at a time, merge before opening the next. Pure extraction only: no change to user-visible behavior, DOM structure, copy, aria labels, data attributes, command IDs, storage keys, or styling. Target 200-500 lines per PR, hard cap 1000. Do not delete code that merely looks old; prove it dead by grep/tests/runtime path first.

## Slice queue (order may shift with real dependencies)

1. session prefetch controller — pages L666-902
2. session window controller — pages L397-665
3. session/project navigation + pinned — pages L904-1053
4. deep-link / project-open / routing controller — pages L1295-1432
5. workspace lifecycle (create/delete/reset/rename/close) — pages L1493-1707, 1986-2095
6. workspace dialog components — pages L1707-1847
7. workspace drag/drop controller — pages L1847-1985
8. settings/provider/server commands — pages L1228-1294
9. context migration extraction — context L216-318
10. context provider side-effect extraction — context L421+
11. visual sidebar split — pawwork-sidebar.tsx (707 lines), last; needs visual check

## Verification per PR

- `bun run typecheck`
- targeted `packages/app` layout unit tests
- sidebar/layout E2E when route/sidebar/command/workspace wiring changes
- `bun run snap` / `bun run dev:desktop` when rendered shell may change (slice 11)
- `git diff` carries only the slice; no copy/DOM/aria/command-id/storage-key drift

## Out of scope

Backend, persistence schema, packaging, updater, permissions, model harness; UI/UX redesign; touching files already modified by open PRs.

Astro-Han:
[...]

Issue #957, opened 2026-05-27
Flaky unit-app: mock.module leakage across test files causes order-dependent failures


Astro-Han:
Sourced from CI investigation on #951.

## Symptom

`unit-app` intermittently fails in CI while passing locally and on re-run (no code change). Two error signatures, both at module-load/first-use time:

- `TypeError: Persist.workspace is not a function`
- `SyntaxError: Export named 'useNavigate' not found in module '@solidjs/router'`

Tests hit: `createChildStoreManager` (`src/context/global-sync/child-store.test.ts`), `session action readiness` (`src/pages/session/session-action-readiness.test.ts`), `directory-scoped prompt history targets`. Each of these passes in isolation.

Example failing run: https://github.com/Astro-Han/pawwork/actions/runs/26518663246 (re-run of the same commit went green).

## Root cause

`bun test` runs all files in one process, and `mock.module(...)` is **process-global and not auto-restored between files**. A test that replaces `@/utils/persist` or `@solidjs/router` with a partial mock and never restores it leaks that mock into every file that runs afterward. When a leaking file is scheduled before tests that use the real module, those tests see the partial mock (missing `Persist.workspace` / `useNavigate`) and fail.

File execution order differs between local (macOS/APFS) and CI (Linux/ext4), and shifts whenever the test file set changes (e.g. merging new files), so the failure is **order-dependent and intermittent** rather than deterministic.

## Suspected leak sources

Files that `mock.module` persist/router; most have no `mock.restore()` / `afterEach` cleanup:

- src/context/{local,terminal,comments,prompt,models}.test.ts
- src/components/file-tree.test.ts
- src/components/prompt-input/{draft-isolation.integration,submit,submit-ownership}.test.ts
- src/hooks/use-providers.test.ts
- src/pages/session/{session-action-readiness,session-layout,use-session-followups}.test.ts

## Possible fixes

- Restore module mocks after each suite (`mock.restore()` in `afterEach`/`afterAll`), or
- Enforce isolation so module mocks cannot leak across files, or
- Audit the list above and add missing restores.

## Impact / priority

Not a product bug and not a hard blocker (re-run clears it), but it wastes time and erodes trust in the suite ("red? just re-run") which risks masking real regressions. Low-priority but worth fixing as the test count grows.

Astro-Han:
Correction to the "re-run clears it" guidance, from #951 CI:

[...]

Issue #875, opened 2026-05-23
[Task] Extract layout side-effect controllers from layout.tsx


Astro-Han:
## Goal

Reduce the risk of `packages/app/src/pages/layout.tsx` by extracting the first non-visual side-effect controller boundary into focused layout modules, without changing the visible shell, sidebar DOM, routing behavior, storage keys, copy, aria labels, or command IDs.

This is a narrow execution slice under the broader frontend debt governance work. It is not a user-facing feature request.

## Scope

In scope:
- Extract update polling and SDK/desktop notification side effects from `layout.tsx` into focused modules or hooks under `packages/app/src/pages/layout/`.
- Keep the public Layout component responsible for shell composition and prop wiring.
- Preserve existing behavior for sidebar rows, session navigation, workspace switching, settings overlay, keyboard commands, and deep links.
- Add focused tests only when extraction creates pure decision logic worth testing directly.

Out of scope:
- No visual redesign.
- No `PawworkSidebar` component split.
- No TitleBar work.
- No workspace lifecycle extraction.
- No session window, prefetch, or navigation extraction.
- No behavior changes to notifications, updates, deep links, commands, or routing.

## Relevant files or context

Likely files:
- `packages/app/src/pages/layout.tsx`
- `packages/app/src/pages/layout/*`
- `packages/app/src/pages/layout/*.test.tsx`
- `packages/app/src/pages/layout/*.test.ts`
- `packages/app/e2e/sidebar*.spec.ts`

Related issues:
- Refs #687
- Related to #606

## Verification

- Run the app typecheck.
- Run targeted layout unit tests that cover the touched helpers/hooks.
- Run focused sidebar/layout E2E coverage if the extraction touches route, sidebar, command, or workspace wiring.
- Run a snap or manual shell check only if the visible rendered shell changes; this issue should avoid that.
- Confirm `git diff` does not include unrelated UI copy, DOM structure, data attributes, aria labels, command IDs, or storage key changes.

## Execution mode

Investigate and propose a plan first. The split plan is posted as an issue comment; implementation should wait for explicit maintainer approval or assignment before opening a PR.

Astro-Han:
## Design note: `layout.tsx` split plan

### Classification

This should stay a `task`, not an `enhancement`.

Reason: the work does not introduce a user-facing workflow or capability. It is a narrow, behavior-preserving maintainability slice. `tech-debt` is the right supplemental context because the value is future change safety and reviewability.

### Current responsibilities

[...]

PR #1058, opened 2026-06-02
refactor(app): extract session prefetch controller from layout.tsx

Astro-Han:
## Summary

Extract the session message prefetch controller out of `pages/layout.tsx` into a focused module. Slice 1 of the layout governance split tracked in #1056 (carries on the #875 follow-up sequence).

- New `packages/app/src/pages/layout/pawwork-session-prefetch.ts` exports `createPawworkSessionPrefetch`, owning the prefetch queue: per-directory queues, bounded concurrency, pending limit, LRU eviction, priority warm of neighboring sessions, and the route/visibility-driven effects.
- `pages/layout.tsx` now calls the controller with a named dependency object and consumes the returned `{ prefetchSession, warm }`.

Mechanical extraction: the moved code is verbatim, with external references rewired to injected reactive accessors (`params`, `route`, `currentDir`, `visibleSessionDirs`, navigation sessions, `globalSDK`, `globalSync`). Effects still register under the Layout owner. No change to user-visible behavior, DOM, copy, aria labels, command IDs, storage keys, or styling.

`layout.tsx`: 2204 -> 1960 lines.

## Touched files

- `packages/app/src/pages/layout.tsx` — remove inlined prefetch block + now-unused imports; wire controller
- `packages/app/src/pages/layout/pawwork-session-prefetch.ts` — new controller

## Review focus

- Reactive boundary: deps passed as accessors/objects, not frozen values; effects under Layout owner.
- Prefetch queue semantics unchanged (concurrency 2, pending limit 10, span 4, max 10 sessions/dir).
- Import cleanup correct (`Message` type, `retry`, session-prefetch fns, `pickSessionCacheEvictions` removed from layout.tsx; `dropSessionCaches` kept).

## Verification

- `bun turbo typecheck` — green (8/8)
- `eslint` on both changed files — clean
- `bun test` layout suite (`src/pages/layout/*.test.ts(x)`) — 158 pass / 0 fail

## Tests

No new test added: this is a verbatim extraction, and the core decision logic it orchestrates (`shouldSkipSessionPrefetch`, `runSessionPrefetch`, `pickSessionCacheEvictions`) already lives in and is tested under `@/context/global-sync/session-prefetch` and `session-cache`. Directly unit-testing the controller's stateful queue would require mocking several modules (cf. known mock.module leakage in #957), not worth it for a no-behavior-change move. The existing 158 layout tests guard regressions.

## Risk

Low. Mechanical move behind a stable interface. Main risk is a reactivity regression (a dep captured as a value), mitigated by passing accessors and verified via typecheck + existing tests.

Refs #1056

## Summary by CodeRabbit

* **Refactor**
  * Refactored session message prefetching with improved concurrency control and queue management to ensure more efficient background loading of messages across sessions and directories.

### DIFF
--- a/packages/app/src/pages/layout.tsx
+++ b/packages/app/src/pages/layout.tsx
@@ -23,7 +23,7 @@ import { decode64 } from "@/utils/base64"
 import { Button } from "@opencode-ai/ui/button"
 import { Dialog } from "@opencode-ai/ui/dialog"
 import { getFilename } from "@opencode-ai/util/path"
-import { Session, type GlobalSession, type Message } from "@opencode-ai/sdk/v2/client"
+import { Session, type GlobalSession } from "@opencode-ai/sdk/v2/client"
 import { usePlatform } from "@/context/platform"
 import { useSettings } from "@/context/settings"
 import { createStore, produce, reconcile } from "solid-js/store"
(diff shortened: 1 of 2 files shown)


Astro-Han:
Re CodeRabbit's two findings (raw-directory map keys; `prefetchedByDir` not cleared on route/SDK reset): both describe **pre-existing prefetch behavior**, not regressions from this PR. This is a verbatim extraction — the deleted block in the diff uses the same raw-directory keys and the same reset effect that intentionally leaves `prefetchedByDir` alone.

- **Map keys:** the original keyed `prefetchedByDir` / `prefetchQueues` by raw directory; `markPrefetched` normalizes via `workspaceKey()` only where it compares against the current dir. Changing the map keys is a behavior change, out of scope for a pure-extraction slice.
- **LRU on reset:** `prefetchedByDir` is pruned by the visibility-driven effect (the first `createEffect` in the controller, dropping any dir not in `visibleSessionDirs()`), so it is not orphaned across route changes. The route/url reset effect leaving it intact matches the original exactly.

Keeping this slice behavior-identical per #1056. If either is worth changing, it should be a separate prefetch bug PR with its own review, not bundled into the governance extraction.

PR #1068, opened 2026-06-02
refactor(opencode): add Automation.Service and effectify automation route handlers

Astro-Han:
## Summary

Add an Effect `Automation.Service` (request-face operations) and rewrite the `server/instance/automation.ts` route handlers to run through `AppRuntime.runPromiseExit(Effect.gen(...))`. This is **PR 1 of 3** of the automation Effect-migration tracked in #950 — the "request face". The sync namespace facade is left intact; the execution face (scheduler/runner) and facade removal follow in PR2 and PR3.

Two review-sized commits:
1. `feat: add Automation.Service effect layer` — Service + `defaultLayer`, registered in `effect/app-runtime.ts`. Delegates to the existing sync namespace and lifts domain throws into the typed error channel.
2. `refactor: wrap automation route handlers in AppRuntime` — each handler becomes a single `runRoute(c, Effect.gen(...))`; reaches operations via `yield* Automation.Service`.

## Why

`automation/**` and `server/instance/automation.ts` were never effectified, while the other server route handlers were systematically migrated in #1015–1035 (`config.ts` is the template). Per `AGENTS.md` opencode Effect rules, a route handler's target form is a single `AppRuntime.runPromise(Effect.gen(...))`. This pays down that debt for the request face without touching the high-risk scheduler/runner cancellation surface.

## Related Issue

#950

## Human Review Status

Pending

## Review Focus

- **Error mapping in `runRoute`.** Automation errors (`ValidationError`/`ConflictError`/`ActiveRunStillRunningError`) extend plain `Error`, not `NamedError`, so the global `ErrorMiddleware` does not map them. `runRoute` uses `runPromiseExit` + `Cause.squash` to recover the original error instance (a plain `runPromise` would reject with an opaque `FiberFailure`, regressing 404/422/409 → 500) and centrally maps 422/409. Any other error (e.g. `NotFoundError`) is re-rejected so the global middleware still maps it to 404.
- **Instance ALS propagation.** Handlers run inside `Instance.provide(...)`; `AppRuntime.runPromiseExit` is invoked synchronously within that scope, so the AsyncLocalStorage context still reaches the sync namespace calls inside the Effect. The real-HTTP route suite is the safety net here.
- The nested `AppRuntime.runPromise` inside the old `modelValidationDetails` is collapsed into `yield* modelValidation(...)` — no runPromise nested inside Effect.gen.

## Risk Notes

[...]

Issue #950, opened 2026-05-27
[Feature] Unified automation & loop: one scheduled-task primitive


Astro-Han:
### What task are you trying to do?

I want PawWork to run a chat on its own, without me sitting and driving every run:

- **On a schedule** — "every morning at 9, summarize today's email into a brief."
- **Repeatedly until something is true** — "keep fixing the code until the tests pass."
- **Once, later** — "remind me / check back on this in 10 minutes."

Today none of this is possible. Every run is manual and in-session.

### Which area would this change affect?

App flow or product behavior (primary). Also touches model harness / tools / session mechanics, since the same capability is exposed to the agent as a single tool and runs inside the engine.

### What do you do today?

Nothing — there is no scheduling or looping. I have to be present, type the prompt, and watch each run. Confirmed in code: there is **no scheduling/background infrastructure at all** (`packages/opencode/src/cli/cmd/serve.ts` is a passive HTTP server that just waits forever; no cron, interval, or detached-run primitives anywhere).

### What would a good result look like?

**One unified "Automation" concept** — not three separate tools (Loop / Schedule / Reminder). It is modeled in two layers: an **Automation** (the rule: a tagged union of one-shot vs recurring) and an **AutomationRun** (the record of each execution). The rule's orthogonal knobs:

- **Trigger rhythm** — one-shot time · interval-after-completion (loop) · cron (automation)
- **Prompt** — what to do
- **Continue vs fresh** — continue the automation's own ongoing session, or start a new empty one (never pollutes your main chat)
- **Stop condition** — count / condition / never (recurring only)
- **Where it runs** — which project (and, for fresh runs, optionally an isolated worktree)

The **same object** is created by the user (natural language, or a manual form) and by the agent itself (self-scheduling, e.g. "I'll check back in 10 minutes"). One tool, so the agent never has to choose between overlapping primitives.

It runs **locally while PawWork is open**. When an unattended run hits a permission prompt, it parks and shows the existing pending marker in the sidebar; I approve when I return.

### What would count as done?

- I can say "every day at 9am, summarize today's email" in chat → an Automation is created and shown back on a confirmation card that echoes the parsed fields.
- I can create the same Automation via a manual form (title, prompt, project, schedule) and from a template.
- An **"Automations"** section in the sidebar lists my automations; I can view, edit, pause, delete, **Run now**, and review each run's history (including skipped runs and why).
[...]

PR #1081, opened 2026-06-03
refactor(automation): move active-run state into InstanceState

Astro-Han:
## Summary

Move automation's per-directory **active-run execution state** (`activeWriters` set + `activeRuns` map, which hold the in-flight `AbortController`s) out of the synchronous `Instance.state` and into `InstanceState`, read back through the same memoized-runtime bridge that `Bus`/`Vcs` already use. The surviving synchronous `Automation.*` facade is unchanged for all callers, so this is a behavior-preserving relocation of the state container.

- `Automation.layer` becomes `Layer.effect`, owns the `InstanceState` handle, and exposes `activeState()` on the service.
- The module-level `state()` helper now reads through `automationRuntime.runSync(svc => svc.activeState())` (lazy, memoized — tests that use bare `Instance.provide` keep working without building the layer).
- Net diff is a single file: `+47 / -34`.

Part 2 of the automation Effect migration (#950). PR1 (#1068) effectified the request face (`Automation.Service` + route handlers); this PR effectifies the execution-face state container.

## Scope decisions (deliberately narrowed)

The original PR2 sketch also proposed moving the scheduler's **owner state** into `InstanceState` and "effectifying" `scheduler.ts` / `runner.ts`. Both were intentionally left out after analysis (verified with an independent review):

1. **Owner state stays in `Instance.state`.** Instance teardown runs `State.dispose()` (old `Instance.state` disposers) and then `runDisposers()` (InstanceState cache-invalidations), and `runDisposers` fires its callbacks *concurrently* (`Promise.allSettled`). The owner-state disposer calls `scheduler.stop() -> stopOwnedRuns() -> Automation.stopRunByID()`, which reads the **active state** to fire each run's `AbortController`. Today the owner disposer runs in the earlier `State.dispose` phase, so active state is guaranteed alive. Moving owner state into `InstanceState` would put its finalizer in the same concurrent batch as the active-state invalidation, creating a force-dispose race where in-flight runs may not be aborted. The cross-directory `owners` map + `Instance.restore` reach (used by `stopDirectoryOwnedRuns` / `stopAllOwnedRuns`) also cannot be expressed through `InstanceState`'s ALS-scoped API. Safely moving it would require an explicit teardown-ordering API change, which is out of scope for this low-risk PR.
2. **No rewrite of `fire` / `scan` / `becomeOwner` / Bus callbacks.** The scheduler's external `Interface` must stay synchronous (tests assert a run's signal aborts synchronously right after `stop()` / dispose), and its run-loop must keep calling the sync `Automation` facade (deleted later in PR3). Wrapping those sync facade calls in Effect would be cosmetic and carries real regression risk across lease / missed-schedule / writer-conflict / self-published-event-suppression / cancellation logic. `runner.ts`'s `AbortSignal -> SessionPrompt.cancel` listener with `finally` cleanup is correctness-critical and is preserved as-is.

## Test plan

- [x] `tsgo --noEmit` clean
- [x] `automation-routes` / `automation-runner` / `automation-scheduler` / `automation-event-fixtures` — 107 pass, 0 fail
- [x] `automation/cron` / `automation/validation` / `tool/automate` / `event-replay` — 45 pass, 0 fail

### DIFF
--- a/packages/opencode/src/automation/index.ts
+++ b/packages/opencode/src/automation/index.ts
@@ -1,5 +1,7 @@
 import z from "zod"
 import { Context as EffectContext, Effect, Layer } from "effect"
+import { InstanceState } from "@/effect/instance-state"
+import { makeRuntime } from "@/effect/run-service"
 import { BusEvent } from "@/bus/bus-event"
 import { Bus } from "@/bus"
 import { Identifier } from "@/id/id"
@@ -265,10 +267,11 @@ export namespace Automation {
     activeWriters: Set<string>
     activeRuns: Map<string, { writerKey: string; controller: AbortController; runID: string }>
   }
-  const state = Instance.state<State>(() => ({
-    activeWriters: new Set(),
-    activeRuns: new Map(),
-  }))
+  // Per-directory execution state. The container lives in InstanceState (owned by the
+  // Service layer below); the sync facade reads it through a runtime bridge (see `state`).
+  function state(): State {
+    return automationRuntime.runSync((svc) => svc.activeState())
+  }
 
   export type RunExecutor = (input: {
     definition: Definition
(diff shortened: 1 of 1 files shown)


Astro-Han:
### Fresh-eye review (Codex, xhigh)

Independent review of the active-state move found **no P1**, one **P2**, dispositioned below.

**[P2] `state()`'s first `runSync` is not unconditionally synchronous.** `makeRuntime()` (`src/effect/run-service.ts:51`) merges `Observability.layer` into the bridge runtime. With no `OTEL_EXPORTER_OTLP_ENDPOINT`, `Observability.layer` is the synchronous `EffectLogger.layer`; with OTEL configured it is built via `Effect.promise`, so a *cold* `runSync` (before `AppRuntime` warms the layer in the shared `memoMap`) could throw "An asynchronous Effect was executed with Effect.runSync".

**Disposition: not fixed in this PR (out of scope, pre-existing).** This is a property of the shared `makeRuntime` bridge, *not* introduced here — `Bus.subscribe` uses the identical `makeRuntime(Service, layer)` + `runSync` path (`src/bus/index.ts:176,196`) and carries the same exposure. It does not trigger in CI/tests (no OTEL; 152 automation tests pass) or in the real route/scheduler paths (Observability is warmed by `AppRuntime.runPromise` before any sync-facade call, exactly as Bus relies on today). A proper fix would decouple `Observability.layer` from the sync bridge in `run-service.ts` and affect every consumer — a separate change, not this state-relocation PR.

Other checked properties (directory-context resolution through the detached `queueMicrotask`, single shared `InstanceState` handle via the global `memoMap`, dispose/reload freshness) were verified clean.

PR #1114, opened 2026-06-03
feat(opencode): sweep stale agent scratch tmp files on startup (#945)

### LINKED ISSUE #945: [Feature] Auto-cleanup for agent tmp directory

Keep the agent's scratch space (`tmp/`) from accumulating stale files indefinitely. The agent writes temporary files during work — PR drafts, patch files, debug artifacts, etc. — but nothing ever cleans them up.

## What do you do today?

Manually delete files when you notice the directory has grown. After a few weeks of normal use, the directory reached 46 entries / 217 MB of stale artifacts. The only way to deal with it is to remember to clean it, which no user should have to do.

The tmp directory stays clean without user intervention. Options:

- Clean the directory on app startup (the simplest and safest approach — it's a scratch space, no persistent data lives there)
- Delete files older than N days
- Tie cleanup to session lifecycle (clean session-scoped temp files when the session ends)

Whichever approach, running sessions must not be affected — don't delete files currently in use.

## What would count as done?

- After quitting PawWork and reopening, the tmp directory does not contain stale files from previous sessions
- Running sessions are not affected by cleanup
- The cleanup is automatic and transparent — users never need to think about it

## What should stay out of scope?

- Session data, logs, or other persistent app state — only the agent's scratch space
- User workspace files — only the tmp directory under `state/pawwork/tmp`

Both

## Extra context

Path: `~/Library/Application Support/ai.pawwork.desktop/state/pawwork/tmp`

This directory is the designated scratch space for agent tool execution (declared in the system prompt). All files written here are temporary by design but never garbage-collected.

Astro-Han:
## Summary

Closes #945. The agent's `${tmp}` scratch dir (`Global.Path.tmp`) collects PR drafts, patches, and debug artifacts with arbitrary filenames and was never cleaned, so it grew without bound — the reporter saw 46 entries / 217MB over a few weeks.

Adds a best-effort, age-based sweep run once at opencode startup, next to the existing cache-version housekeeping in `global/index.ts`.

## Design

- `sweepScratch` removes any top-level scratch entry whose **whole subtree** has gone untouched longer than `SCRATCH_MAX_AGE_MS` (24h). Scratch is write-and-forget, so 24h is safe and far tighter than `Truncate`'s 7d (whose outputs are referenced later by saved-path hints).
- Age is `max(mtimeMs, ctimeMs)` over the subtree, computed **recursively**, so a stale parent dir holding one freshly written nested file is kept (directory mtime alone does not reflect nested writes); a restored-mtime file (`cp -p` / unzip) reads as recent via `ctime` and is kept.
- `lstat`-based: symlinks are removed as links, never followed to their target.
- Uncertain entries (stat/readdir failure) return `Infinity` and are left in place.
- **Not awaited** at import: startup never blocks on it, and removal failures are swallowed per entry.

## Why age-based, not wipe-on-startup

A standalone `opencode` CLI can run concurrently with the desktop session, sharing the same per-user `Path.tmp`. An unconditional startup wipe could clobber another live process's in-use files. Age-based keeps any recently touched file.

**Trade-off (called out for review):** a file from a session that ended minutes ago survives until it ages out, so this does not *immediately* clear the previous session's files on the very next launch. The issue's actual pain is unbounded multi-week accumulation, which this fully resolves. Doing literal "clear last session on reopen" safely would need session-scoped tmp dirs or per-process ownership markers — a much larger change, out of proportion for a P3.

## Review follow-up (codex)

Codex review raised a [P2]: the lock-free age check has an irreducible TOCTOU — a process that begins writing into an *already-stale* subtree between the freshness scan and `fs.rm` can still race with removal. After a second codex consult on the right altitude, the conclusion was to **accept the residual** rather than add ownership locks: `${tmp}` is documented as per-command short-lived scratch, the sweep only touches entries untouched for 24h, and the remaining race requires a live process to start writing into an already-stale subtree within the scan→rm window — outside intended usage, while a cooperative ownership/liveness protocol is out of scope for this P3 GC. The follow-up commit narrows the code comments so they no longer claim a hard cross-process guarantee (they previously said "cross-process safe" / "never deleted", which the lock-free check cannot promise). `Flock` was evaluated as a single-process gate and ruled out: its public API only blocks/throws (no non-blocking try-acquire) and scratch writers do not hold that lock.

## Verification

- `bun test test/global` — 9 pass. New `scratch.test.ts` covers: remove-stale, keep-fresh, ctime-recent (`cp -p`), recursive keep (stale dir + one fresh nested file), recursive remove (whole subtree stale), symlink no-follow (link removed, target survives), missing-dir no-op.
- `bun run typecheck` clean.
- The sweep runs in the opencode server process (CLI and embedded desktop server alike); behavior is identical on macOS/Windows. No Electron-specific branch touched.

Design reviewed with codex consults before implementation and after review.

## Summary by CodeRabbit

* **New Features**
  * Automatic cleanup of temporary scratch files: Files and directories older than 24 hours are now cleaned up automatically on startup
  * Non-blocking operation: Cleanup runs in the background without delaying application startup
  * Disk space optimization: Removes stale temporary files to help manage disk usage

* **Tests**
  * Comprehensive test suite added for cleanup functionality, covering file retention, deletion, and edge cases

### DIFF
--- a/packages/opencode/src/global/index.ts
+++ b/packages/opencode/src/global/index.ts
@@ -25,8 +25,8 @@ if (version !== CACHE_VERSION) {
   await Filesystem.write(path.join(Path.cache, "version"), CACHE_VERSION)
 }
 
-// Drop agent ${tmp} scratch artifacts left untouched past the retention window so
+// Drop per-command ${tmp} artifacts left untouched past the retention window so
 // the scratch dir does not grow without bound (#945). Not awaited: startup must
-// not block on it, and it is age-based, so a concurrently running opencode
-// process's recent files are never swept.
+// not block on best-effort cleanup. The sweep relies on ${tmp} being short-lived
+// scratch, not long-lived session-owned directories.
 void sweepScratch({ dir: Path.tmp, now: Date.now(), maxAgeMs: SCRATCH_MAX_AGE_MS }).catch(() => {})
--- a/packages/opencode/src/global/index.ts
+++ b/packages/opencode/src/global/index.ts
@@ -2,6 +2,7 @@ import fs from "fs/promises"
 import path from "path"
 import { Filesystem } from "../util/filesystem"
 import { Global, Path } from "@opencode-ai/core/global"
+import { sweepScratch, SCRATCH_MAX_AGE_MS } from "./scratch"
 
 export { Global }
 
@@ -23,3 +24,9 @@ if (version !== CACHE_VERSION) {
   } catch (e) {}
   await Filesystem.write(path.join(Path.cache, "version"), CACHE_VERSION)
 }
+
+// Drop agent ${tmp} scratch artifacts left untouched past the retention window so
+// the scratch dir does not grow without bound (#945). Not awaited: startup must
+// not block on it, and it is age-based, so a concurrently running opencode
+// process's recent files are never swept.
+void sweepScratch({ dir: Path.tmp, now: Date.now(), maxAgeMs: SCRATCH_MAX_AGE_MS }).catch(() => {})
--- a/packages/opencode/src/global/scratch.ts
+++ b/packages/opencode/src/global/scratch.ts
@@ -42,9 +42,11 @@ async function newestTimestamp(entry: string): Promise<number> {
 /**
  * Remove scratch entries whose whole subtree has gone untouched for longer than
  * `maxAgeMs`. Best-effort and idempotent: a missing directory is a no-op and a
- * per-entry removal failure is swallowed so it never blocks startup. Because it
- * is age-based it is cross-process safe — a concurrently running opencode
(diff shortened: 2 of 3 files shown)


Astro-Han:
Addressed both gemini review points in 072038f25a:
- **Short-circuit traversal**: replaced `newestTimestamp` (full-subtree max) with `isSubtreeStale(entry, cutoff)`, which returns on the first node at/after cutoff — one recent file keeps the entry without walking the rest of a large kept tree. Verdict is identical to the prior max-timestamp compare.
- **Windows rm retries**: `fs.rm` now passes `maxRetries: 30, retryDelay: 100` to ride out transient AV/indexer locks on otherwise-stale entries.

test/global still 9 pass, typecheck clean.

PR #1133, opened 2026-06-03
fix(installation): type and sanitize upgrade failures

Astro-Han:
## Summary

Type and sanitize `Installation` upgrade failures: give `UpgradeFailedError` a `message`, and replace raw command stderr with a sanitized, user-facing string.

## Why

Two user-facing defects in `installation/index.ts`:

1. **Empty desktop error.** `UpgradeFailedError` had no `message` getter, so the TaggedError default `.message` is empty. The desktop/server upgrade API surfaces `err instanceof Error ? err.message : String(err)` (`server/instance/global.ts:507`), so a failed upgrade returned a blank error to the user.
2. **Raw stderr leak.** The failure branch put raw `result.stderr` into the error, leaking tokens / arbitrary command output.

Fix (boundary: `installation/index.ts` + its test only):
- `override get message()` returns `stderr` so the desktop error is non-empty.
- `upgradeFailure(method, result?)` produces a sanitized `Upgrade failed for {method} (exit code {code}).`; `choco` keeps its actionable "not running from an elevated command shell" hint.
- Type the curl install-script path with `mapError` instead of `orDie`, so a fetch/spawn failure surfaces as a typed `UpgradeFailedError` rather than a defect.

Re-implements upstream `anomalyco/opencode` #28883 (`536ee857c6`). Thanks to the upstream author (Shoubhit Dash).

## Related Issue

No tracked issue; surfaced via the upstream value-mining ledger.

## Human Review Status

Pending

## Review Focus

- `upgradeFailure` covers every failure exit: package managers (with exit code), `choco` (hint preserved), unknown method, and the curl `mapError` (no result → "Upgrade failed for curl.").
- `get message()` returning `stderr` is what the desktop API needs; verify no caller depended on the old empty message.

## Risk Notes

Low. Error *contents* change (sanitized, non-empty); the error type and the `Effect.Effect<void, UpgradeFailedError>` signature are unchanged. The curl path moves from defect (`orDie`) to typed failure, which callers already handle (`cli/cmd/upgrade.ts` checks `instanceof UpgradeFailedError`).

## How To Verify

```text
bun test test/installation/installation.test.ts  → 12 pass (8 existing + 4 new upgrade tests)
red->green: the 4 new tests fail without the fix (raw stderr leak + empty message)
bun run typecheck (tsgo --noEmit)                → clean
```

## Screenshots or Recordings

N/A — no UI change.

## Checklist

- [x] **Type label** — this PR carries exactly one of `bug`, `enhancement`, `task`, `documentation`.
- [x] **Routing labels** — this PR carries at least one of `app`, `ui`, `platform`, `harness`, `ci`.
- [x] **Priority label** — this PR carries exactly one of `P0`, `P1`, `P2`, `P3`.
- [x] Human Review Status above is set to `Pending`, `Approved by @<reviewer>`, or `Not required: <reason>`.

### DIFF
--- a/packages/opencode/src/installation/index.ts
+++ b/packages/opencode/src/installation/index.ts
@@ -82,7 +82,13 @@ export namespace Installation {
 
   export class UpgradeFailedError extends Schema.TaggedErrorClass<UpgradeFailedError>()("UpgradeFailedError", {
     stderr: Schema.String,
-  }) {}
+  }) {
+    // Without this, the TaggedError default message is empty, so the desktop upgrade API
+    // (server/instance/global.ts surfaces `err.message`) returns a blank error.
+    override get message() {
+      return this.stderr
+    }
+  }
 
   // Response schemas for external version APIs
   const GitHubRelease = Schema.Struct({ tag_name: Schema.String })
@@ -156,6 +162,17 @@ export namespace Installation {
           return "opencode"
         })
 
+        // Sanitized, user-facing failure text. Never surface raw stderr (it can contain tokens or
+        // arbitrary command output); choco keeps its actionable elevated-shell hint.
+        const upgradeFailure = (
+          method: Method,
+          result?: { code: ChildProcessSpawner.ExitCode; stdout: string; stderr: string },
+        ) => {
+          if (method === "choco") return "not running from an elevated command shell"
+          if (result) return `Upgrade failed for ${method} (exit code ${result.code}).`
+          return `Upgrade failed for ${method}.`
+        }
+
         const upgradeCurl = Effect.fnUntraced(
           function* (target: string) {
             const response = yield* httpOk.execute(HttpClientRequest.get("https://opencode.ai/install"))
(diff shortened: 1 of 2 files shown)


Astro-Han:
Codex review: GATE PASS (no P1). It raised one P2 — the tests gate the curl *fetch* failure (404) but not a curl *spawn* failure. Decision: not adding a separate spawn-failure test. The single `Effect.mapError(() => upgradeFailure("curl"))` wraps the **entire** `upgradeCurl` pipeline uniformly, so the 404 fetch-failure test already gates that exact line; a spawn failure hits the identical code path. Simulating a spawn failure would require constructing an Effect `PlatformError` (the spawner factory's error type), coupling the test to Effect internals for near-duplicate coverage. Codex independently verified at runtime that a failing spawner yields the typed, sanitized `Upgrade failed for curl.` error and that real defects (`Effect.die`) still reject rather than masquerade as `UpgradeFailedError`.

Astro-Han:
Applied for the curl path (adapted: PawWork's `log.error` returns `void`, so it's wrapped in `Effect.sync` inside `tapError`). The raw cause is now logged before the message is sanitized. Scoped to curl only: its `mapError` carries no exit code and is hard for a non-technical user to reproduce, whereas the package-manager paths already encode the exit code in the message and match upstream.

Issue #1264, opened 2026-06-11
[Task] Align sent-bubble attachment rendering with the composer chip design


Astro-Han:
## What needs to be done

PR #1247 (issue #1242) unified the composer's attachment entry points into one chip design: file chips with icon + filename + size, image chips with thumbnails, a bare close-small remove glyph, and a capability warning triangle, all sharing one right-edge rail inside the chip.

The sent message bubble still renders attachments with the older design (`packages/ui/src/components/message-file.ts` classification + `user-message.tsx` `renderAttachments`): square tiles with a filename overlay, different sizing, no shared visual language with the composer.

Bring the bubble's attachment row in line with the composer chip design so an attachment looks the same before and after sending.

## Scope notes

- Rendering only; the wire format and the `metadata.attachment` chip classification from #1242 stay as they are.
- The bubble row is display-only (no remove button), so the chip visual carries over minus interactive affordances; click-to-expand for images should be kept.
- Follow `docs/DESIGN.md` and verify with the existing `attachment-chips` snap target plus a bubble-side snap or manual walk.

## Context

- Composer chip implementation: `packages/app/src/components/prompt-input/attachment-chips.tsx`
- Bubble rendering: `packages/ui/src/components/message-part/user-message.tsx`
- Raised during PR #1247 manual review; deliberately kept out of that PR's boundary.

Astro-Han:
Closing as V1 track — superseded by V2 on the official DeepSeek Harness (see #1568). This V1 app/harness/UI area will not be fixed on V1; DSH now owns this surface on `v2/dsh`. If it reappears on V2, please reopen with V2 steps.

PR #1266, opened 2026-06-11
fix(app): fork dialog crash from palette and blank attachment-only previews

Astro-Han:
## Summary

Two fixes from the post-merge review of #1247, both on the fork-dialog user path:

1. **Fork dialog crash from the command palette (found by the new e2e, pre-existing).** Palette command flows call `dialog.close()` and then run the command handler inside the 100ms close animation, while `active` is still set. `useDialog.show` anchored the new dialog to `ctx.active.owner` — the dying palette's owner at layout level — so `DialogFork`'s `useSync()` threw "Sync context must be used within a context provider" and the session route fell into the error boundary. Since the palette is the only entry point for `session.fork`, the fork dialog was unreachable. Fix: a dialog mid-close no longer serves as the owner anchor; the new dialog uses its caller's owner.

2. **Attachment-only messages rendered as a blank row in the fork list.** Such a message persists an empty text part, and the preview only read text parts. The revert banner's `[image:name]` / `[file:path]` mapping is now extracted into a shared `promptPreviewText` helper (`utils/prompt.ts`); the revert banner delegates to it, and the fork dialog falls back to it whenever the text part is empty or missing. Messages with neither text nor file parts stay hidden.

Also adopts the review's verification ask on dedupe semantics: a new pinning test locks that a selection-scoped inline pill and a whole-file chip for the same path submit as **two** parts — they are semantically different references, and collapsing on path would silently narrow the whole-file attachment to the selection.

## Why

Post-merge review of #1247 flagged the blank fork row (P3) and asked for explicit dedupe-semantics coverage. The owner-anchoring crash surfaced while writing the e2e for the blank row and is a hard prerequisite for that path to work at all, so it lands in the same review boundary.

## Related Issue

Follow-up from the post-merge review of #1247 (fixed directly per maintainer direction, no standalone issue). Related: #1264 tracks sent-bubble attachment rendering alignment separately.

## Human Review Status

Pending

## Review Focus

- `packages/ui/src/context/dialog.tsx`: the owner-anchoring change affects every dialog-from-dialog flow. The rule is minimal — only a dialog that is already closing stops anchoring; an active, non-closing dialog keeps the previous chaining behavior.
- `promptPreviewText` is now the single preview mapping for revert banner and fork dialog; the revert banner's behavior must be unchanged (same map/join/collapse/fallback sequence, moved verbatim).
- `dialog-fork.tsx`: messages with neither a non-synthetic text part nor a file part are still skipped, so non-promptable rows do not appear with a bare placeholder.

## Risk Notes

- The dialog owner change alters context resolution for dialogs opened while another dialog is closing: they now resolve through the caller's owner chain. For every current call site this is the owner that actually has the providers the dialog needs; flows opening a dialog while another is fully open (not closing) are unchanged.
- `e2e/models/model-picker-visual.spec.ts` fails 3 hover-color assertions locally on a clean tree as well (environment flake, pre-existing); verified unrelated by stash-rerun.

## How To Verify

```text
bun run test:e2e e2e/session/fork-dialog-preview.spec.ts (packages/app): 1 passed
  - real user path: seed attachment-only message -> open palette -> run Fork from message
    -> dialog lists "[file:<path>]" placeholder row (RED before both fixes: crash, then blank row)
bun run test:e2e e2e/app/palette.spec.ts e2e/app/palette-viewport.spec.ts e2e/sidebar/sidebar-session-search.spec.ts: 9 passed
bun test (packages/app): 1886 pass / 0 fail (new: promptPreviewText unit tests; selection-pill + chip pinning test)
bun test (packages/ui): 722 pass / 0 fail
bun run typecheck (packages/app, packages/ui): clean
bun run lint (repo root): clean
```

Visual check: the changed surface (fork dialog list rows) is exercised end-to-end by the new `fork-dialog-preview` e2e through the real palette route; row markup/styling is unchanged, only the text content of the preview row.

## Screenshots or Recordings

Covered by the e2e assertion on the visible row text; no styling change. Can attach the e2e trace on request.

## Checklist

- [x] **Type label** — this PR carries exactly one of `bug`, `enhancement`, `task`, `documentation`.
- [x] **Routing labels** — this PR carries at least one of `app`, `ui`, `platform`, `harness`, `ci`.
- [x] **Priority label** — this PR carries exactly one of `P0`, `P1`, `P2`, `P3`.
- [x] Human Review Status above is set to `Pending`, `Approved by @<reviewer>`, or `Not required: <reason>`.
- [x] **(conditional)** I manually checked visible UI or copy changes when needed, with screenshots or recordings (e2e asserts the visible row content through the real user path).
- [x] **(conditional)** I called out docs, release notes, dependencies, permissions, credentials, deletion behavior, generated content, or local file changes when relevant.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed "Open file" command in the palette to properly display the file picker dialog.
  * Fixed fork dialog preview to correctly display file attachments.
  * Improved dialog handling for better context preservation during dialog transitions.

* **Tests**
  * Added E2E tests for file-open command and fork dialog with attachments.
  * Added tests for attachment handling in prompts.

### DIFF
--- a/packages/app/e2e/files/file-open.spec.ts
+++ b/packages/app/e2e/files/file-open.spec.ts
@@ -1,6 +1,7 @@
 import fs from "node:fs/promises"
 import path from "node:path"
 import { test, expect } from "../fixtures"
+import { openPalette } from "../actions"
 import { promptSelector } from "../selectors"
 
 test("can open a file tab from the search palette", async ({ page, project }) => {
@@ -39,4 +40,22 @@ test("can open a file tab from the search palette", async ({ page, project }) =>
 
   const tabs = page.locator('[data-component="tabs"][data-variant="normal"]')
   await expect(tabs.locator('[data-slot="tabs-trigger"]').first()).toBeVisible()
+})
+
+// The palette's own "Open file" row re-shows DialogSelectFile from inside the
+// closing palette; that second show has no LayoutPageContext wrapper of its
+// own, so it must be re-provided or useLayoutPage throws and the picker never
+// appears.
+test("palette Open file row chains into the file-only picker", async ({ page, gotoSession }) => {
+  await gotoSession()
+
+  const palette = await openPalette(page)
+  await palette.locator('[data-slot="list-item"][data-key="command:file.open"]').click()
+
+  const picker = page
+    .getByRole("dialog")
+    .filter({ has: page.getByPlaceholder(/search files/i) })
+    .first()
+  await expect(picker).toBeVisible()
+  await expect(picker.locator('[data-slot="list-item"][data-key^="command:"]')).toHaveCount(0)
 })
--- a/packages/app/e2e/session/fork-dialog-preview.spec.ts
+++ b/packages/app/e2e/session/fork-dialog-preview.spec.ts
@@ -6,6 +6,7 @@
  */
 import { writeFileSync } from "node:fs"
 import path from "node:path"
+import { pathToFileURL } from "node:url"
 import { test, expect } from "../fixtures"
 import { withSession, openPalette } from "../actions"
 
@@ -33,7 +34,7 @@ test("fork dialog previews an attachment-only message as a file placeholder", as
         {
           type: "file",
           mime: "text/plain",
-          url: `file://${filePath}`,
+          url: pathToFileURL(filePath).href,
           filename,
           metadata: { attachment: true },
         },
--- a/packages/app/e2e/session/fork-dialog-preview.spec.ts
+++ b/packages/app/e2e/session/fork-dialog-preview.spec.ts
@@ -1,0 +1,55 @@
+/**
+ * Fork dialog preview for attachment-only messages (post-merge review of
+ * #1247): a message submitted with only an attachment chip carries an empty
+ * text part and used to render as a blank row in the fork list. The preview
+ * now falls back to the same [file:path] placeholder the revert banner uses.
+ */
+import { writeFileSync } from "node:fs"
+import path from "node:path"
+import { test, expect } from "../fixtures"
+import { withSession, openPalette } from "../actions"
+
+test("fork dialog previews an attachment-only message as a file placeholder", async ({ page, project }) => {
+  await project.open()
+  const sdk = project.sdk
+
+  await withSession(sdk, "fork attachment preview", async (session) => {
+    project.trackSession(session.id)
+
+    const filename = "fork-preview-attachment.txt"
+    const filePath = path.join(project.directory, filename)
+    writeFileSync(filePath, "attachment body\n")
+
+    await sdk.session.prompt({
+      sessionID: session.id,
+      noReply: true,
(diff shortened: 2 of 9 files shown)


Astro-Han:
Round disposition for the re-raised P2 (restore does not require `metadata.attachment`): **maintained, third round** — same trade as documented in #1247's review threads and in the code comment on `attachmentFromFilePart`.

Requiring the tag turns every untagged source-less `file://` part — context items, comment mentions, and all pre-#1247 history — into a silent drop on undo/fork, because the prompt model has no context-row restore channel to receive them. Restoring them as visible chips loses nothing: the resubmitted wire payload is identical, and the only delta is bubble rendering plus visibility in the composer.

Reversal condition, left on record: when a context-row restore channel exists, restore tagged parts as chips and untagged ones as context rows; at that point the suggested tag-required tests become the right pins. Until then the tag-required variant trades silent data loss for label accuracy, which is the worse side of the trade.

Astro-Han:
Codex design-review round: **P1 adopted** in e42c497431. The closing-flag rule would have anchored the palette's own `file.open` branch (close() then show() from inside the closing palette root) to an owner `show()` immediately disposes. Replaced with an ancestry rule: a caller inside the active dialog root keeps the dialog's outer owner (survives the dispose); a caller outside uses its own owner. The `closing` field is gone — one rule covers both flows.

Verified: fork-dialog-preview + file-open + palette e2e (8 passed), packages/ui 722 + packages/app 1886 unit tests, typecheck and lint clean.

Codex verdicts kept on record: #1265 consolidation point and namespace-alias style judged appropriate (no new reverse dependency; app keeps SDK-only boundary, frontend mirror stays); `promptPreviewText` extraction correct; leaving `followupPreviewText` separate is right — it previews a live unsent Prompt with first-non-empty-line semantics, not restored history parts.

Astro-Han:
Codex fresh-eye round 2, both findings adopted in d4babf0d0b:

- **P1 (palette's chained Open file picker loses LayoutPageContext)**: real, and pre-existing on dev — the old `ctx.active.owner` chain anchored the second `DialogSelectFile` the same way, above the `LayoutPageContext.Provider` that the layout wraps around the FIRST palette element only, so `useLayoutPage` threw and the picker never appeared. The ancestry rule kept that anchoring (by design, the outer owner survives the dispose); the fix re-provides the context around the second show. Pinned by a RED→GREEN e2e clicking the actual `command:file.open` palette row.
- **P2 (fork list stamps [attachment] on empty-text, no-file messages)**: guard reordered — trim the text first, skip when there is neither visible text nor a file part, so the placeholder fallback only serves messages that actually carry attachments.

Codex verdicts on record: the ancestry owner rule judged the correct minimal shape (better than always-caller-owner, deferred show, or removing the chain); #1265 re-reviewed with no findings.

Verified: file-open (incl. new chained test) + fork-dialog-preview + palette e2e 9 passed; app 1886 unit tests, typecheck, lint clean.

PR #1331, opened 2026-06-15
fix(permission): persist "always allow" grants so they survive a restart

Astro-Han:
## Summary

On an "always allow" reply, the granted rules were pushed into the in-memory `approved` list but never written back to `PermissionTable`. Capture the project id in the permission `State` at load, and on an "always" reply upsert the full cumulative `approved` ruleset under the project's primary key so the grant persists.

## Why

The `approved` list is loaded from the project's `PermissionTable` row at instance startup (`permission/index.ts:186`), but nothing ever wrote it back — `approved.push(...)` only mutated memory. So every app restart (or instance reload) dropped all "always allow" grants and the next action re-asked. This affected **every** tool that supports "always" (edit, bash, browser, …), and was the real reason "always" felt like it never stuck.

## Related Issue

None — found while investigating the embedded-browser permission prompts (context: PR #1330).

## Human Review Status

Pending

## Review Focus

- The upsert in `reply()`: it writes the whole cumulative `approved` array under `project_id` (the table's primary key) via `onConflictDoUpdate`. Confirm that's the right shape and that writing inside the `reply` effect (synchronous `Database.use`, current instance's DB) is correct.
- That the write is gated to `existing.info.always.length > 0`, so plain "once" replies and empty-`always` browser asks don't write.

## Risk Notes

Behavior change limited to permission persistence: "always allow" now survives restarts as intended. No schema/migration change — `PermissionTable` already exists and is already read at startup; this only starts writing the column that was already being read. No platform/UI surface touched.

Skipped conditional checklist items: visible-UI item (no UI or copy changed); platform item (no Electron/packaging/updater/signing/paths surface — server-side permission persistence).

## How To Verify

```text
opencode typecheck (tsgo --noEmit): clean
bun test test/permission/: 113 pass / 0 fail (includes new reload test)
new test "reply - an 'always' grant survives an instance reload":
  approve always -> Instance.disposeAll() -> reload same dir -> ask again resolves with no prompt
  confirmed it FAILS (re-asks, ask pends -> times out) with the persistence write reverted
```

## Screenshots or Recordings

None — no visible UI or copy changed.

## Checklist

> **How to use this checklist:**
>
> - Tick a box by replacing `[ ]` with `[x]`. Do not edit, add, or remove items.
> - The bot-applied label items can only be honestly ticked AFTER the PR is opened and the labeler / priority-triage bots have run — return to the PR description and tick them then.
> - Most items are required. The few that are conditional are explicitly marked **(conditional)**; for those, leave unticked if they truly do not apply and explain why in Risk Notes. All other items must be ticked before requesting human review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Permissions marked as "always allow" now correctly persist and remain in effect after application restarts, instead of being lost on reload.

### DIFF
--- a/packages/opencode/src/permission/index.ts
+++ b/packages/opencode/src/permission/index.ts
@@ -173,6 +173,21 @@ export namespace Permission {
     return evalRule(permission, pattern, ...rulesets)
   }
 
+  // Union two rulesets, dropping exact duplicates, existing entries first. Used
+  // to merge the persisted permission row with an instance's in-memory grants
+  // so neither side is lost and the row does not grow unboundedly.
+  function mergeRulesets(existing: Ruleset, incoming: Ruleset): Ruleset {
+    const key = (rule: Ruleset[number]) => JSON.stringify([rule.permission, rule.pattern, rule.action])
+    const seen = new Set(existing.map(key))
+    const merged = [...existing]
+    for (const rule of incoming) {
+      if (seen.has(key(rule))) continue
+      seen.add(key(rule))
+      merged.push(rule)
+    }
+    return merged
+  }
+
   export class Service extends Context.Service<Service, Interface>()("@opencode/Permission") {}
 
   export const layer = Layer.effect(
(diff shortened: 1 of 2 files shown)


Astro-Han:
## Follow-up: merge persisted grants instead of clobbering (commit 88fde5a63c)

An independent Codex review flagged a lost-update in the first commit that I verified is real:

**The bug.** The persistence write replaced the entire project row with the current instance's cached `approved`. Project id is the shared `git-common-dir` (`project/project.ts` resolves via `git rev-parse --git-common-dir` and caches the id in the shared `.git/opencode`), so different worktrees / subdirectories of one repo resolve to the **same** row — but instance state is keyed by `Instance.directory` (`project/instance.ts:98`), so each holds its **own** `approved` loaded at startup. An "always" reply in instance B (whose `approved` was loaded before instance A's grant) overwrote the row with only B's grants, dropping A's. PawWork's multi-worktree workflow hits this directly.

**The fix.** Re-read the current row and union it with `approved` inside one synchronous `Database.use` callback before writing. Single process + no `await` between read and write makes the read-merge-write atomic against other fibers, so the row converges to the union of all instances' grants. `mergeRulesets` dedupes by `(permission, pattern, action)` to bound growth.

**Test.** Added a regression test: two sibling instances of one git repo each grant a different "always"; both survive an instance reload. Confirmed it **fails** (the first grant re-asks on reload) when the merge is reverted to a whole-row write.

Verification: `bun test test/permission/` → 114 pass / 0 fail; `tsgo --noEmit` clean.

PR #1404, opened 2026-06-19
feat(remote): WeChat iLink adapter with one-step QR pairing (#1188)

Astro-Han:
## Summary

Adds **WeChat** as a remote-control channel for the mobile-companion bridge, over Tencent's official **iLink Bot API** (`bot_type=3`, the WeChat ClawBot slot) — a NAT-friendly HTTP long-poll, structurally like the Telegram adapter (no SDK, no public IP, no relay we operate). It is a pure adapter against the post-#1390 `PlatformPairer` contract: no engine/supervisor/gateway change.

- **remote-bridge** (`platforms/wechat/`): `client.ts` (iLink long-poll), `login.ts` (QR login → bot token + base URL + paired user id), `platform.ts` (the `Platform`; no `reconstructReplyCtx` — iLink has no proactive push, so a restored push is logged and skipped).
- **desktop-electron**: `WeChatPairer` (one-step QR pairing, main-process QR encoding via `qrcode`), `RemoteAccount`/`isAccount` wechat variant, widened `PairingProgress`.
- **app**: WeChat channel row + brand mark, QR sign-in connect dialog (`qr` pairing-event phase; auto-approves on confirm because the in-WeChat tap already authorized), en/zh copy.

## Why

Second of three flat PRs re-landing mobile-companion Wave 1, after the #1390 multi-channel foundation. The iLink wire contract was **verified live against the real service first** (a `bot_type=3` QR login + a `getupdates`/`sendmessage` round-trip with a real WeChat account), which corrected three things a prior spike got wrong:

| | spike assumed | live reality | fix |
|---|---|---|---|
| QR | `qrcode_img_content` is an image | it's a `liteapp.weixin.qq.com` **URL** | main process QR-encodes it (new `qrcode` dep) |
| login poll | one-shot, 15s timeout | `get_qrcode_status` **long-polls ~30s** | client timeout 40s |
| binding | needs a "message the bot" step | confirm returns `ilink_user_id` (== inbound `from_user_id`) | **one-step pairing**: scan + confirm IS the binding |

Confirmed correct as-is: all field names, `channel_version=1.0.2`, status `confirmed`, the `message_type`/`message_state`/item enums, and the `context_token` reply mechanism.

## Related Issue

Part of #1188 (mobile companion). Follows #1390.

## Human Review Status

Pending

## Review Focus

[...]

Issue #1426, opened 2026-06-20
[Task] WeChat: recover from an expired login session (in-app re-scan) instead of staying degraded


Astro-Han:
## Goal

When a WeChat (iLink) login session expires, the channel guides the user back to a re-scan and recovers **in place — keeping the conversation** — instead of degrading and requiring a manual disconnect + reconnect. This is the "re-login path" that #1404 deliberately left deferred (it currently only does "expired → degrade").

## Key finding: most of the machinery already exists

A 2026-06-29 code read confirms this is largely **wiring existing parts together**, not new protocol work:

- **QR pairing flow is built end-to-end** — `client.ts` `getBotQrcode` + `getQrcodeStatus`, `remote-connect-dialog.tsx` (`DialogConnectRemote`), and the main-process pairing stream (`startPairing` / `onPairing` / `confirmPairing`). WeChat already opens straight into the QR and treats scan + in-app confirm as the authorization.
- **Re-pair already swaps the token while keeping the conversation** — `gateway.ts` `addPlatform` runs prepare-first / swap-after-success and, on a same-name re-pair, **retires the old loop only on success and keeps its session pointers** (gateway.ts doc: "a re-pair continues the conversation rather than forgetting it"); a failed build leaves the old channel serving. A degraded channel is still in `desiredPlatforms`, so a same-name `addPlatform` hits exactly this retire+swap path.
- **The fatal-vs-transient signal already exists at the bridge boundary** — `isFatalWeChatError` separates session-expired/auth failures (`ret -14`, HTTP 401/403) from transient blips. It just isn't surfaced to the UI.

So the recovery itself reuses the existing pairing dialog + `addPlatform` swap. The remaining work is the three gaps below.

## Scope — the three gaps

1. **A recovery affordance on the degraded channel row.** Today a `degraded` row in `remote-surface.tsx` offers only "Disconnect". Add a "Reconnect / Re-scan" action that opens `DialogConnectRemote(platform)` for that channel; a successful re-scan swaps the fresh token in via `addPlatform` (same name → keeps session pointers). UX decision (propose, don't assume): one-click "reconnect" on the degraded row vs. auto-opening the QR flow on the fatal error. Recommend one-click, to avoid stealing focus mid-work.

[...]

Issue #1188, opened 2026-06-05
[Feature] Mobile companion: monitor and interact with running agents from phone


Astro-Han:
### What task are you trying to do?

When PawWork agents are running long tasks on my desktop, I want to check progress, see results, and give feedback from my phone — without being physically at my computer.

### Which area would this change affect?

App flow or product behavior

### What do you do today?

Today I have to stay at my computer or use a generic remote desktop app (which is clunky for this purpose — tiny text, no voice input, bad for mobile screens). If I walk away during a long agent run, I either wait to come back or miss when the agent finishes and needs my input. The automation/scheduling feature makes this worse — tasks fire while I'm away, and I can only see results when I return to the desktop.

### What would a good result look like?

A lightweight mobile interface (web-based or native) that connects to the running PawWork desktop instance:

1. **Session monitoring:** See which sessions are running, their progress, and recent output — at a glance on a phone screen.
2. **Push notifications:** Get notified on the phone when an agent completes, errors, or needs permission/input (extends the desktop notification system to mobile).
3. **Quick interaction:** Approve permissions, answer agent questions, or send follow-up prompts from the phone without a full keyboard.
4. **Result review:** Read agent output, view generated artifacts (text, tables, basic HTML), and share results — all from mobile.

### What would count as done?

**Phase 1 — Remote monitoring + push notifications:**
- A web-based mobile UI (accessible via local network or secure tunnel) shows active sessions and their status.
- Push notifications reach the phone when sessions complete or need attention.
- The user can view session output and conversation history from the phone.

**Phase 2 — Remote interaction:**
- The user can approve/deny permission requests from the phone.
- The user can send text prompts or follow-up messages to active sessions.
- The user can start new sessions from saved task cards / automations.

### What should stay out of scope?

- Full feature parity with the desktop app — mobile is a companion, not a replacement.
- File editing or terminal access from mobile.
- Building a standalone mobile app in v1 — a responsive web UI served by the desktop instance is sufficient.
- Voice input on mobile (users already have OS-level voice keyboards).

### Which audience does this matter to most?

Both

### Extra context

**User signal:** On Reddit, AFK Dev (remote desktop for AI coding agents with voice input + notifications) and similar tools show real demand for monitoring agents away from the desk. As PawWork adds more automation/scheduling capabilities, the need to check on and interact with results remotely becomes more natural.

[...]

PR #1454, opened 2026-06-20
refactor(remote-bridge): incremental channel add/remove (stop restarting every channel)

### LINKED ISSUE #1414: [Task] Wave 2: incremental channel add/remove (stop restarting every channel on connect/disconnect)
## Goal

A connect or disconnect of one remote channel never restarts the others: the shared PawWork event stream stays up, and only the affected channel starts or stops. The UI flap-suppression added in #1404 then becomes unnecessary and is removed.

## Scope

In scope:
- `supervisor.ts`: per-platform `AbortController`; allow adding/removing a supervised platform on a running supervisor (a registry, not a fixed `Promise.all` over a fixed array).
- `App` (`gateway.ts`): `addPlatform(platform)` / `removePlatform(name)` that register/unregister with the `Engine` and start/stop just that platform's loop, leaving the event stream and other platforms running.
- `RemoteBridgeRuntime` (`remote-bridge.ts`): `confirmPairing` / `disconnect` call `addPlatform` / `removePlatform` instead of the full `startBridge` rebuild. Keep the full rebuild only for cold start and a fatal-stream restart.
- Remove the flap-suppression (the `startBridge` pre-mark skip and the `onStatus` connected→connecting guard) once unaffected channels genuinely stay up.

Out of scope:
- WeChat session auto-relogin (separate follow-up).
- Any platform protocol change.

## Relevant files or context

- `packages/remote-bridge/src/supervisor.ts` — lines 10-13 document the Wave 1 whole-bridge-restart decision (no per-platform AbortController).
- `packages/remote-bridge/src/gateway.ts` — `createApp` / `App` fix the platform set at construction.
- `packages/desktop-electron/src/main/remote-bridge.ts` — `startBridge` / `confirmPairing` / `disconnect`; the flap-suppression.
- Split out of #1404 (raised in review as "delete the rebuild-flap suppression"; the correct resolution is this root fix, not deleting the interim suppression before its replacement exists). Part of #1188.

## Verification

- New per-channel tests: adding or removing one channel never changes another channel's reported status; the shared event stream stays connected across a single-channel add/remove; a fatal stream error still rebuilds all channels.
- Manual: with Telegram connected, connect/disconnect WeChat and confirm Telegram keeps serving (its poll is never aborted).

## Execution mode

Investigate and propose a plan first — post the plan as an issue comment and wait for an explicit "approved" comment before writing code or opening a PR.

Astro-Han:
## Summary

Connecting or disconnecting one remote channel no longer restarts the shared PawWork event stream or the other channels — only the affected channel starts or stops. Per-channel lifecycle becomes a first-class operation across three layers, and the UI flap-suppression that #1404 added as an interim is removed now that unaffected channels genuinely stay up.

- **`remote-bridge/supervisor.ts`** — a `PlatformSupervisor` registry replaces the one-shot `Promise.all`. Each platform runs under its own child `AbortController` linked to the run signal; `add()` / `remove()` / `stopAll()` start, stop, and await a single platform without touching the others. Each entry carries a generation token so a retired or replaced loop's late status can't clobber a newer entry under the same name. A thin `supervisePlatforms()` wrapper is kept for fixed-set callers.
- **`remote-bridge/gateway.ts`** — `App` is now a platform registry holding the injected factory and a `desiredPlatforms` source-of-truth map. `addPlatform(config)` runs the same `hasRemoteAudience` gate as cold start (so an incrementally-added channel can't bypass it), builds via the factory, registers with the Engine, and supervises it; a same-name re-pair retires the old loop but keeps its pointers. `removePlatform(name)` retires the loop and prunes that platform's session pointers. `run()` seeds the supervisor synchronously after hydrate and freezes the cold-start `onReady` set to that snapshot. The message handler drops an inbound from a platform that is no longer the live instance.
- **`remote-bridge/engine.ts`** — `unregisterPlatform(name)` drops the platform from the reconstruct index and discards active deliveries targeting it.
- **`remote-bridge/session-pointers.ts`** — `clearPlatform(name)` prunes one platform's mappings by remote-key prefix, keeping the event cursor and other platforms. On a successful write this lets a disconnect→reconnect of the same platform start fresh. The cleanup is best-effort: clearPlatform prunes in memory and then writes atomically, so a write failure leaves the on-disk snapshot intact and a stale pointer can revive after an immediate restart — an accepted behavior pinned by a `session-pointers` test, with a self-healing tombstone/retry left as follow-up.
- **`desktop-electron/remote-bridge.ts`** — `confirmPairing` / `disconnect` go incremental on a live bridge; a full `startBridge()` is kept only for cold start, last-channel teardown, and recovery after a fatal stream. The live bridge is one `{ app, ac, runPromise }` handle cleared whenever `run()` settles, so a post-fatal `confirmPairing` rebuilds instead of adding onto a dead app. Both flap-suppressions (the `startBridge` pre-mark skip and the `onStatus` connected→connecting guard) are deleted.

## Why

The whole-bridge restart on every connect/disconnect was the root cause of the UI flap that #1404 suppressed at the UI layer. Suppression only hid the blink; it could not fix the real cost — restarting the shared event stream and re-draining every channel whenever one is added or removed (a window in which messages can be missed). This is the root fix #1404's review deferred to Wave 2: don't restart the unaffected channels, and the suppression becomes unnecessary.

A second-AI design pass (first-principles review) surfaced three correctness points folded in here: the audience gate must run on incremental add (kept in the gateway, not bypassable); a removed/replaced platform's in-flight inbound must be dropped (gateway message-handler liveness guard) and its routing/active delivery dropped (`unregisterPlatform`); and a single-platform disconnect prunes that platform's session pointers so a same-platform reconnect normally starts fresh (best-effort: a failed pointer write can revive a stale pointer after a restart, pinned by tests).

## Related Issue

Closes #1414. Part of #1188. Follow-up sibling: #1426 (WeChat session auto-relogin) is intentionally out of scope here.

## Human Review Status

Pending

## Review Focus

- **`PlatformSupervisor` lifecycle correctness** (`supervisor.ts`): the entry is registered before the loop starts (so the first synchronous `starting` passes the token guard) and `done` is filled on the same tick (so `remove`/`stopAll` always await the real loop); the generation token drops a retired loop's late callbacks.
- **The live-handle clearing on `run()` settle** (`desktop remote-bridge.ts`): this is what makes "is a bridge live?" reliable, so a fatal stream rebuilds rather than adds onto a dead app.
- **The audience gate staying inside `App.addPlatform`** (`gateway.ts`): an incrementally-added channel goes through the same `hasRemoteAudience` check as cold start.
- **`addPlatform` vs `removePlatform` pointer semantics**: re-pair (addPlatform replacing a live same-name channel) keeps session pointers; disconnect (removePlatform) prunes them.

## Risk Notes

- **Deletion / credentials behavior**: `removePlatform` now prunes a single platform's session pointers via `clearPlatform` (orphaned parent links are left as-is; no remaining remote key references them). The last-channel teardown (stop bridge → `credentials.clear()` → `rmSync(statePath)`) is unchanged.
- **Platform/packaging surface**: none touched (no packaging, updater, signing, paths, shell, or permissions changes) — the change is OS-agnostic Electron main-process async logic, identical on macOS and Windows. The macOS/Windows checklist item is left unticked for that reason.
- **No visible UI or copy change**: the channel rows and status pills render exactly as before; only the underlying restart behavior changed. The UI/copy checklist item is left unticked for that reason.
- **Manual real-app regression deferred (needs a live WeChat scan)**: `bun run dev:desktop` walk — with Telegram connected, connect/disconnect WeChat and confirm Telegram keeps serving (its poll is never aborted) — is not yet run, as it requires a live WeChat pairing, same constraint as #1404. The behavior is covered by unit tests asserting the shared stream is never rebuilt and the survivor is never stopped.

## How To Verify

```text
remote-bridge typecheck (tsgo --noEmit): exit 0
remote-bridge unit (bun test ./src): 174 pass, 0 fail (12 files)
  - supervisor: add on running supervisor; remove stops only target; retired loop's late ready dropped; no abort-listener accumulation
  - engine: unregisterPlatform drops active delivery + reconstruct routing
  - gateway: addPlatform keeps shared stream up (one /global/event connect); removePlatform stops only target; wildcard-audience add refused; stale-instance inbound dropped
  - session-pointers: clearPlatform prunes one prefix, keeps cursor + other platforms (memory + on-disk)
desktop-electron typecheck (tsgo -b): exit 0
desktop-electron unit (bun test ./src): 375 pass, 0 fail (55 files)
  - two channels: connect adds incrementally (build=1, events=[add:wechat]); disconnect removes only one (build=1)
  - connecting a second channel leaves the first's stream + status untouched (no flap)
  - fatal stream degrades all + clears the handle, so the next connect rebuilds (build=2)
eslint (changed in-scope file desktop-electron/src/main/remote-bridge.ts): 0 errors
```

## Screenshots or Recordings

No visible UI change (main-process / gateway logic only).

## Checklist

https://claude.ai/code/session_01WVUVErxQ1AEf2mYT2bJ5hT

## Release Notes
* **New Features**
  * Incremental platform connect/disconnect while the bridge stays online, including prepare-first replacements.
  * Per-platform supervision with lifecycle phases (starting/serving/degraded) and more detailed serving status.
  * Runtime support for unregistering/removing a platform and keeping the rest online.
* **Bug Fixes**
  * Dropped stale inbound messages from removed platforms and prevented routing to disconnected targets.
  * Improved teardown/shutdown safety, including retry behavior after transient disconnect failures.
  * Isolated credential save failures so incremental joins/adds/removes don’t overwrite persisted credentials.

### DIFF
--- a/packages/desktop-electron/src/main/remote-bridge.test.ts
+++ b/packages/desktop-electron/src/main/remote-bridge.test.ts
@@ -14,7 +14,11 @@ import {
   type RemotePlatform,
 } from "./remote-bridge"
 
-function memoryStore(initial: RemoteAccount[] = [], available = true): CredentialStore & { value: RemoteAccount[] } {
+function memoryStore(
+  initial: RemoteAccount[] = [],
+  available = true,
+  failSave = false,
+): CredentialStore & { value: RemoteAccount[] } {
   return {
     value: initial,
     isAvailable() {
@@ -24,6 +28,10 @@ function memoryStore(initial: RemoteAccount[] = [], available = true): Credentia
       return this.value
     },
     save(accounts) {
+      // The real store throws when secure storage is unavailable or the file write
+      // fails. `failSave` models a save that fails even though pairing's upfront
+      // availability check passed — the keyring locks or the disk errors mid-confirm.
+      if (failSave) throw new Error("secure storage is unavailable on this system, cannot save the connection")
       this.value = accounts
     },
     clear() {
@@ -85,7 +93,14 @@ function servingApp(names: string[]): BridgeApp {
       for (const name of names) onStatus?.({ name, phase: "serving" })
       return hangUntilAbort(signal)
     },
-    { addPlatform: async (config) => emitStatus?.({ name: config.name, phase: "serving" }) },
+    {
+      addPlatform: async (config, beforeCommit) => {
+        // Mirror the real gateway: run the commit hook (the runtime saves its
+        // credential here) after the build, before the channel starts serving.
+        await beforeCommit?.()
+        emitStatus?.({ name: config.name, phase: "serving" })
+      },
+    },
   )
 }
 
@@ -109,7 +124,10 @@ function bridgeHarness() {
         return hangUntilAbort(signal)
       },
       {
-        addPlatform: async (added) => {
+        addPlatform: async (added, beforeCommit) => {
+          // Mirror the real gateway: the commit hook (credential save) runs after the
+          // build and before the live swap, so a hook that throws records no add.
+          await beforeCommit?.()
           state.events.push(`add:${added.name}`)
           emitStatus?.({ name: added.name, phase: "serving" })
         },
@@ -576,6 +594,72 @@ test("a failed re-pair on a live bridge keeps the working channel connected and 
   expect(build).toBe(1) // ...but the shared bridge was never rebuilt
   expect(runtime.getStatus().channels[0].state).toBe("connected") // working channel preserved
   expect(store.value).toEqual([oldAccount]) // old credential not overwritten by the new one
+  await runtime.stop()
+})
+
+test("a re-pair whose credential save fails keeps the working channel and never half-commits", async () => {
+  // Prepare-first commit hook: addPlatform persists the credential BEFORE swapping out
+  // the old loop, so a save failure (the keyring locks, the disk errors) aborts the
+  // swap with the old channel still serving — never a live new channel backed by a
+  // stale stored credential, and never an in-memory account list out of sync with disk.
+  const oldAccount = sampleAccount("telegram")
+  const store = memoryStore([oldAccount], true, true) // available so pairing proceeds; save throws
+  const harness = bridgeHarness()
+  const newAccount: RemoteAccount = { platform: "telegram", token: "999:xyz", allowFrom: "99", userName: "new" }
+  const runtime = new RemoteBridgeRuntime(
+    deps({
+      credentials: store,
+      buildApp: harness.buildApp,
+      pairers: [
+        fakePairer("telegram", {
+          pair: async (_start, emit) => {
+            emit({ phase: "awaitingBind", platform: "telegram", hint: "message" })
+            return newAccount
+          },
+        }),
+      ],
+    }),
+  )
+  await runtime.startIfConfigured() // cold start with the old account → connected
+  const channel = () => runtime.getStatus().channels.find((c) => c.platform === "telegram")
+  expect(channel()?.state).toBe("connected")
+  expect(harness.state.build).toBe(1)
+
+  await runtime.startPairing("telegram", { token: "x" })
+  await expect(runtime.confirmPairing("telegram")).rejects.toThrow(/secure storage/)
+
+  // No incremental add was committed, no rebuild, exactly one channel still serving,
+  // and neither memory nor disk took the new account.
+  expect(harness.state.build).toBe(1)
+  expect(harness.state.events).toEqual([])
+  expect(runtime.getStatus().channels).toHaveLength(1)
+  expect(channel()?.state).toBe("connected")
+  expect(store.value).toEqual([oldAccount])
+  await runtime.stop()
+})
+
+test("a second channel whose credential save fails is not added and leaves the first connected", async () => {
+  // Same commit-before-swap guard on the incremental-add path: a save failure while
+  // connecting a SECOND channel adds no orphan live channel and rolls the in-memory
+  // swap back, so the first channel keeps serving and the store holds only its account.
+  const store = memoryStore([sampleAccount("telegram")], true, true)
+  const harness = bridgeHarness()
+  const runtime = new RemoteBridgeRuntime(
+    deps({ credentials: store, buildApp: harness.buildApp, pairers: [fakePairer("telegram"), fakePairer("wechat")] }),
+  )
+  const stateOf = (platform: RemotePlatform) => runtime.getStatus().channels.find((c) => c.platform === platform)?.state
+  await runtime.startIfConfigured() // telegram connected from the saved account
+  expect(stateOf("telegram")).toBe("connected")
+  expect(harness.state.build).toBe(1)
+
+  await runtime.startPairing("wechat")
+  await expect(runtime.confirmPairing("wechat")).rejects.toThrow(/secure storage/)
+
+  expect(harness.state.events).toEqual([]) // WeChat never joined the live set
+  expect(harness.state.build).toBe(1)
+  expect(stateOf("wechat")).toBeUndefined()
+  expect(stateOf("telegram")).toBe("connected")
+  expect(store.value).toEqual([sampleAccount("telegram")]) // store still holds only telegram
   await runtime.stop()
 })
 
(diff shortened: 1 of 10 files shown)


Astro-Han:
@coderabbitai review

Astro-Han:
Addressed both P2 items in `8ac29894e3`.

**P2 (stop during in-flight add/remove rebuilds the bridge mid-shutdown).** Confirmed: `stop()` aborts the live handle on its synchronous prefix, so an in-flight `addPlatform`/`removePlatform` throws `BridgeClosedError` after the commit, and both desktop catch blocks then rebuilt unconditionally — standing up a fresh stream during shutdown only to tear it down again. Fix: in each catch, skip the rebuild when `handle.ac.signal.aborted` (a user-initiated stop), since the credential / trimmed accounts are already persisted and the queued stop is taking over; the next launch builds from the persisted accounts. A fatal stream does **not** abort the handle, so it still rebuilds to recover. Two runtime tests park the op at a gate, issue `stop()`, then release — asserting `build === 1` (no rebuild). Both fail (`build === 2`) when the guard is removed.

**P2 (non-last disconnect leaves a half-state when pointer cleanup fails).** Confirmed: `removePlatform` did `await retirePlatform` then `await clearPlatform`, and a `clearPlatform` write failure threw past the retire — the desktop rethrew it and never reached `statusMap.delete`, so the UI kept showing a channel whose loop was already stopped and credential already deleted. Fix: the live retire is the disconnect's commit, so forgetting session pointers is best-effort — a failed pointer write is now logged and swallowed, not thrown. The desktop then reaches `statusMap.delete` and the UI is consistent; a stale on-disk pointer self-heals on the next pointer save. Gateway test asserts `removePlatform` resolves (and warns) when `clearPlatform` rejects; it fails when the swallow is removed.

Note on a residual: a *fatal stream that races a stop* within the in-flight window can still rebuild once during shutdown (the fatal path nulls `current` so `stop()`'s abort no-ops, leaving `handle.ac` un-aborted). It produces the same benign one-time flap and is much rarer than the deterministic stop-during-add case; covering it would need a dedicated `stopping` flag, which seemed like over-engineering for an identical outcome. Happy to add it if you'd prefer the seam fully closed.

Astro-Han:
Addressed both items in `042a7d5faf`.

**待验证 — does a failed pointer write revive a stale pointer on restart? Verified: yes.** Added a file-backed `SessionPointers` test for the write-failure restart chain. `clearPlatform` prunes the in-memory map and then writes atomically (unique temp + rename), so a write failure leaves memory pruned but the on-disk snapshot intact; a restart reloads it and — since `addPlatform` keeps pointers on reconnect — the stale mapping resurfaces. The test forces the failure with a read-only directory (skipped under root, where that can't block writes) and asserts the revival.

Decision: **accept best-effort for this PR.** The harm is bounded (a reconnected channel's first message may route to an old session; if it's gone the engine surfaces an error and the user resends, starting fresh — no data loss, no security impact) and the trigger is a rare conjunction (clear write fails ∧ no later save heals the file ∧ immediate exit ∧ later reconnect of the *same* platform). The clean durable fix can't be "a fresh connect clears pointers" — cold start is also a non-re-pair `addPlatform`, so that would wipe every conversation on every launch; the only correct seam is making the disconnect's forget durable (a tombstone / retry), which is more persistent state than this IO-failure tail warrants. The test pins the behavior as deliberate; happy to open a follow-up issue for a tombstone if you'd prefer belt-and-suspenders.

**P3 — trimmed the invariant comments.** Compressed the multi-paragraph commit/rebuild explanations in `addPlatform`, `removePlatform`, `confirmPairing`, and `disconnect` to one-line invariants now that tests cover the races, and corrected `removePlatform`'s comment which overclaimed the stale pointer was "never resurfaced". No logic change.

remote-bridge 181 tests / desktop-electron 586 / both typecheck / eslint in-scope all green.

Astro-Han:
Fixed in `66fd65761c`.

P1 confirmed: the revival test forced the write failure with POSIX `chmod 0o500`, which can't make a directory unwritable on Windows (chmod is a no-op there) or under root (bypasses permission bits), so `clearPlatform` wouldn't reject and the assertion would fail / pass implicitly. Moved the guard into the test declaration — `test.skipIf(process.platform === "win32" || process.getuid?.() === 0)(...)`, matching the existing owner-only-permissions test in the same file — and removed the in-body root early-return. Verified: on POSIX non-root the test runs and passes; forcing the condition true renders it as an explicit `1 skip`, not a silent pass. (For the record, `unit-remote-bridge` itself runs ubuntu-only, so this is portability/clarity hardening rather than a live CI break, but the explicit skip is the right form.)

remote-bridge 181 tests (1 skipped on Windows/root) / both typecheck / prettier clean.

Issue #1465, opened 2026-06-22
[Task] Rebuild diagnostics package export: privacy redaction, size budgets, in-app review


Astro-Han:
## Goal

Rebuild the "Prepare Diagnostics Package" export so the artifact a user uploads to the external feedback form: (a) does not leak secrets the user cannot see (API keys, tokens, absolute paths, usernames, emails), (b) is size-bounded with no oversized single block, (c) is reviewable in-app before upload, (d) is localized.

Background: a non-technical Windows user exported a 4.3MB problem report that leaked private data and was unreviewable. The full report's `logTail` and `sessionExport.messages` go in with zero redaction today, while `rendererDiagnostics` already uses a strict allowlist — the privacy treatment is badly asymmetric.

## Scope

Lands as 3–4 flat PRs from `dev` (sequential, not stacked):

- **PR1 — Structured privacy redaction** (security lens, first): session export via a structure-aware allowlist (role / time / part-type / tool-name / byte-size + per-part length-capped body + secret redaction, unknown fields omitted); `logTail` / `rendererError.details` get a thorough secret blacklist (`sk-`/`Bearer`/`AKIA`/`AIza`/`glpat-`/`hf_`/`npm_`/`rk_live_`/PEM/basic-auth + absolute paths + username + email). Keeps the existing JSON format.
- **PR2 — Component-level size budgets**: independent byte caps per source (log total / per-line, session total / per-message / per-part, `rendererError.details`, `rendererDiagnostics`); `tailFile` per-line + total-byte caps without reading the whole file into memory; widen the hardcoded `rendererDiagnostics` 5-minute slice window. The whole-report 5MB truncation becomes a final backstop only.
- **PR3 — Flow + contract + in-app review**: stop auto-copying the clipboard summary, keep a minimal redacted optional-copy fallback (`ReportProblemResult`/`summaryCopied` contract + error page); in-app review panel listing what's included + narrowed copy (no absolute "all secrets removed" claim) + localization.
- **PR4 (optional, last) — Structural cleanup**: drop the markdown prose wrapper, `logTail` as a line array, field-naming consistency.

Out of scope: merging / releasing; in-app submission / direct-API backend (the external form stays); the vendored agent engine (`packages/opencode/`); unrelated files.

## Relevant files or context

[...]

PR #1471, opened 2026-06-23
fix(desktop): bound problem report size with per-component byte budgets

Astro-Han:
## Summary

Bound the "Prepare Diagnostics Package" export with **per-component byte budgets** — PR2 of the diagnostics-package rebuild (#1465). Each component is capped to its own budget **after redaction and before** the overall 5 MB truncation ladder (which becomes a fallback):

- **Per-component budgets** — fixed and private (not a configurable API; no production caller tunes them): `logTailBytes` 1 MB, `sessionMessagesBytes` 1 MB, `sessionMessageBytes` 256 KB, `rendererErrorDetailsBytes` 64 KB (bounds both the renderer-error **summary** and **details**), `rendererDiagnosticsBytes` 1 MB. The cap helpers (`capLogTailBytes` / `capMessageParts` / `capSessionMessagesBytes` / `headBytes` / `capEvents`) are exported and unit-tested directly at small limits, with a few default-budget tests covering the end-to-end wiring.
- **Hard caps, no force-keep:** `capSessionMessagesBytes` drops even the newest message if it does not fit, so the cap is a true ceiling; `capMessageParts` keeps the **latest** parts of an oversized turn (the tool output / error nearest the failure), matching the ladder's oldest-first drop order.
- **Renderer diagnostics re-budgeted after redaction:** the slice is byte-capped at the source, but redaction here can re-expand it, so it is re-bounded with the same `capEvents()` ruler. The component pass and the overall fallback share that one mechanism (incident events dropped only when a tight limit leaves no alternative), replacing a divergent bespoke fallback loop that could leave an all-protected oversized slice unbounded.
- **UTF-8-safe byte slicing** (`headBytes` / `tailBytes`) so capping never splits a multi-byte codepoint at the budget boundary.
- **Consistent truncation ledger:** every omitted-bytes counter is tracked identically in the component pass and the overall fallback ladder, validated by `isTruncation`.
- **Structure-agnostic private-key redaction** (`problem-report-redact.ts`): defense moved out of the truncation layer into the redactor as anchored, linear-time (ReDoS-safe) rules covering complete `BEGIN..END` blocks, `BEGIN`-truncated keys, markerless ≥2-line base64 bodies, orphaned `END` markers (separate line, same line, PGP `=CRC` checksum line), armor-headered bodies, and CRLF.

## Why

The export could grow without bound: a single huge log tail, one giant session message, or an oversized renderer-error blob could dominate the report or trip the overall 5 MB cap in a way that silently drops other components. Per-component budgets make each part's ceiling explicit and keep the report balanced, with the overall ladder only as a fallback. The budget/truncation work also surfaced a real redaction gap: a pre-redaction tail cut could strand a private-key body whose `BEGIN` marker was removed, defeating a `BEGIN`-keyed redactor — so private-key defense now lives entirely in the redactor.

## Related Issue

Refs #1465 (umbrella). This is PR2; PR1 #1470 is merged, PR3 #1472 follows. Rebased flat onto `dev` after #1470 merged.

## Human Review Status

`Pending`

## Review Focus

- The **redact-before-truncate invariant**: redaction runs on the full payload before any component capping; the only pre-redaction truncation is `tailFile` (byte + line tail of the log), which now only drops a partial first line — multi-line-secret defense lives entirely in the redactor.
- Hard caps are true ceilings (drop the newest message if it does not fit), and the omitted-bytes ledger stays consistent between the component pass and the fallback ladder.
- Renderer diagnostics now go through one `capEvents()` ruler in both the component pass and the overall fallback (re-derived from the redacted baseline so the omitted ledger stays correct), instead of a separate fallback loop that broke on protected events.
- UTF-8-safe byte slicing at budget boundaries.
- The structure-agnostic private-key redactor rules (anchored, linear-time) and whether the enumerated marker/armor/CRLF cases are complete.

## Risk Notes

Accepted residuals (confirmed by an independent adversarial review; left as documented tradeoffs rather than risking the ReDoS-safety / false-positive balance):

- A private-key body wrapped at **non-standard narrow widths (< 16 chars/line)** whose `BEGIN` marker was truncated away keeps the same `{16,}` per-line floor used everywhere else (the floor avoids redacting legitimate short base64 tokens/hashes). Real tools emit 64–76 char lines, so this only affects deliberately re-wrapped keys. PR3's human-review step is the backstop.
- The overall fallback ladder still uses UTF-16 `slice()` for log / renderer-error text. It is effectively unreachable under default budgets (component budgets sum well under 5 MB), so a lone-surrogate split there is cosmetic.
- **No visible UI in this PR** (the in-app review panel is PR3 #1472) — the UI/copy checklist item below is left unticked for that reason; no screenshots.

## How To Verify

```text
bun test (packages/desktop-electron) — 656 pass, 0 fail (full suite, latest head)
bun run typecheck (packages/desktop-electron, tsgo -b) — clean
eslint (changed files) — clean
Targeted coverage: pure-helper edge cases (each cap helper at small limits), default-budget
  end-to-end wiring, hard-cap ceilings, ledger consistency, renderer-diagnostics re-budget +
  all-protected drain, UTF-8 boundary slicing, and ReDoS-perf regression on the private-key redactor.
```

## Screenshots or Recordings

None — no visible UI in this PR (the in-app review panel is PR3 #1472).

## Checklist

> **How to use this checklist:**
>
> - Tick a box by replacing `[ ]` with `[x]`. Do not edit, add, or remove items.
> - The bot-applied label items can only be honestly ticked AFTER the PR is opened and the labeler / priority-triage bots have run — return to the PR description and tick them then.
> - Most items are required. The few that are conditional are explicitly marked **(conditional)**; for those, leave unticked if they truly do not apply and explain why in Risk Notes. All other items must be ticked before requesting human review.

- [x] **(conditional)** I considered macOS and Windows impact for platform, packaging, updater, signing, paths, shell, or permissions changes. Leave unticked only if no platform/packaging surface was touched.

### DIFF
--- a/packages/desktop-electron/src/main/logging.test.ts
+++ b/packages/desktop-electron/src/main/logging.test.ts
@@ -141,4 +141,42 @@ describe("desktop logging", () => {
     expect(tail).toContain("Backend log")
     expect(tail).toContain("backend failed")
   })
+
+  test("bounds the tail to the most recent bytes of a large file", async () => {
+    logDir = mkdtempSync(join(tmpdir(), "pawwork-logging-test-"))
+    const mainPath = join(logDir, "main.log")
+    // Larger than the 512 KB byte cap so the oldest content falls outside the tail.
+    const filler = Array.from({ length: 60_000 }, (_, i) => `entry ${i}`).join("\n")
+    writeFileSync(mainPath, ["OLDEST_LINE", filler, "NEWEST_LINE"].join("\n"))
+    fakeLog.transports.file = {
+      maxSize: 0,
+      getFile: () => ({ path: mainPath }),
+    }
+
+    const { diagnosticsLogTail } = await import(`./logging?logging-test=${crypto.randomUUID()}`)
+    const tail = diagnosticsLogTail({})
+
+    expect(tail).toContain("NEWEST_LINE")
+    expect(tail).not.toContain("OLDEST_LINE")
+    // The main-log section is bounded by the byte-tail read (plus the section header), not the file.
+    expect(Buffer.byteLength(tail, "utf8")).toBeLessThanOrEqual(512 * 1024 + 4_096)
+  })
+
+  test("drops a mid-line fragment instead of emitting it before redaction", async () => {
+    logDir = mkdtempSync(join(tmpdir(), "pawwork-logging-test-"))
+    const mainPath = join(logDir, "main.log")
+    // A single line larger than the 512 KB byte cap with no newline: the byte tail starts mid-line.
+    // Emitting that fragment would feed redaction a prefix-less secret, so it must be dropped.
+    writeFileSync(mainPath, "S".repeat(600_000))
+    fakeLog.transports.file = {
+      maxSize: 0,
+      getFile: () => ({ path: mainPath }),
+    }
+
+    const { diagnosticsLogTail } = await import(`./logging?logging-test=${crypto.randomUUID()}`)
+    const tail = diagnosticsLogTail({})
+
+    expect(tail).toContain("(empty)")
+    expect(tail).not.toContain("S".repeat(100))
+  })
 })
--- a/packages/desktop-electron/src/main/logging.ts
+++ b/packages/desktop-electron/src/main/logging.ts
@@ -1,9 +1,14 @@
 import log from "electron-log/main.js"
-import { readFileSync, readdirSync, statSync, unlinkSync } from "node:fs"
+import { closeSync, fstatSync, openSync, readdirSync, readSync, statSync, unlinkSync } from "node:fs"
 import { dirname, join } from "node:path"
 
 const MAX_LOG_AGE_DAYS = 7
 const TAIL_LINES = 1000
+// Read at most this many bytes from the END of each log file. The main log rotates at 5 MB (maxSize
+// below), but the backend log path is external and unbounded — without this a multi-GB log would
+// load fully into memory just to take the tail. This also bounds any single line, so no separate
+// per-line cap is needed (and a pre-redaction per-line cut would risk slicing a secret's prefix off).
+const TAIL_MAX_BYTES = 512 * 1024
 const CONSOLE_TRANSPORT_INITIALIZED = Symbol.for("pawwork.consoleTransportInitialized")
 
 export function initLogging() {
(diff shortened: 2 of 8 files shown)


Astro-Han:
Addressed the review in place (no split) — pushed as `86bf454`.

### Changes

- **P2 — `capMessageParts` kept the *earliest* parts.** Flipped to keep the **latest** parts (`parts.slice(parts.length - kept)` in both the binary search and the final trim). The tail of a turn is the tool output / error nearest the failure, and it now matches the rest of the ladder, which drops oldest-first to keep the most recent context.

- **P3 — `Options.budgets` was a test-only production API.** Removed it, along with `DEFAULT_COMPONENT_BUDGETS`, the `ComponentBudgets` type, and `resolveBudgets()`. Budgets are now a private fixed `COMPONENT_BUDGETS` const. The small-budget edge cases that previously needed an override are now tested by calling the pure cap helpers directly with tiny limits — `headBytes` / `capLogTailBytes` / `capMessageParts` / `capSessionMessagesBytes` are exported for this. A lean set of integration tests still proves the helpers are wired into `buildProblemReport` under the real default budgets (1 MB log / 1 MB session / 256 KB per message / 64 KB per renderer-error field).

- **P2 — scoped renderer slice had no lower time bound.** Removed the `NEGATIVE_INFINITY` special case; scoped (session/trace) slices now use the same bounded ~30-min lookback as unscoped ones, so a manual trigger collects minimally by default and a session's hours-old event is not pulled in. An explicit `from` still opens a wider window when a reviewer needs older history. The slice test now asserts both: bounded by default, widened with `from`.

### On splitting out the private-key redactor (P3)

Kept as one PR. The private-key body-matching logic lives in `problem-report-redact.ts` and is part of the same redaction-before-budgeting pipeline this PR bounds; it is small, already has focused unit tests in `problem-report-redact.test.ts`, and splitting it would create a stacked PR with a trivial-but-real `feedback.ts` overlap for no review benefit. The diff here is scoped to size-budgeting + the redaction it depends on.

### Verification

```text
bun run typecheck      — clean (tsgo -b, 9/9 tasks)
eslint (changed src)   — 0 errors (test files are eslint-ignored by config)
bun test (electron)    — 654 pass, 0 fail, 1989 expect() calls, 72 files
  problem-report + renderer-diagnostics-slice — 47 pass, 0 fail
```

Astro-Han:
Addressed both items — pushed as `baa61f6`.

### P2 — renderer diagnostics re-budgeted after redaction (implemented)

Verified the gap: the slice is byte-capped only at the source; `buildProblemReport` then `redactRendererDiagnostics()`, which can re-expand it, and the component-budget pass skipped it (`"not re-budgeted here"`). The overall fallback used a bespoke loop that `findIndex(!protected)` → `break` on the first protected event, so an all-protected oversized slice could ride past `maxBytes` and the report would throw.

Fix, with `capEvents()` as the single ruler:
- Added `rendererDiagnosticsBytes` (= `SESSION_EXPORT_RENDERER_DIAGNOSTICS_MAX_BYTES`, 1 MB) to the fixed component budgets.
- Re-bound the slice **post-redaction** with `capEvents()`, re-derived from an immutable redacted baseline so the omitted-bytes ledger stays correct across the component pass and the overall ladder.
- Replaced the bespoke fallback loop with a `capEvents()`-based shrink that drops even incident events once a tight `maxBytes` leaves no alternative, so the slice drains to empty instead of overflowing. Removed the now-dead `isProtectedRendererDiagnosticEvent` duplicate predicate (its rule already lives inside `capEvents`).
- Tests: (a) all-protected oversized slice drains under a 5 KB `maxBytes` — no throw, report ≤ maxBytes, ledger > 0; (b) 1.2 MB of events under the default 5 MB limit are re-bounded to the 1 MB component budget (only the component cap, not the overall ladder, could have trimmed them).

### P3 — PR description aligned to current head (done)

Rewrote the description: removed `resolveBudgets()` / budgets-override / NaN-budget references; it now states fixed private component budgets (incl. `rendererDiagnosticsBytes`), the exported pure cap helpers tested at small limits, the default-budget end-to-end wiring, and the renderer-diagnostics re-budget above.

### Verification

```text
bun run typecheck      — clean (tsgo -b)
eslint (changed src)   — 0 errors
bun test (electron)    — 656 pass, 0 fail, 72 files
  problem-report.test  — 45 pass, 0 fail (incl. the 2 new renderer-diagnostics tests)
```

Issue #1478, opened 2026-06-28
[Bug]  Update deleted the custom imported skills.


asaprajx:
### What happened?

When I updated PawWork to v2026.6.12, it deleted all the skills that I added to the folder where i installled PawWork ```E:\Program Files\PawWork\resources\skills```by  making subfolders and adding skill.md files in them. All of the skills were functional, and I deleted the zip files after importing them to PawWork. Now i will have to download all of them again :(

I tested the imported skills when I first added them in PawWork:

<img width="909" height="299" alt="Image" src="https://github.com/user-attachments/assets/295eb192-6166-47cb-960e-d85bb7ccee20" />

But they're gone now, and only the default skills are there:

<img width="746" height="536" alt="Image" src="https://github.com/user-attachments/assets/45f41784-8609-42d3-8fe6-7d8fa0040e4d" />

### Which area seems affected?

Platform shell, packaging, updater, signing, paths, or permissions

### How much does this affect you?

Breaks an important workflow

### Steps to reproduce

1. Add a skill from GitHub to the folder where you have installed PawWork: ```C:\...\PawWork\resources\skills```
2. Check the skills section to see that the skill is added to PawWork.
3. update to a newer version of PawWork.
4. See the error; the imported skills are gone.

### What did you expect to happen?

I expect the skills that are added by users to stay in the skills folder of PawWork.

### PawWork version

2026.6.12

### OS version

Windows 11

### Can you reproduce it again?

Only once so far

Astro-Han:
Hi @asapjayz, thanks for the detailed report, and I'm sorry this happened. You set those skills up, tested them, deleted the zips, and the update wiped them anyway. That's frustrating, and it shouldn't have cost you that work. I want to help you get them working again.

Short version of the cause: `resources\skills` sits inside PawWork's install directory, and the Windows updater replaces that whole directory on every update. Anything you add there gets cleared and replaced with the built-in defaults. It was never a safe place for your own skills, but nothing in the app told you that, and that's on me.

Workaround you can use today:

Put your custom skills here instead:
`%USERPROFILE%\.agents\skills\<skill-name>\SKILL.md`

That folder lives in your home directory, outside the install, so updates leave it alone. PawWork already reads skills from there by default, with no setting to change, and they show up across all your projects.

Windows tip: if Explorer won't let you create a folder named `.agents`, paste `%USERPROFILE%\.agents\skills` into the address bar, or create it from a terminal.

[...]

PR #1482, opened 2026-06-29
feat(skills): add Open skills folder action to the Skills surface

Astro-Han:
## Summary

Adds a desktop-only **Open skills folder** button to the Skills surface that opens `~/.agents/skills`, the update-safe home for user-added skills, creating it on demand.

- `/path` API gains a `skills` field plus an `ensureSkills` query flag (mirrors `ensureConfig`); `Skill.userSkillsDir()` single-sources the path next to `builtinRoots`. SDK regenerated.
- New `OpenSkillsFolderButton` (extracted for isolated testing) gated on `canOpenLocalPath`, so the web build hides it.
- Skills header relaid out: title and actions share one vertically-centered row, with the description on its own line below.

## Why

Custom skills dropped into the install bundle's `resources/skills` are wiped on every update, because the Windows NSIS updater replaces the whole install directory (`RMDir /r $INSTDIR`). There was no in-app way to add a skill and no signposted safe location, so users found the only visible skills folder, which happens to be the volatile one, and lost their work silently and irreversibly on update.

`~/.agents/skills` lives under the home dir (outside the install bundle) and is already scanned by skill discovery, so it survives updates. This change makes that location discoverable from the app.

## Related Issue

#1478

## Human Review Status

Pending

## Review Focus

- The `/path` contract change and the regenerated SDK diff (`skills` is a required field now; the fixed default-path stores were updated to carry it).
- The desktop gate: the button must never appear in the web build (no local filesystem to open).

## Risk Notes

- Scope is deliberately the minimal fix: a discoverable, update-safe place to add skills. It does **not** migrate existing `resources/skills` content or warn when a custom skill is still found there; both are tracked as follow-ups. Residual risk: a user who keeps using `resources/skills` directly still loses content on update, but with a signposted button the path of least resistance now leads away from the trap.
- Platform: the action is desktop-only (`platform.openPath`); the path resolves to `~/.agents/skills` on both macOS and Windows. No packaging/updater/signing code is touched.
- A user-facing release note would be worth adding when this ships; not included here.

## How To Verify

```text
opencode typecheck (tsgo --noEmit): pass
app typecheck (tsgo -b): pass
sdk/js typecheck (tsgo --noEmit): pass
eslint (changed app files): 0 errors
unit  src/pages/skills/open-skills-folder.test.ts: 2 pass (ensures+opens path; no-op when path absent)
component  src/pages/skills/skills-folder-button.test.ts: pass (button shown only when openPath present; click -> path.get({ensureSkills:true}) -> openPath)
e2e  e2e/skills/skills-panel.spec.ts: 4 pass (incl. new assertion that the web surface hides the button)
SDK regen: openapi.json + v2 gen carry `skills` (response) and `ensureSkills` (query); unrelated models-snapshot regen reverted
```

Visual: walked the Skills route in `bun run dev:desktop` (zh locale) confirming the button renders and the header reads calmly; `bun run snap skills-surface` confirms the web header layout and that the gate hides the button.

## Screenshots or Recordings

image

## Checklist

- [x] **Routing labels** — this PR carries at least one of `app`, `ui`, `platform`, `harness`, `ci`. The labeler bot assigned `app` (packages/app), `ui` (**/*.tsx per labeler.yml), and `harness` (packages/opencode + packages/sdk); all correct for the changed paths.
- [x] **Priority label** — this PR carries exactly one of `P0`, `P1`, `P2`, `P3`. The triage bot assigned `P2`; confirmed (high-severity but low-frequency, and this is the discoverability fix rather than the wipe itself).
- [x] **(conditional)** I manually checked visible UI or copy changes when needed, with screenshots or recordings. Leave unticked only if no visible UI or copy changed.
- [x] **(conditional)** I considered macOS and Windows impact for platform, packaging, updater, signing, paths, shell, or permissions changes. Leave unticked only if no platform/packaging surface was touched.

## Summary by CodeRabbit

* **New Features**
  * Added an “Open folder” action in the Skills area for supported desktop apps.
  * Skills data now includes a dedicated folder path, enabling the app to open the correct location directly.
  * Added localized wording for the new Skills folder action and error message.

* **Bug Fixes**
  * Improved handling so the Skills folder is only shown/opened when available, with clearer fallback behavior if it can’t be opened.


Astro-Han:
Addressed all three CodeRabbit findings:

1. **Search field a11y** (`skills-surface.tsx`) — added `aria-label` reusing the placeholder copy so the icon-only label no longer leaves the textbox unnamed. (5191af3)
2. **Test determinism** (`skills-folder-button.test.ts`) — replaced the fixed 20ms sleep with a promise the `openPath` mock resolves, so the assertion waits on the real signal instead of a timer. (135d67f)
3. **Raw `fs.mkdir` in the Effect handler** (`handlers/root.ts`) — converted both the config and skills mkdirs in `getPaths` to `AppFileSystem.Service.makeDirectory` and provided `AppFileSystem.defaultLayer` to the production router runtime. Note: the codebase's FS abstraction is the `AppFileSystem` wrapper (`@opencode-ai/core/filesystem`), not the bare `FileSystem.FileSystem`, which is why this also adds the layer; `Effect.orDie` keeps the prior fail-fast behavior. (4d7876a)

Verification: `opencode` + `app` typecheck pass; `packages/app` skills tests (3) pass; `test/server/instance-root-routes.test.ts` (6) pass — exercises `/path` through the real route runtime including the newly provided `AppFileSystem` layer; eslint 0 errors on the changed app source.

PR #1499, opened 2026-07-09
feat(rust): PR2 mock LlmClient + tool loop + permissions with interruption

Astro-Han:
## Summary

Add the `pawwork-core` turn loop (M0, experimental Rust line): a generic `Agent` that drives one user turn over an `LlmClient`, a tool runtime, and a permission gate, appending every step to the JSONL ledger and mirroring it onto a wire event stream. Core only — no CLI interaction (that is a later PR). Zero compile dependency on `packages/*`; does not ship in the desktop product.

- **llm** — `LlmClient` async trait (generic, RPITIT + `Send`) + a scripted `MockLlmClient` with a hang-until-cancel step. Half-parsed tool calls are unrepresentable by contract, so the loop has no "is this half a call?" branch.
- **tool** — `Tool` trait (`dyn` via a hand-boxed `Send` future, no `async-trait`) + `ToolRuntime`. A `prepare`/`run` split rejects a malformed, unknown, or out-of-workspace call *before* any `tool.started` and before `run()`. Read-only `read`/`list` with a workspace path fence (`canonicalize` + component-wise `Path::starts_with`; advisory, not a sandbox) and output caps.
- **permission** — `PermissionGate` async trait (generic) + a structured `PermissionRequest` (tool name + rendered summary). The typed action enum lands with `edit`/`shell`, its first real producers.
- **agent** — `run_turn`. Appends ledger events and emits `AgentEvent::Ledger` after each durable append (ledger is the source of truth; the channel is a live mirror). Interruption via `CancellationToken` at four checkpoints (model-round top, model call, permission wait, between tools); the `turn.interrupted` finalize is itself uninterruptible. Permission denial synthesizes a tool error and continues — it is not an interruption.
- **ci** — second commit `ci(labeler)`: routes the Rust `crates/` line (plus root `Cargo.toml`/`Cargo.lock`, `rust-toolchain.toml`) to the existing `harness` routing label in `.github/labeler.yml`. Rust-only PRs previously matched no routing glob, so pr-triage's label policy failed for every one of them; PR1 only passed by incidentally touching `.github/**`.

## Why

Step 2 of the Rust agent v0 plan (after PR1 #1497 scaffold + ledger). This lands the runnable loop skeleton against a mock model, so the OpenAI-compatible SSE client (PR3) and the interactive CLI (PR4) can slot into fixed `LlmClient` / `PermissionGate` / event-stream seams without reshaping the loop.

Key decisions, from a Fable + codex architecture review (recorded in the local design doc, section 13):
- **Runtime**: async on tokio, **multi-thread default flavor** — chosen to force `Send` discipline now for future subagents, not because `select!` needs it. Core stays runtime-agnostic (tokio `sync` feature only; the flavor is the binary's `#[tokio::main]`, deferred to the CLI PR).
- Auto-allowed read-only tools record **no** permission events; only real user decisions do.
- Chat history is built in-memory this turn; ledger→messages replay for cross-session resume is deferred to the resume/CLI PR.

## Related Issue

No dedicated issue — continues the experimental Rust line started in #1497.

## Human Review Status

Pending

## Review Focus

- The path fence (`tool/fence.rs`): component-wise containment, symlink and sibling-prefix escapes. This is the one security-relevant surface.
- Interruption correctness in `agent.rs`: the four checkpoints, and that the `turn.interrupted` finalize is not itself cancellable.
- The reject-before-start contract: unknown tool / invalid JSON / path escape must produce a `tool.finished(ok=false)` with **no** `tool.started`.

## Risk Notes

Experimental line; rust-ci is not a required check during M0 and the crates do not ship in the product. Two commits by concern: the core loop (Rust) and a `.github/labeler.yml` routing fix (CI config only, no behavior change). No macOS/Windows packaging surface touched (pure library + tests; the fence's symlink test is cfg-split for both). New dependencies: `tokio` (`sync` only in the lib) and `tokio-util` (CancellationToken) — both mainstream, justified inline. No visible UI or copy.

## How To Verify

```text
cargo test --workspace --locked: 41 passed (16 ledger + 25 new)
  - loop: plain text turn, channel mirrors ledger in order, read tool round-trips result into history
  - reject-before-start: unknown tool / invalid JSON / path escape -> tool.finished(false), no tool.started
  - permission: confirm allow (requested->decided->started->finished), deny (synthesized error, continues)
  - interruption: cancel before start (records nothing), mid-model, mid-permission-wait
  - llm error: records error then turn.interrupted
  - fence matrix: inside/root/absolute-out/dotdot/symlink-out/missing/sibling-prefix
cargo fmt --check: clean
cargo clippy --workspace --all-targets -- -D warnings: clean
CI: rust SUCCESS, pr-triage SUCCESS
```

## Screenshots or Recordings

N/A — no visible UI.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -34,3 +34,9 @@ harness:
           - "packages/sdk/**"
           - "packages/core/**"
           - "packages/remote-bridge/**"
+          # Rust agent line (crates/): same agent-engine layer, in Rust. Gives the
+          # Rust-only PRs a routing home so pr-triage's label policy passes.
+          - "crates/**"
+          - "Cargo.toml"
+          - "Cargo.lock"
+          - "rust-toolchain.toml"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72f5acc6cb2ba439de613abc23857ec3d78374d8ed5ac84e9d11336e87da8649"
 
 [[package]]
+name = "bytes"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -19,6 +25,12 @@ name = "futures-core"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7e3450815272ef58cec6d564423f6e755e25379b217b0bc688e295ba24df6b1d"
+
+[[package]]
+name = "futures-sink"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c39754e157331b013978ec91992bde1ac089843443c49cbc7f46150b0fad0893"
 
 [[package]]
 name = "futures-task"
@@ -97,6 +109,8 @@ version = "0.0.0"
 dependencies = [
  "serde",
  "serde_json",
+ "tokio",
+ "tokio-util",
  "uuid",
 ]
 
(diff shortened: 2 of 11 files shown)


Astro-Han:
ChatGPT (web, Code Review Project) reviewed this PR: **建议通过** with two P2s, no P0/P1. Both addressed in `fix(rust): bound the turn loop and the read/list tools` (e291c9024c):

- **P2 — turn loop had no round budget**: a model that keeps returning tool calls would loop forever, growing the ledger/context unbounded. Added a per-turn `max_model_rounds` guard (default 64); on exhaustion the turn records `error` + `turn.interrupted("budget exceeded")`. New test drives a mock that returns a tool call every round and asserts it stops after exactly N.
- **P2 — read/list capped output but not work**: `read` slurped the whole file into memory before truncating (a multi-GB workspace file could OOM) → now `File::open` + `Read::take(MAX+1)`. `list` collected an unbounded directory before truncating → now stops scanning at a 10k ceiling and marks the result truncated, keeping sorted-first-1000 correct for normal dirs.

The reviewer also noted `tool.run` is not wrapped in cancellation; that was a deliberate PR2 deferral (read/list are fast, and now bounded), with real mid-run cancellation landing alongside `shell`.

Verification after fixes: `cargo test --workspace --locked` 42 passed; `fmt --check` clean; `clippy --all-targets -D warnings` clean.

PR #1556, opened 2026-08-12
fix: [Bug] Toast message Spam: Session status hydration timeout after 5000ms

### LINKED ISSUE #1550: [Bug] Toast message Spam: Session status hydration timeout after 5000ms
### What happened?

After updating to version 2026.8.1, whenever I open PawWork, there's always a pop-up toast message saying `failed to reload pawwork`, and after every time an assistant completes its response, it spams the right status bar with a similar toast, `saying Failed to reload Workspace`. There are no functional bugs, but it feels annoying and it started after the updating to v2026.8.1.

Not sure

Minor issue or polish

### Steps to reproduce

1. Open Pawwork; see the first toast.
2. Let an assistant finish its task or interrupt in between the task execution; see the other toast message.

I didn't expect anything; I have a good PC with mid-high-end specs, so I think it's not that my PC isn't capable of running PawWork. I think PawWork should open and operate without the toasts as it used to be

### PawWork version

2026.8.1

### OS version

windows 11

Yes, every time

### Diagnostics

here are some screenshots regarding this issue,.

Image

Image

webbrain-one:
Closes #1550

## Summary by CodeRabbit

* **Tests**
  * Added a 30-second timeout to automated test commands to prevent tests from running indefinitely.


Astro-Han:
Hi @webbrain-one, thanks for the quick response to this report and for opening the PR - we appreciate the contribution.

After investigating, we found this change doesn't address the reported toast spam:

- The toasts come from production code in the packaged app: `bootstrapDirectory` in `packages/app/src/context/global-sync/bootstrap.ts` wraps the `session.status` fetch in a 5-second timeout (`SESSION_STATUS_TIMEOUT_MS`). A timeout currently rethrows into the slow-bootstrap error path, which fires the project-level "Failed to reload {project}" toast on every bootstrap pass - app launch and after each assistant turn.
- This PR only adds `--timeout 30000` to the `bun test` commands in `packages/app/package.json`. That changes the test runner's per-test timeout and has no effect on the shipped app.

The matching "5000ms" numbers are coincidental: bun test's default per-test timeout is also 5s, so a locally timing-out test looks similar to the runtime diagnostic, but the production timeout is a separate constant.

We're closing this PR and will land a fix on the bootstrap side instead: the timeout is already recorded as a diagnostic event and the status recovers over SSE, so it shouldn't escalate to a user-facing error toast. Thanks again for flagging the issue - happy to walk through the reproduction in #1550 if you're interested.

PR #1563, opened 2026-08-17
fix(terminal): keep app XDG namespace out of user-command environments

### LINKED ISSUE #1528: [Bug] Crush CLI fails to load configuration in PawWork integrated terminal
### What happened?

When I launch the [crush CLI](https://github.com/charmbracelet/crush) via the built-in terminal sidebar in PawWork, it fails to load my existing configuration (API keys, MCPs, skills) and prompts me for a new setup. However, running the exact same crush binary in Windows Terminal or the Zed editor sidebar loads my configuration perfectly. **Im not sure if the issue is with crush CLI itself or PawWork.**

#### My observations:
1. 
Windows Terminal: Works perfectly (loads config/skills/MCPs).
2. 
Zed Editor Sidebar: Works perfectly (loads config/skills/MCPs).
3. 
PawWork Sidebar: Fails to load config (prompts for setup).

I have verified that `$env:USERPROFILE` is identical across these terminals, so it does not appear to be a simple path or environment variable mismatch.

Not sure

Minor issue or polish

### Steps to reproduce

1. Configure crush CLI in a standard terminal (keys, MCPs, skills).
2. Open PawWork.
3. Open the integrated terminal sidebar in PawWork.
4. Run crush.
5. Observe that it prompts for new setup instead of loading existing config

crush should load the user configuration consistently regardless of which terminal emulator is hosting the session.

### PawWork version

2026.7.3

### OS version

Windows 11

Yes, every time

### Diagnostics

I cannot use my configured tools/skills within the PawWork sidebar, forcing me to duplicate my setup or switch terminals constantly, which is prone to configuration drift.

It's a minor issue, I wanted to use smallar models in crush alongside large models in pawwork for better workflow.

Here are the screenshots.
Windows terminal (opened in pawwork folder)👇
Image
PawWork built-in terminal (opened in pawwork folder)👇
Image

Astro-Han:
## Summary

- The desktop injector (`desktop-electron` `buildServerEnv`) now publishes `PAWWORK_USER_ENV_RESTORE` alongside its XDG_* namespace injection: a JSON map of each injected key to the user's pre-existing value (or `null` when the user had none).
- New `restoreUserEnv` (`packages/opencode/src/util/env.ts`) executes that instruction at every user-command boundary — PTY terminals (`pty/index.ts`), the bash tool (`tool/shell.ts` shellEnv), and composer `!command` shells (`session/prompt.ts`) — so user terminals see the user's environment, not the app's runtime namespace.
- PTY under bun-pty blank-overrides unset keys (bun-pty merges the parent's native environment, so deleted keys resurrect); under Electron's node-pty the deletion is a true unset.
- The Playwright e2e backend harness now mirrors the injector (sandbox XDG + restore instruction), and a new terminal spec walks the real user path with native pollution.

## Why

Fixes #1528. The app injects `XDG_CONFIG_HOME` etc. to namespace the embedded server's config/data/cache/state. Because the server runs in-process and reads `process.env` at module load, the injection mutates the shared environment, and every user-command child inherited it. XDG-following CLIs (crush) then treated `%APPDATA%\ai.pawwork.desktop\config` as their config home, found nothing, and prompted for first-run setup. Same failure would hit any Go/Rust CLI that follows XDG.

The restore-instruction design keeps the injected-key list in exactly one place (the injector); consumers execute it verbatim instead of keeping a second key list. Standalone opencode has no instruction and is unaffected.

## Related Issue

Fixes #1528

## Human Review Status

Pending

## Review Focus

- `restoreUserEnv` semantics (`packages/opencode/src/util/env.ts`): restore user value / unset injected key, return deleted keys for merge-resurrect suppression.
- The bun-pty blank-override branch in `pty/index.ts` and why deletion alone is insufficient there (see the pre-existing `OPENCODE_SERVER_*` comment for the same constraint).
- The e2e harness change (`packages/app/e2e/backend.ts`) now publishes the instruction for every e2e backend — intentional, it mirrors the desktop runtime.

## Risk Notes

- Under bun-pty, keys the user never had become empty strings in terminals instead of unset (bun-pty cannot unset natively-inherited keys). Go zero-value lookup, shell `:-` expansion, and Rust `dirs`' empty filter all treat `""` as unset; exotic consumers that read an empty XDG var as a literal path would misbehave — none known.
- Composer `!command` restoration is covered through the shared executor and the two e2e-tested surfaces (no dedicated session-prompt e2e seam exists); the wiring is identical one-liner.
- Windows-specific conpty env semantics were not verified locally (no Windows host); `@lydell/node-pty` passes the env explicitly per platform and the bash-tool probe test runs cross-shell. Desktop smoke CI covers the packaged path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Terminals and shell commands now correctly expose the host user’s XDG configuration, data, cache, and state settings.
  * Prevented application sandbox paths and environment restoration metadata from leaking into child shells and PTY sessions.
  * Improved handling of unset or malformed environment values.

* **Tests**
  * Added coverage for terminal, PTY, shell, and environment restoration behavior across supported shells.

### DIFF
--- a/packages/app/e2e/backend.ts
+++ b/packages/app/e2e/backend.ts
@@ -97,6 +97,8 @@ export function createBackendEnv(input: {
   sandbox: string
   llmUrl?: string
 }): Record<string, string | undefined> {
+  const base = input.base ?? process.env
+  const sandboxXdgKeys = ["XDG_CONFIG_HOME", "XDG_DATA_HOME", "XDG_CACHE_HOME", "XDG_STATE_HOME"] as const
   const env = {
     ...(input.base ?? process.env),
     OPENCODE_DISABLE_LSP_DOWNLOAD: "true",
@@ -112,6 +114,14 @@ export function createBackendEnv(input: {
     OPENCODE_E2E_ENABLED: "true",
     OPENCODE_E2E_LLM_URL: input.llmUrl,
   }
+  // Mirror the PawWork desktop injector (desktop-electron server.ts): the
+  // sandbox XDG namespaces the e2e backend itself, and the restore
+  // instruction carries the host's pre-existing values so user-command
+  // surfaces (terminals, bash tool) see the user's environment, not the
+  // sandbox. Same wire contract: the key list lives with the injector.
+  env.PAWWORK_USER_ENV_RESTORE = JSON.stringify(
+    Object.fromEntries(sandboxXdgKeys.map((key) => [key, base[key] ?? null])),
+  )
   for (const key of Object.keys(env)) {
     if (INTERNAL_SERVER_AUTH_ENV.has(key.toLowerCase())) delete env[key]
     else if (PROVIDER_ENV_PATTERN.test(key) || PROVIDER_ENV_EXTRA.has(key)) delete env[key]
--- a/packages/app/e2e/terminal/terminal-user-env.spec.ts
+++ b/packages/app/e2e/terminal/terminal-user-env.spec.ts
@@ -1,0 +1,25 @@
+import { runTerminal, waitTerminalReady } from "../actions"
+import { test, expect } from "../fixtures"
+import { terminalSelector } from "../selectors"
+import { terminalToggleKey } from "../utils"
+
+// The backend sandbox namespaces its own XDG dirs and publishes the PawWork
+// restore instruction, exactly like the desktop app. A terminal is a user
+// shell: XDG-following CLIs (crush in #1528) must see the user's config home,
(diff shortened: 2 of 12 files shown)


Astro-Han:
Resolving: the e2e terminal spec runs on the POSIX e2e runner (no Windows e2e check exists in CI); the probe's POSIX syntax is valid there. The restore logic itself is cross-platform (node-pty conpty honors deletion on Windows, verified via source). Cross-shell probe portability is a test-only follow-up, out of scope for this product fix.

PR #1597, opened 2026-08-23
fix(desktop): read global AGENTS.md via declarative dshHome

Astro-Han:
PawWork isolates DSH data under `~/.pawwork/dsh` (`dsh-dev`) to avoid v1 collision, but the global instructions file lives at `~/.pawwork/AGENTS.md`. `dsh-agent-instructions` only probes `$DSH_HOME/AGENTS.md`, so the global file was never injected.

**Previous approach (copy migration) was rejected in adversarial review**: a one-shot `cp ~/.pawwork/AGENTS.md → $DSH_HOME/AGENTS.md` creates a permanent two-source divergence (3 files with dev/prod), violates single-source-of-truth and Occam's coherence test, and adds fallback-path branching. Reviewers A/B/D converged on this.

**Redo as declarative single-source fix** (minimal, cleanest, lowest long-term entropy):

- `product.cordis.patch.yml`: add `agent-instructions` patch `{ dshHome: "~/.pawwork", maxBytes: 65536 }` so the native loader reads the user's file in place. Sessions/config stay in `$DSH_HOME`; global is shared across channels, no copy, no marker, idempotent and reversible. Reuses upstream `dsh-home-paths` contract, no fork.

Verification: `pnpm --filter @pawwork/desktop exec vitest run src/main/pawwork-home.test.ts` — 11 passed (original suite, no new migration code to test). Patch replaces the standard preset's `maxBytes: 65536` row; other fields use defaults.

Closes the v1→v2 regression where global instructions were silently dropped after home relocation.

### DIFF
--- a/packages/desktop-electron/resources/dsh/home/product.cordis.patch.yml
+++ b/packages/desktop-electron/resources/dsh/home/product.cordis.patch.yml
@@ -18,6 +18,10 @@
           - id: mimo-v2.5-free
           - id: nemotron-3-ultra-free
           - id: north-mini-code-free
+- id: agent-instructions
+  config:
+    dshHome: "~/.pawwork"
+    maxBytes: 65536
 - id: llm-deepseek
   disabled: true
 - insert:


Astro-Han:
感谢 CodeRabbit 指出 fallback 路径问题。

**Fix in 43a3e98d**:

- `migrateGlobalAgentsMd(home, onEvent, resolvedHome?)` 现以 `dirname(resolvedHome ?? home)` 求 `legacy` 源，fallback 时不再误读 `userData/AGENTS.md`
- `index.ts` 中先 `const resolvedHome = resolveDshHome(...)`，再 `migrateGlobalAgentsMd(dshHome, ..., resolvedHome)`
- 新增回归测试：`uses resolved home for legacy source when DSH home fell back to legacy userData path` — 15 passed

Astro-Han:
Redo per adversarial review synthesis (4 independent reviewers):

- **Problem definition**: correct mechanism, but framed as generic gap instead of v1→v2 regression with stranded `~/.pawwork/AGENTS.md`.
- **First principles/Occam**: copy migration fails coherence — one-shot copy leaves 2 authorities (`~/.pawwork` + `~/.pawwork/dsh`, 3 with dev) that diverge after first edit.
- **Optimality**: `mkdirSync` was redundant, failure branch untested, ordering invariant undocumented.
- **DSH plugin**: `migrateDshHome` must stay product launcher (cannot run inside home), but `migrateGlobalAgentsMd` should be declarative config, not copy. DSH already exposes `Config.dshHome`.

This force-push replaces the 2 copy-migration commits with the declarative single-source fix (`+4 lines` in `product.cordis.patch.yml`). No new product code to test; original suite still 11 passed. Addresses CodeRabbit fallback-path concern by deleting the code entirely.

PR #1600, opened 2026-08-24
fix(desktop): give Automations one page head

Astro-Han:
## What

The settings panel header already carries the shell's own actions and Close. The Automations section drew a second title bar underneath it with its own primary action, so two action clusters sat 8px apart and competed for the same corner.

- The section now owns **one page head** — title and intro only, at the DSH Plugins section's 18/600 + 13px, so both pages read the same.
- The primary action moves into a **single toolbar row** beside the search field and the filters it acts on, which also lifts the list about 60px up the page. Its label follows the move: `New automation` / 新建自动化, with the page intro carrying the fact that creation happens in chat.
- List rows split schedule from next run into **their own columns**, so a column of rows can be read down either edge.

## What this deliberately does not do

The panel header stays exactly as DSH ships it. `Open configuration file` is registered by `dsh-client-ui-settings-general` on the global `settings.action` slot, and its meaning does drift page to page — but hiding it means an attribute selector that closes the whole extension point, and anything DSH or another plugin registers there later would disappear with no test able to notice. That belongs upstream, not in a PawWork override. Moving our own action out of the corner is what the collision actually required.

So this change adds no new coupling to DSH internals: it touches only `resources/dsh/automations`, and uses only primitives, `--dsw-*` tokens and our own `.pawwork-*` classes.

## Verification

- `pnpm test` (380 tests), `pnpm typecheck`, `pnpm lint`
- New regression test: the page head contains no buttons and the primary action resolves inside the toolbar
- `pnpm smoke:ci` — real Electron, settings → Automations → create-in-chat → row → editor → save → delete
- Screenshotted in a real Electron window, empty state and a seeded three-row list

## Summary by CodeRabbit

* **UI Improvements**
  * Expanded the Automations page to use available width and simplified responsive layouts.
  * Added clearer status icons and labels for paused, completed, and run-limit automations.
  * Added an **Ended** filter and improved active automation filtering.
  * Displayed the next scheduled run for active automations.
* **Validation**
  * Prevented invalid, never-occurring schedules from being created or updated.
  * Added localized error messages for invalid schedules across automation actions.
* **Tests**
  * Expanded coverage for automation statuses, filters, and schedule validation.

### DIFF
--- a/packages/desktop-electron/resources/dsh/automations/lib/automations.cjs
+++ b/packages/desktop-electron/resources/dsh/automations/lib/automations.cjs
@@ -256,11 +256,7 @@ class AutomationStore {
     if (input.startedAt !== null) assertTimestamp(input.startedAt, 'startedAt');
     this.document.runs.push(structuredClone(input));
     const definition = this.document.definitions.find((entry) => entry.id === input.automationId);
-    if (definition?.kind === 'recurring'
-      && definition.stop?.kind === 'count'
-      && this.completedRunCount(definition.id) >= definition.stop.count) {
-      definition.nextFireAt = null;
-    }
+    if (definition?.kind === 'recurring' && this.runLimitReached(definition)) definition.nextFireAt = null;
     this.save();
     return 'imported';
   }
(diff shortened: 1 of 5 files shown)


Astro-Han:
Both review threads addressed:

- `ci-smoke.ts`: the TypeScript generic inside the CDP evaluation string was the CI failure; it now reads as plain JavaScript (`779d766`).
- `automations.cjs`: correct — `claimDue()` clears `nextFireAt` before the run lands, so a one-shot in flight reported `terminalReason: "completed"` alongside its `activeRun`. The list RPC now withholds the terminal reason while a run is active, with a test that fails without the fix.

PR #1617, opened 2026-08-31
fix(desktop): stop killing a slow DSH start

### LINKED ISSUE #1614: [Bug] Windows Not Compatible ( Update of 8.8.26 )
### What happened?

Here are the logs i am getting, this update doesn't load at all on windows. previous version was working fine. 

Image

[main.log](https://github.com/user-attachments/files/31604863/main.log) 

App flow or product behavior

Blocks me from using PawWork

### Steps to reproduce

1. Install the latest update 
2. App won't loading at all. 

I except to load the program and work / code with it 

### PawWork version

v8.8

### OS version

Windows 10 

Yes, every time

### Diagnostics

_No response_

Astro-Han:
Closes #1614.

## What was wrong

DSH prints nothing at all until its ready line, so elapsed silence carries no information about whether the runtime is wedged or merely slow. Both startup deadlines read that silence as failure:

- `dsh-sidecar.ts` force-terminated the runtime 30s after spawn
- `dsh-lifecycle.ts` failed the start 30s after the product page began loading

On a first launch that is still unpacking behind an antivirus scan, that converts a start which was about to succeed into an app that can never open. The reporter's log is the whole story:

```
[00:05:18] spawning DSH sidecar
[00:05:49] DSH lifecycle failed: DSH did not announce readiness within 30000ms
```

Thirty seconds, zero output — which looks like a dead sidecar but is what a healthy one looks like too, right up until the moment it is ready.

## Evidence

A Windows reproduction job (`repro-1614.yml`, second commit) installs each release on `windows-latest` and records the real startup timeline. A healthy start emits exactly one stdout line, the ready line, ~2.5s after spawn:

```
06:19:43.248  spawning DSH sidecar
06:19:45.741  DSH stdout { chunk: 'dsh web: http://127.0.0.1:59843' }
```

There is no progress output to distinguish slow from stuck, and process sampling confirms the sidecar is a normal child of the main process throughout. The same job also rules out pre-existing v1 state as the trigger: v1 leaves no `<userData>\dsh` behind, and the v1→v2 upgrade leg starts in 2s.

## What changed

The failures that are real already announce themselves — the process exits, or the spawn errors — and both paths are untouched. The deadlines now only report:

- the sidecar's threshold swaps the startup page for one that says a first launch after an update can take a few minutes
- the product threshold logs, and does not navigate: the window is already loading the product page, and navigating away would abort the very load being waited on

Nothing is cancelled when either passes. A start that needs four minutes now takes four minutes instead of failing at thirty seconds.

## Verification

- `pnpm run typecheck`, `pnpm run lint`, `pnpm run test` (469 vitest + 138 + 10 node) all pass
- Verified in a real Electron app on macOS with the sidecar held silent: the slow page renders in the system locale and the app stays alive well past the old 30s deadline instead of showing the startup-failure dialog

## Not covered here

This does not explain why the reporter's machine was slow, and deliberately so — the fix does not need to know. If it turns out to be something we can address at the source, that is a separate change.

## Summary by CodeRabbit

* **New Features**
  * Added slow-start detection for the desktop runtime and product interface.
  * Displays a localized notice when startup takes longer than expected while continuing initialization.
  * Preserves the loading state until the product is ready.

* **Bug Fixes**
  * Prevents slow startup from being incorrectly treated as a failure or terminating the runtime.

* **Tests**
  * Added coverage for slow, non-failing startup behavior and readiness transitions.

* **Chores**
  * Added a Windows workflow for reproducing and collecting evidence for startup issues.

### DIFF
--- a/.github/workflows/repro-1614.yml
+++ b/.github/workflows/repro-1614.yml
@@ -1,0 +1,213 @@
+# Reproduction harness for #1614: PawWork v2 never finishes starting on Windows
+# after an update from v1, with the DSH sidecar producing no output at all
+# before the 30s readiness timeout.
+#
+# ci.yml already installs and smokes a freshly packaged v2 on windows-latest and
+# passes, so a clean install is not the trigger. The only thing the reporter's
+# log has that CI does not is pre-existing v1 state under
+# %APPDATA%\ai.pawwork.desktop, which v2 renames into ~\.pawwork\dsh on first
+# run. This workflow runs both halves of that comparison so the difference is
+# attributable rather than assumed.
+#
+# PAWWORK_CI_SMOKE=true is set for the launch (but NOT PAWWORK_CI_SMOKE_HOME, so
+# every path stays the real user-profile path a shipped install would use). It
+# only changes two things that matter here: the startup failure exits the
+# process instead of opening a modal dialog that would hang the runner, and a
+# successful start writes ci-smoke-ready.json we can poll for.
+#
+# Neither leg fails the job on a reproduction - both always upload their logs,
+# and the verdict goes to the run summary.
+name: repro-1614
+
+on:
+  workflow_dispatch:
+    inputs:
+      baseline_version:
+        description: v1 version installed first (upgrade leg only)
+        default: "2026.8.5"
+      target_version:
+        description: v2 version under test
+        default: "2026.8.8"
+  push:
+    branches: [debug/1614-windows-upgrade-repro]
+
+permissions:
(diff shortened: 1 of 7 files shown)


Astro-Han:
Superseded by #1619, which is the same fix as pure deletion: both deadlines removed, no slow-start page, no new option or state. The reproduction workflow is dropped too — its verdict grepped for a log string the fix itself deletes, so it could only ever report "not reproduced". Its evidence is quoted in #1619.

PR #1624, opened 2026-09-01
fix(desktop): offer the plugin removal button when a plugin fails to load

Astro-Han:
## Problem

A profile plugin compiled against an older DSH takes the whole app down. cordis imports every loader entry with `Promise.allSettled`, rethrows the failures and rolls the entire configuration tree back, so a single stale plugin ends the boot with `DSH exited with code 1` — no window, and nothing the user can do from inside the app.

2026.9.1 moved DSH from 0.1.1-rc.2 to 0.1.2-alpha.3 (#1618, #1620), and 0.1.2-alpha.1 dropped `settingsNamespace` / `deepEqualJson` / `installSettingsSection` from `@deepseek-ai/dsh-settings` and `CallId` from `@deepseek-ai/dsh-llm`. Plugins that were fine before now fail at their top-level named imports:

```
Error: failed to import loader entry better-sidebar (dsh-better-sidebar): The requested module
'@deepseek-ai/dsh-settings' does not provide an export named 'settingsNamespace'
```

`showDshFailure` already knows how to recover from this class of failure — it drops one row from `dsh.profile.bundles` and restarts — but attribution only matched `cannot resolve profile bundle "<name>"`, the shape a rolled-back market install leaves behind (#1601). A plugin that is installed and merely will not import matched nothing, so the dialog showed a Node stack and a Quit button.

## Change

- Attribute loader failures from `failed to import loader entry <id> (<name>)`. The parenthesized half is the package, and the package is what the manifest holds. Only the `import` stage counts: the same failure resurfaces one level up as `failed to apply loader entry include (cordis:include)`, and that wrapper names loader plumbing no profile can declare.
- `unresolvedProfileBundle` becomes `failingProfileBundle`, returning the bundle plus a `missing` / `incompatible` cause so the dialog can say which one it is. "Not fully installed" is wrong for a plugin that is installed and merely incompatible.
- Grow the retained DSH stderr from 4 KB to 40 KB. Node prints one failed import as a chain of nested causes over deep pnpm paths, well past 4 KB on its own, so with several failing plugins the sentence that names them fell off the front of the tail and nothing was attributed at all. What the dialog shows stays capped at 4 KB.

Out of scope, and deliberately not attempted: pre-flighting plugin API compatibility before boot. It needs a hand-written ESM/exports subpath resolver (`dsh-llm` alone has eight subpath entries), false positives would silently disable working plugins, and it still misses failures down the dependency chain.

## Verification

`pnpm --filter @pawwork/desktop test` (478), `pnpm typecheck`, `pnpm lint`.

Real app, driven over CDP against a scratch `DSH_HOME` whose `profiles/web` carries `dsh-better-sidebar@0.16.1` and `dsh-plugin-subscriptions@0.5.3` (both reference removed exports) with both declared in `dsh.profile.bundles`. Before the change the dialog offered only `["Try Again","Show Log","Report a Problem","Quit"]`. After:

```
The plugin "dsh-better-sidebar" is not compatible with this version of PawWork, which stops the
runtime from starting. Removing it lets PawWork open again; you can install it again from Settings
once the plugin ships an update.
buttons: ["Remove Plugin and Retry","Try Again","Show Log","Report a Problem","Quit"]
```

Clicking through removes one bundle per restart — `dsh-better-sidebar`, then `dsh-plugin-subscriptions` — and the third start reaches the product URL. `@deepseek-ai/dsh-base`, `@deepseek-ai/dsh-web-app` and the `dependencies` entries are untouched. Checked in both locales; the Chinese copy reads 「插件『X』与当前版本的爪印不兼容…」.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desktop plugin failure detection for missing or incompatible plugins.
  * Updated failure dialogs with cause-specific messaging and details, making troubleshooting clearer.
  * Improved handling of complex loader errors and filtered out unrelated failures from plugin repair prompts.
  * Increased captured error context while keeping displayed excerpts concise for readability.

### DIFF
--- a/packages/desktop-electron/src/main/dsh-profile-repair.test.ts
+++ b/packages/desktop-electron/src/main/dsh-profile-repair.test.ts
@@ -2,16 +2,30 @@ import { describe, expect, test } from "vitest"
 import { mkdtempSync, readFileSync, writeFileSync } from "node:fs"
 import { tmpdir } from "node:os"
 import { join } from "node:path"
-import { removeProfileBundle, unresolvedProfileBundle } from "./dsh-profile-repair"
+import { failingProfileBundle, removeProfileBundle } from "./dsh-profile-repair"
 
 // The exact shape DSH dies with when a market install left a bundle behind.
-const FAILURE_OUTPUT = [
+const MISSING_OUTPUT = [
   "file:///app/node_modules/@deepseek-ai/dsh-app-boot/lib/index.js:523",
   "\tthrow new Error(`${binName}: cannot resolve profile bundle ${JSON.stringify(packageName)} from the dsh installation`);",
   "",
   "Error: dsh: cannot resolve profile bundle \"dsh-lark-bot\" from the dsh installation or /home/u/.pawwork/dsh/profiles/web;"
     + " run 'dsh plugin --profile web install' if its dependency is not installed",
   "    at resolveBundleDir (file:///app/node_modules/@deepseek-ai/dsh-app-boot/lib/index.js:523:8)",
+].join("\n")
+
+// The shape a plugin built against a removed DSH export dies with: the package
+// is installed, the loader entry just cannot be imported. `better-sidebar` is
+// the entry id and `dsh-better-sidebar` the package, and only the package is
+// ever a bundle.
+const INCOMPATIBLE_OUTPUT = [
+  "Error: failed to import loader entry better-sidebar (dsh-better-sidebar):"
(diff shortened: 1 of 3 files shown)


Astro-Han:
Superseded by #1625 — the branch was renamed to `fix/plugin-load-failure-recovery` to match the repo convention, and GitHub closed this PR instead of retargeting it. Same commit, same review state.

PR #1639, opened 2026-09-03
fix(opencode): address the zen gateway as the opencode desktop client

Astro-Han:
## The problem

The six OpenCode Free models are the product's default and its only out-of-the-box way to get an answer. The zen gateway meters that free tier per client identity, and it reads the identity off `user-agent`.

From one IP within one second:

| `user-agent` | `POST https://opencode.ai/zen/v1/chat/completions` |
| --- | --- |
| `opencode/stable/1.0.0/opencode` | 200 |
| `opencode/x/y/z` | 200 |
| `deepseek-harness/0.1.2-alpha.5 (+https://github.com/deepseek-ai/deepseek-harness)` — what PawWork sends today | 429 |
| curl default / `PawWork/2026.9.3 Electron/41` / `node` / empty | 429 |

The 429 body is `{"type":"error","error":{"type":"FreeUsageLimitError","message":"Rate limit exceeded. Please try again later."}}`. So PawWork's free traffic is metered in a pool shared with every unrecognized client, not the one these models are published under. Nothing else in the request distinguishes them — `authorization: Bearer public` is identical, and `public` is still the current sentinel upstream (`packages/core/src/plugin/provider/opencode.ts`, opencode main `8d4ef01621`); it is not a token and has not expired.

## Why it regressed

v1 already sent this identity. #940 added `httpIdentity()` — `User-Agent: opencode/latest/<version>/desktop` plus `x-opencode-client` — for every outbound request to an OpenCode backend, down to rewriting Electron's `userAgentFallback`. #1573 moved the LLM path onto DSH, where `dsh-llm`'s `attributionHeaders()` hardcodes its own user-agent, and the identity went with it.

## The change

Both Free routes declare the current opencode desktop client identity, taken from the upstream tree rather than from v1: its embedded server builds `opencode/<channel>/<version>/<client>`, a release build's channel is `latest`, and the desktop sets the client to `desktop`.

`dsh-llm-pi-ai` dropped any deployment header colliding with attribution, so a route could not state an identity at all. The patch inverts that precedence — a name a route declares wins, case-insensitively — on the completion path and on model discovery alike, so a route admitted as one client is not listed as another. A route that declares nothing sends attribution unchanged.

The four per-session headers that client also sends (`x-session-affinity`, `X-Session-Id`, `x-opencode-project`, `x-opencode-session`) are deliberately absent: they carry one session's identity, and a route profile can only declare constants, which would report every user's every turn as the same session.

The periodic catalog refresh re-asserts the headers alongside `api` and `baseURL`, so an existing install picks them up without waiting for a fresh home, and the no-churn check accounts for them.

## Verification

Against the real app over CDP, with a recording proxy standing in for the gateway: the turn answered (`assistant/message`, `turn/end`) and both outbound requests carried

```
user-agent: opencode/latest/1.18.15/desktop
x-opencode-client: desktop
authorization: Bearer public
```

with no attribution header. `dsh-product-home.test.ts` pins the packaged patch against `OPENCODE_CLIENT_HEADERS` so the two cannot drift, and a unit test pins the identity's shape. Typecheck, lint and 167 tests pass.

## Summary by CodeRabbit

- **New Features**
  - OpenCode provider connections now send desktop client identification headers.
  - Provider discovery and completion requests consistently use configured client identity information.

- **Bug Fixes**
  - Custom provider headers now correctly override default attribution headers, regardless of capitalization.
  - Providers that cannot authenticate through supported API keys are no longer shown in the catalog.

### DIFF
--- a/packages/desktop-electron/resources/dsh/home/product.cordis.patch.yml
+++ b/packages/desktop-electron/resources/dsh/home/product.cordis.patch.yml
@@ -16,6 +16,9 @@
     providers:
       opencode:
         apiKeyEnv: OPENCODE_API_KEY
+        headers:
+          user-agent: opencode/latest/1.18.15/desktop
+          x-opencode-client: desktop
         displayName: OpenCode Free
         api: openai-completions
         baseURL: https://opencode.ai/zen/v1
@@ -27,6 +30,9 @@
           - id: nemotron-3.5-lightning-free
       opencode-responses:
         apiKeyEnv: OPENCODE_API_KEY
+        headers:
+          user-agent: opencode/latest/1.18.15/desktop
+          x-opencode-client: desktop
         displayName: OpenCode Free (Responses)
         api: openai-responses
         baseURL: https://opencode.ai/zen/v1
--- a/packages/desktop-electron/resources/dsh/product/lib/opencode-free.cjs
+++ b/packages/desktop-electron/resources/dsh/product/lib/opencode-free.cjs
@@ -15,6 +15,33 @@ const OPENCODE_MODELS_URL = 'https://models.dev/api.json';
 const LLM_PI_AI_NAMESPACE = 'llm-pi-ai';
 const OPENCODE_ROUTE_BASE_URL = 'https://opencode.ai/zen/v1';
 /**
+ * How PawWork addresses the zen gateway.
+ *
+ * The gateway meters free-tier traffic per client identity, and an unrecognized
+ * `user-agent` lands in a pool that answers `FreeUsageLimitError` (HTTP 429)
+ * long before the desktop-client pool does. Both routes send the desktop client
+ * identity so a PawWork user gets the free access the models are published
+ * under. `user-agent` is lowercase because the adapter compares header names
+ * case-insensitively and prefers the spelling a route declares.
+ *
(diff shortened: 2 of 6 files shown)


Astro-Han:
Closing: the premise was wrong. v2 has had this identity since #1573 — `resources/dsh/zen-identity.mjs` is preloaded into the sidecar with `--import` and rewrites `user-agent` for every `opencode.ai/zen*` request, above pi-ai.

Measured on the real app against the real gateway, with a log point inside the wrapper: both outbound `https://opencode.ai/zen/v1/chat/completions` requests of a turn go through it, so what ships today is `opencode/latest/1.16.2/cli`, which the gateway accepts. There is no regression to restore.

My earlier verification was invalid: the recording proxy sat on `127.0.0.1`, which `isOpenCodeZenUrl` rejects, so the wrapper never fired and I measured only the route headers this PR adds — which the wrapper would overwrite in real traffic. The 429s came from my own curl probes with a non-opencode user-agent, never from the app.

The route headers here are dead weight and the pi-ai patch has no reason to exist. What remains is freshness only: the constant is pinned at `1.16.2/cli` while the current upstream desktop client is `1.18.15/desktop`. That is a one-line change to the existing wrapper, filed separately.

PR #1656, opened 2026-09-10
fix(automations): treat a definition with no usable timestamp as unscheduled

Astro-Han:
The automation scheduler read `nextFireAt` straight out of the stored document, and only `null` counted as "not scheduled". A definition carrying anything else broke both of its consumers.

| Consumer | With a non-timestamp `nextFireAt` |
| --- | --- |
| `arm()` | `Math.min` over the targets yields `NaN`, so the timer is armed on `NaN` and re-arms about once a millisecond for the life of the sidecar |
| `start()` | an unpaused definition is passed to `claimDue`, whose run record rejects a non-integer timestamp, so `start()` throws |

Measured on `origin/main` with a definition whose record has no `nextFireAt`:

```
TimeoutNaNWarning: NaN is not a number. Timeout duration was set to 1.
timer arms in 300ms: 260
```

```
Error: triggeredAt must be a non-negative integer
    at AutomationStore.createRunRecord
    at AutomationStore.claimDue
    at AutomationScheduler.start
```

The `start()` rejection is caught and logged through `ctx.logger.warn`, and the sidecar has no console exporter for the cordis logger, so that path ends with every automation dead and nothing in `main.log`.

`Number.isFinite` is now asked once, in a predicate both consumers share, instead of each of them restating what counts as scheduled.

## Scope

Every supported writer stores a real timestamp or `null`: `createDefinition` computes one, `importDefinition` normalizes the rhythm before it does, and `NaN` does not survive `JSON.stringify`. So this guards a state that reaches the scheduler only from a document written by something else. The `TimeoutNaNWarning` this started from does not reproduce — zero occurrences in `main.log` on this machine or in a fresh raw smoke run, and both real stores hold a valid `null`. It is hardening, not a live-bug fix.

## Verification

`node --test` 165/165 with one new regression test, vitest 542/542, tsgo and eslint clean. Removing the guard fails the new test; the test drives the state through the store file, the way it would actually arrive.

### DIFF
--- a/packages/desktop-electron/resources/dsh/automations/lib/automations.cjs
+++ b/packages/desktop-electron/resources/dsh/automations/lib/automations.cjs
@@ -50,6 +50,15 @@ function recurringNext(createdAt, everyMs, after) {
   return createdAt + (Math.floor(elapsed / everyMs) + 1) * everyMs;
 }
 
+// The store persists whatever the document on disk holds, so a definition can
+// reach the scheduler without a usable timestamp. Only a real one counts as
+// scheduled: as a delay it arms a timer every millisecond for the life of the
+// process, and as a claim target it fails the run record's own validation and
+// leaves the scheduler unstarted.
+function scheduledAt(definition) {
+  return Number.isFinite(definition.nextFireAt) ? definition.nextFireAt : null;
+}
+
 function definitionNext(definition, after, completedRunCount = 0) {
   if (definition.paused) return null;
   if (definition.stop?.kind === 'count' && completedRunCount >= definition.stop.count) return null;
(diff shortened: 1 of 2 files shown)


Astro-Han:
Closing without merging. The state this guards cannot be produced by any supported writer: createDefinition computes a timestamp, importDefinition normalizes the rhythm before it does, NaN does not survive JSON.stringify, and the store writes atomically. The TimeoutNaNWarning this started from does not reproduce — zero occurrences in main.log or a fresh raw smoke run, and both real stores on this machine hold a valid null.

Guarding the field at its two consumers also copies the store's authority outward, and it would make the read-side gap look handled when it is not: the store validates everything it writes and nothing it reads. That gap has the same reachability problem, so it is not worth building either until something can actually produce an invalid document.

